PR #1, opened 2016-12-05
account for reverse intensity functionality being not present

waxenegger:
This PR deals with the possibility of reverse intensity not being functionality available in the backend.

In that case the -r added to the image retrieval url will prevent other channel info supplied to not be processed properly so that we really want no -r/r appended.

Furthermore the checkbox will be disabled in that case


jburel:
Closing and re-opening PR to kick start travis (forgot to hook it up)

PR #3, opened 2016-12-07
version bugfix

waxenegger:
The built js referenced in the html is at the moment hard-coded (incl. a version name based on package.info) which can break things with version bumping.

Given that releases happen via pypi and code is tagged appropriately with each release, the need to have yet another version to maintain within the java script is obsolete. 

Therefore the version in package info was set to 0.0.0 and the webpack config does not include any version number in the built js file name.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -20,7 +20,9 @@
   </head>
 
   <body class="container-fluid">
-      <script type="text/javascript" src="{% static 'omero_iviewer/bundle1.0.0.js' %}"></script>
+      <script type="text/javascript"
+              src="{% static 'omero_iviewer/bundle.js' %}">
+      </script>
   </body>
 
 </html>
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "0.0.1",
+  "version": "0.0.0",
   "description": "OMERO.iviewer",
   "scripts": {
     "dev": "webpack-dev-server --config webpack.dev.config.js --hot --inline --progress --devtool eval",
--- a/webpack.prod.config.js
+++ b/webpack.prod.config.js
@@ -11,7 +11,7 @@ module.exports = {
   },
   output: {
     path: path.join(__dirname, 'build'),
-    filename: 'bundle' + pkg.version + '.js',
+    filename: 'bundle.js',
     publicPath: '/static/omero_iviewer/'
   },
   plugins: [


jburel:
tested locally
no more version number and index no longer has a hardcoded value

src/index.html
```diff
@@ -20,7 +20,9 @@
   </head>
 
   <body class="container-fluid">
-      <script type="text/javascript" src="{% static 'omero_iviewer/bundle1.0.0.js' %}"></script>
+      <script type="text/javascript"
+              src="{% static 'omero_iviewer/bundle.js' %}">
```
aleksandra-tarkowska: if that is internal lib, could you suffix with application version. I think client browser won't pick up newer changes when we upgrade?

aleksandra-tarkowska:
How viewer is going to version statics? Could you suffix with application version. I think client browser won't pick up newer changes when we upgrade? Please look how this is done in web https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webclient/templates/webclient/login.html#L26

waxenegger:
@aleksandra-tarkowska I don't think the two are necessarily linked, caching & other version related issues. Nor does it have to be the end of versioning. This was primarily a bugfix, otherwise the deploy would be broken.

But I think that, if we are not talking about caching alone, versions should be linked to the code for most things to make sense, bugs, release, logging, etc. And, so the various places would have to be kept in sync always. It may make better sense for the latter purposes to inject the version into the js to make use of it, rather than have it in the name and outside the apps 'realm' so to speak.

aleksandra-tarkowska:
@waxenegger could you provide the example of what you are talking about?

aleksandra-tarkowska:
@waxenegger could you provide the example of what you are talking about, I am not sure I understand?

I am sure that 
```
+      <script type="text/javascript"
+              src="{% static 'omero_iviewer/bundle.js' %}">
+      </script>
```

bundle.js will be cached in client browser and user has to do hard refresh to load the latest

As the view render the template in https://github.com/ome/omero-iviewer/blob/master/plugin/omero_iviewer/views.py#L20 what is the problem of using url_suffix?

waxenegger:
@aleksandra-tarkowska I'm not exactly sure what you mean with example. I mean that if for instance one wanted a version within the app then it can be handed in as part of the build still. I'm confident such a mechanism can be achieved.

waxenegger:
Cannot say I have ever noticed that it got cached (without increment) and I used both webserver and dev server during development. After a refresh I always got the up-to-date script with 4 different browsers I used.

aleksandra-tarkowska:
@waxenegger because you refresh, user won't. I am only saying that iviewer should add its own https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/custom_context_processor.py#L4, this can be injected automaticity in `iviewer/settins.py`. Then Django template will render it with the desired app version and problem will be sorted. This has to happen on the view level not JS level!

waxenegger:
But if the user doesn't refresh the page, they won't ever get the newer version unless you wrote your own loader or told webpack to reload dynamically.

Either way, I don't mind if we put something in the view.

aleksandra-tarkowska:
I think refresh button and requesting url are 2 different things from the browser perspective

waxenegger:
yes but keep in mind: this is a <script> fetch and then webpack does the rest. it is not a regular ajax call or other browser request in the context of an already loaded script/page.

but, as I said, I will put in the prefix, shouldn't take too long.

waxenegger:
ok, problem. while the suffix will be applied to the css it won't for the script, the reason being, as I suspected that webpack has a say in this:

 	// This file contains only the entry chunk.
 	// The chunk loading function for additional chunks
 	__webpack_require__.e = function requireEnsure(chunkId, callback) {
 		// "0" is the signal for "already loaded"
 		if(installedChunks[chunkId] === 0)
 			return callback.call(null, __webpack_require__);

 		// an array means "currently loading".
 		if(installedChunks[chunkId] !== undefined) {
 			installedChunks[chunkId].push(callback);
 		} else {
 			// start chunk loading
 			installedChunks[chunkId] = [callback];
 			var head = document.getElementsByTagName('head')[0];
 			var script = document.createElement('script');
 			script.type = 'text/javascript';
 			script.charset = 'utf-8';
 			script.async = true;

 			script.src = __webpack_require__.p + "" + chunkId + ".bundle.js";
 			head.appendChild(script);
 		}
 	};

 	// expose the modules object (__webpack_modules__)
 	__webpack_require__.m = modules;

 	// expose the module cache
 	__webpack_require__.c = installedModules;

 	// __webpack_public_path__
 	__webpack_require__.p = "/static/omero_iviewer/";

 	// Load entry module and return exports
 	return __webpack_require__(0);



// WEBPACK FOOTER //
// webpack/bootstrap 9a3b33259e7469841fd0

Essentially it reloads, reinserting a new script tag so that the suffix is lost.
To hook into this is probably not going to be a views only thing I'm afraid. What's more, the solutions I have come across are not such that I'm able to use the omero suffix...

aleksandra-tarkowska:
yes great, I will give you an example code how to add custom sufix as we cannot reuse omeroweb one that is omero version.

waxenegger:
ok thanks.

Just to be absolutely paranoid I guess I'll still add a hash for the chunks in the webpack config since, strictly speaking, the bundle chunks would just be incremented in a predictable/repeatable fashion and that would mean that a stale chunk could potentially be loaded but not if the main bundle contains a new hash. This has the very pleasant side effect that it does not affect the build since only the main bundle has to be named in the html...

aleksandra-tarkowska:
please have a look at https://github.com/ome/omero-mapr/pull/4/

if you find any improvement to that setup please let me know

waxenegger:
This is dealt with in: https://github.com/ome/omero-iviewer/pull/4

my approach is slightly slimmer as I'm not using custom settings (seemed a bit over the top in my case) but the final goal should have been achieved.

PR #2, opened 2016-12-06
Compile in setup.py

jburel:
Compile code in setup.py

cc @aleksandra-tarkowska @waxenegger

PR #4, opened 2016-12-08
cache buster

waxenegger:
Appending a suffix to the resources requested is going to prevent potential caching along the way for when version numbers don't change.

image


waxenegger:
@jburel the approach I took was a minimalist one to achieve the goal. Please let me know if the VERSION being in its own version.py is problem to the pypi setup. From first glance it didn't seem so but I'm by no means an expert with that.

aleksandra-tarkowska:
I am afraid that is not a pep standard

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#8](https://ci.openmicroscopy.org/job/iviewer-merge/8/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/8/consoleText) for more details.
Possible conflicts:
  - PR #2 jburel 'Compile in setup.py'
    - plugin/setup.py

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#9](https://ci.openmicroscopy.org/job/iviewer-merge/9/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/9/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - plugin/setup.py

jburel:
--exclude

waxenegger:
@jburel I have brought this up-to-date with master since quite a few PRs meanwhile got merged. I believe this should not conflict with any other PR at the moment so that we could try and take off the exclude I suppose.

@aleksandra-tarkowska I'm still in the dark about your PEP comment.
I believe you were trying to say that version.py is not good as a location to define the version. As mentioned back when we discussed it this was merely a means to define version in one place and being able to use it in more than one location without having to define and maintain it a multiple times. If there is a standard to that end then I'd like to ask you to please detail the proper way of doing so since I have no clue as to what a PEP compliant solution ought to look like.

waxenegger:
@jburel noticed this one had a conflict which should be easy to fix. This has been alive for a long time. Wouldn't it be good idea to include it in the release?

joshmoore:
My experience has been that in general (even for testing) having this in place is _very_ useful and prevents confusion.

waxenegger:
--exclude

plugin/omero_iviewer/version.py
```diff
@@ -0,0 +1,24 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
+from omero_iviewer.utils import get_version
+
+
+VERSION = (0, 0, 8)
```
jburel: Note: this is for testing purpose only
To be replaced by 0.1.0 when we are ready

PR #5, opened 2016-12-08
open_with fix

will-moore:
This fixes the "Open with" that was broken due to the changing of Label in Open_with menu.

To test:
 - setup open_with (if not done already) as described on https://github.com/ome/omero-iviewer/tree/master/plugin/omero_iviewer (uses label "OMERO.iviewer"), restart web etc.
 - Right-click menu in Tree, Open_with > OMERO.iviewer  should open iviewer in a new window with the selected image.
cc @waxenegger

### DIFF
--- a/src/openwith.js
+++ b/src/openwith.js
@@ -1,3 +1,3 @@
-OME.setOpenWithUrlProvider("omero_iviewer", function(selected, url) {
+OME.setOpenWithUrlProvider("OMERO.iviewer", function(selected, url) {
     return url + selected[0].id + "/";
 });


jburel:
@will-moore: thanks

jburel:
@will-moore: we should follow the same name convention for figure
it is currently ``omero_figure``, ``OMERO.figure`` will be more suitable

jburel:
https://trello.com/c/649GT6lU/21-unify-open-with-name to figure so we unify before we release anything

jburel:
Discussed with @will-moore this am:
A "label" will be added to Open with so we can clearly distinguish between "app id" e.g. "omero_iviewer" and label.
If no label provided then the id will be used for  display in the menu

Closing this PR, I will open a PR to fix the README
README to be updated again when the Open With work is done

PR #6, opened 2016-12-08
title supposed to show iViewer: image name

waxenegger:
Analog to the present web viewer the title now reflects the image name of the image viewer, prefixed with "iViewer".

image

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,7 +16,7 @@
         {% endfor %}
       </script>
 
-      <title>Viewer NG</title>
+      <title>iViewer</title>
   </head>
 
   <body class="container-fluid">
--- a/src/app/header.js
+++ b/src/app/header.js
@@ -94,13 +94,13 @@ export class Header extends EventSubscriber {
      */
      onImageConfigChange(params = {}) {
          let conf = this.context.getImageConfig(params.config_id);
+
          this.setTitle(
              conf && conf.image_info && conf.image_info.image_name ?
                 conf.image_info.image_name : null);
-         if (conf === null) return;
 
-         this.image_info =
-             this.context.getImageConfig(params.config_id).image_info;
+         if (conf === null) return;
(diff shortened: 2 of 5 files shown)


src/index.html
```diff
@@ -16,7 +16,7 @@
         {% endfor %}
       </script>
 
-      <title>Viewer NG</title>
+      <title>iViewer</title>
```
jburel: should be OMERO.iviewer unless we drop the "omero" internally

waxenegger: I think that's a matter of taste. I don't have a strong feeling toward one or the other. To be succinct in the title is perhaps preferable, especially if image names are long. After all, as you said, it's just within the app (for display) and has no technical implications.

will-moore:
Working fine for me.
I don't have a strong opinion on "OMERO.iviewer:" vv "iViewer:" in the title.
It would be nice to have "OMERO" there but since space is at a premium, it's probably better to show more of the Image name.

screen shot 2016-12-12 at 16 24 05

Good to merge.

joshmoore:
Sorry for coming late to this, but I'd propose we discuss this again before going live with iviewer. In general, the tab naming strategy is an issue. For IDR, bookmarks are all just: "webclient". Now "iViewer" is naming itself. What other strategies do we have?

If I had to choose between "$APP" and "$WEBSITE" I'd prefer the latter. i.e. OMERO or IDR on all my bookmarks rather than iViewer on some, webclient on others, etc.

If we can have all in a fairly tidy fashion with sensible separators, I'd go for: "$PAGE / $APP / $WEBSITE"

jburel:
Not late at all.
mainly merging so avoid conflict. Still some discussions needed for the app so we have a common way of dealing with thing
overall I will prefer OMERO to any "derive" product unless it is very specific and not going to make it back to mainline

PR #9, opened 2016-12-09
Right panel styling

will-moore:
Looking to improve the invisible text in disabled buttons on right panel.
Used more bootstrap styles.
Also put the buttons into a ```btn-group``` container.

Original panel width:
screen shot 2016-12-09 at 13 51 56

Narrow panel (wraps to new row):
screen shot 2016-12-09 at 12 37 56

Without the ```btn-group``` (after first commit only):
screen shot 2016-12-09 at 12 24 27

cc @waxenegger

### DIFF
--- a/src/css/app.css
+++ b/src/css/app.css
@@ -64,6 +64,9 @@ a.ui-button:active,
 .ui-icon-triangle-1-s,.ui-icon-triangle-1-n {
     height: 100%;
 }
+.full-width {
+    width: 100%;
+}
 /** JQUERY UI adjustments END **/
 
 /** SPECTRUM colorpicker adjustments START **/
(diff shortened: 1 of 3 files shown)


waxenegger:
looks good @will-moore, I'll give that a try for any possible side-effects but the style is definitely improved, no question there.

waxenegger:
The button groups look good. The only thing I'm not sure is the effect if one goes really narrow

image

Now that's probably more of a philosophical question of what one wants to happen but I feel that if the buttons are getting stacked on top, the channels get pushed down, and perhaps that does not matter because the person who pushed the panel over might not have wanted to do much with them anyhow in that narrow state. If s/he pushed over to just have enough space to toggle channels still, they stacking effect might not be so helpful.

But playing around with it, one could put a min-width on the div that is the btn-group class I suppose and that would at least keep the individual buttons in the group together. Not sure if this is the best idea but it's an idea.

will-moore:
I personally think that having the buttons wrap into a single column is fine for a very narrow panel. If they didn't wrap you wouldn't be able to use them, and forcing the channel buttons down a bit doesn't hurt since they're still usable.

waxenegger:
Fair enough. I think this is good to be merged then. 

As a matter of fact I'm thinking of doing the same for the rois panel buttons.

PR #8, opened 2016-12-09
prefixing ability for resource requests

waxenegger:
see: https://trello.com/c/faOotWHI/246-support-prefix

the prefix is now taken into account, so that if I set 'omero.web.prefix' to '/me/fancy/prefix' urls use it.

prefix

image

Test:
Check on cowfish using open-with iviewer or by using the url: https://cowfish.openmicroscopy.org/webtrial/omero_iviewer/fill_in_existing_image_id

One way to test is to open your browser's developer tools (e.g. chrome: F12) and check the network tab to see if there are any requests for static or dynamic resources failing, specifically with 404 errors.

PR #11, opened 2016-12-14
Zoom Control Improvement

waxenegger:
-) Resolution Levels have been refined to give more freedom to zoom.
-) Zoom Level Display (and Change) Field
-) There is a fit (into view extent) button
-) There is a 1:1 (full extent) button

Note:
Openlayers requires the definition of resolution levels beforehand as well as tiled retrieval. This means that zoom values that do not correspond to a resolution exactly snap to the nearest value. This, however, works for what are deemed small images as well as pyramidal data.

screenshot_2016-12-14_18-35-07


snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#15](https://ci.openmicroscopy.org/job/iviewer-merge/15/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/15/consoleText) for more details.
Possible conflicts:
  - Conflict detection failed
    - libs/ol3-viewer.js
  - PR #8 waxenegger 'prefixing ability for resource requests'
    - ol3-viewer/src/ome/ol3/globals.js

jburel:
--exclude

waxenegger:
@jburel I redid the commits to exclude the merge commits. 

with https://github.com/ome/omero-iviewer/pull/8 merged I think we can take off the exclude to see if merge-build works for this PR.

jburel:
My PR https://github.com/ome/omero-iviewer/pull/12 will conflict with yours since you made a change to ``libs/ol3-viewer.js``
Do you need to commit that change?
since everything will now be compiled as part of the install

waxenegger:
I suppose I can take it out. It shouldn't matter for the results of the merge build, in fact it has to go, as you said, if this PR is meant to be included. 

For future PRs that would need to compile ol3-viewer.js it would be great if we could give https://github.com/ome/omero-iviewer/pull/12 top priority next week (for review and merge) so that locally I won't always have to omit the compiled file (unless I merge it locally before it has become part of master which I'd rather not do).

jburel:
If I understand correctly, clicking on ``+/-`` bring to the most "suitable" zoom level 
it is not an increment/decrement by for example 5

waxenegger:
@jburel +/- is the next resolution/zoom up or down. it will use the predefined levels. the "suitable"/snapping happens only if you enter a value.

jburel:
Indicating that it is a percentage will be useful

waxenegger:
@jburel since I was using the space for the percent info I figured I could as well make it into a button so that when clicked the zoom value will be applied. that's at least as transparent as applying the zoom when the focus on the input field is lost. the mouse over title should help a bit as well...

jburel:
Clicking "enter" in the text field should apply the zoom level similar to the "%" button

waxenegger:
That's a good idea. Will add the handler.

jburel:
Entering the value now works
Very small point: (more a clarification)
looking at 
https://cowfish.openmicroscopy.org/webtrial/omero_iviewer/8152/?dataset=702
when I enter the value 145 %, it takes me to the "closest" value 142%
but when I click on ``+/-`` it takes me to 143% 
where does this diff come from?

waxenegger:
Interesting. I suspect a round/truncation discrepancy. Have to check that...

jburel:
zooming using the trackpad does not really.
The value is displayed is modified but the image "moves" but is not zoomed
This is probably for another PR

jburel:
The last commit fixes the display issue described above https://github.com/ome/omero-iviewer/pull/11#issuecomment-268226939

waxenegger:
I struggle to reproduce the trackpad problem. In general events are registered for an element and outside the ol3 viewer viewport zoom should not happen. I could not see any zoom or other reaction within the viewport when I used the trackpad on my laptop. 

if I'm within the viewport, then with the two finger gesture (up and down) I could zoom appropriately.

I wonder if the movement comes from the frame moving, perhaps a browser specific reaction to the trackpad interaction....

jburel:
I will keep an eye on it and add an entry on the card with more details.
The changes done in this PR work as expected. Merging

PR #13, opened 2016-12-16
style improvements

waxenegger:
This is a follow up for https://github.com/ome/omero-iviewer/pull/9.

It uses bootstrap styling for the buttons in the regions tab of the right hand panel, refactors some styling in the header and fixes a problem that slipped in with the previous PR. 

For future reviewing and comparison, this is what it looked like before:

old_style

Note: to be able to get to the roi tab, you'll have to check "Show ROIs" in the header section of the page.

### DIFF
--- a/src/app/header.html
+++ b/src/app/header.html
@@ -11,31 +11,30 @@
             </div>
         </div>
 
-        <div class="checkbox col-xs-6">
+        <div class="header-content col-xs-9">
             <button
-                disabled="disabled" class="btn-xs split_channels disabled-color"
+                disabled="disabled"
+                class="btn btn-default btn-sm split_channels disabled-color"
                 click.delegate="toggleSplitChannels()"
+                style="line-height: 20px;margin-bottom:5px"
                 value="normal">
                 Split Channels
             </button>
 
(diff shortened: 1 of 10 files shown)


jburel:
Much easier to read the text on the greyed out controls

jburel:
Something I noticed, not related to the PR,
When drawing "points", the user has to always go back the right-hand side to select it
This is not going to work when user just wants to track elements and click on cells for example

waxenegger:
Yes, the default behavior at the moment is that you select drawing a type of geometry, and once you've done so it leaves drawing mode. 

Are you suggesting we should make an exception for points only, or stay in drawing mode until there is an explicit decision to leave it?

jburel:
Currently in insight, if you are in ``ROI drawing mode``, for any shape you stay it that mode until the shape button is no longer selected

waxenegger:
ok, we can synchronize that behavior. but let's do that in a separate PR.

jburel:
as I said, I noticed it while looking at the changes. It does not need to be done in this PR.
I will update the card

jburel:
comment added to card. Merging this PR

PR #15, opened 2016-12-20
remain in draw mode until deselect or mode change

waxenegger:
To behave like Insight in regards to drawing we don't stop drawing after one successful draw of a shape.

Instead we remain in that mode until the icon is deselected by clicking on it again or  another edit mode is selected (translate or modify - from the edit dropdown box).

Test: Can be tested on cowfish. Click "Show ROIs" in header then go click right hand panel tab "Rois", then choose a shape (click) and start drawing. When finished you can continue drawing another one of the same shape. Draw mode will be aborted if the active shape icon is clicked again or the "Translate" or "Modify" option of the Edit dropdown menu is chosen.

Note: the drawing is slightly different in the iviewer in that it will want you to not hold down to initiate the drawing. you'll need to click only once in the spot you want to start. Release and move cursor to let the shape take form. When finished click again once.


jburel:
The feature works as described.
I need to review some of the controls e.g. select all and modify, this is out of the scope of this PR

waxenegger:
--exclude 

testing possible side-effect

dominikl:
Works like expected 👍  
Code looks reasonable, but I'm not familiar enough with the iViewer code to do a proper code review.

jburel:
why do we have ``libs/ol3-viewer.js`` removed again?

jburel:
I had some very strange behaviours when using undo/redo and clicking save 
I end up with more ROIs in the table that I have shapes on the images.
Also if I click, save the "undo/redo" history should probably become empty

jburel:
the issue with saving was due to that I was trying to save ellipses (handling of transforms)
This is not related to this PR

Discussed with @waxenegger

PR #16, opened 2016-12-21
Channel Settings: add colors to selection before luts list

waxenegger:
When selecting a color or lookup for a channel, the list that shows did so far only comprise the lookup entries, as well as an opion to open the color picker.

To be in synch with the standard web viewer, the colors red, green, blue, white, yellow, magenta, cyan have been added to the selection list prior to the luts entries.

luts


jburel:
Works as expected and matches the implementation of the current web viewer

PR #17, opened 2016-12-26
enable copy & paste of rnd settings for compatible images

waxenegger:
According to https://trello.com/c/HYnRRhE8/243-review-first-round the pasting of copied rendering settings should not be restricted to the same image only but instead to compatible ones whereby compatible is defined to be equal number of channels and data type/range.

This PR aims to remove that restriction, including reverse intensity into the copy & pasting as well.

Test: 

Choose a dataset that has compatible images (i.e. same number of channels + pixel range) and you should be able to paste after you have done a copy on one of the.

### DIFF
--- a/src/model/image_info.js
+++ b/src/model/image_info.js
@@ -321,6 +321,17 @@ export default class ImageInfo {
      * @memberof ImageInfo
      */
     requestImgRDef(callback = null) {
+        if (callback === null)
+            callback = (rdef) => {
+                if (rdef === null || typeof rdef.c !== 'string') return;
+                let channels = Misc.parseChannelParameters(rdef.c);
+                // we only allow copy and paste with same number of channels
+                // and compatible range
+                if (!Misc.isArray(channels) ||
+                        channels.length != this.channels.length ||
+                        rdef.pixel_range != this.range.join(":"))
+                            this.copied_img_rdef = null;
+            }
         $.ajax({
             url : this.context.server +
                     this.context.getPrefixedURI(WEBGATEWAY) +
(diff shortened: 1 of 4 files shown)


src/model/image_info.js
```diff
@@ -321,6 +321,17 @@ export default class ImageInfo {
      * @memberof ImageInfo
      */
     requestImgRDef(callback = null) {
+        if (callback === null)
+            callback = (rdef) => {
+                if (rdef === null || typeof rdef.c !== 'string') return;
```
jburel: Checks are done server-side to see if the images are "compatible"
see https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/server/src/ome/logic/RenderingSettingsImpl.java#L1090

waxenegger: ok, that's good to know that there is an additional safeguard.

the front-end logic does control the disabled state of the button though which I think is helpful so that the user does not think that paste makes sense and then waits for something to happen.

jburel:
Note that other checks are also done server-side e.g. make sure that the pixels type of source and target match

jburel:
The controls check works for channel

jburel:
As mentioned previously
checking the channels is only one of the check performed server-side
the pixels type is also checked for example so nothing will happen if the images are not compatible
It might be better to introduce a method ``isCompatible`` so we do not replicate the same logic in 2 different places

waxenegger:
ok, so I'm going to then take out the checks and the button is going to be active always with the backend request being sent regardless. A shared check routine (isCompatible) in the backend is even more of a round trip effectively doubling the http requests (in the allow case) so a request that results in nothing happening is preferable to a check request.

PR #18, opened 2016-12-27
Dimension slider additions

waxenegger:

Arrows were added for 1 increments. If focus is on slider use of arrow keys is possible too.

image


snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#46](https://ci.openmicroscopy.org/job/iviewer-merge/46/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/46/consoleText) for more details.
Possible conflicts:
  - PR #17 waxenegger 'enable copy & paste of rnd settings for compatible images'
    - src/settings/settings.html

jburel:
A "play-movie" arrow next to the slider will also be very useful
It has proved to be a very practical feature for the scientists
It does not have to be done in this PR

waxenegger:
Yep, I did think of that while doing this ;-)

jburel:
Works as expected, there is a "light" grey box around each arrow.
Is that intended?

Do you want to implement the "play" button in this PR?
I do not think we need two buttons per slider as it is currently in the web
One on the left for t and bottom for z should be enough

waxenegger:
@jburel we can take out the grey box. initially I thought it would add "something" to visibility, not too sure now.

let's do the play in a different PR, it's more complicated within ol3, I need to work with the eventing of the framework and not just vanilla image loads so I need to think this through a bit more ;-)

jburel:
ok let's do the "play" in a different PR
if you could take out the grey box, that will be good. Not convinced by the effect

waxenegger:
I agree, the background was not a good idea. Took it out.

jburel:
Looks good on cowfish.

jburel:
I just noticed one thing in the code (not in the PR description):
When the focus is on z slider: pressing the up/down arrow key changes the z-section
When the focus is on t slider: pressing left/right arrow key changes the timepoint

in that case, the next timepoint or z is displayed above the slider, similar to what happens when moving the slider's knob
I don't think this is necessary in that case

waxenegger:
I'm not sure I can follow. My code does not have an impact on the display of the slider per se. The handle will always appear at the very edge of the bar when the min/max is hit. That's a visual. Or are you suggesting that both sliders are moved at a keystroke, or that a min/max is exceeded.

jburel:
I am not referring to the value next to slider e.g. T 5/30
We have 3 cases:

1) When you move the slider knob: the "value" set when the knob is released is displayed above the slider. This is useful when moving the knob
2) When changing the value using the arrow next to slider: this is not displayed. This is fine since we either increase/decrease by one
3) When changing the value using the "keys", the value is also increased or decreased by one but the "next" value is displayed above the slider e.g. ``T: 6``. It should match case 2)

waxenegger:
Ah, apologies, I see what you mean. You meant that temporary display of where you'd go if you used the handle. ok. yes I'm going to look into that.

jburel:
the temporary value is no longer displayed
If I keep one of the arrow key down, the slider's knobs moves very slowly and the value is changed when I stop pressing the key. I think that's acceptable

waxenegger:
Funny, the last one I should have noticed and the latest commit should take care of it.

jburel:
Much better with the last commit. This improves the experience. Thanks @waxenegger 
Merging now

PR #19, opened 2016-12-30
dimension player

waxenegger:
follow up on: https://github.com/ome/omero-iviewer/pull/18

This PR adds the feature that z/t can be played in a forwards/backwards fashion like the present web viewer.


jburel:
The movie player works as expected.
The icons are a bit small so it is difficult to see, especially when it is in the "play mode". The same icons are used in the actual web viewer but there is a border around the button.

jburel:
The icons are much nicer. Thanks

PR #20, opened 2017-01-03
trello requested fixes for region edit constraints and cursor appearance

waxenegger:

This PR addresses the following:

1) A 10px upper limit on the stroke width was removed
2) A 1px lower limit on the stroke width was changed to 0px so that any filled geometry (lines exl.!!) can have no stroke
3) The font size min bound was lowered to 1, the upper limit of 1000 removed
4) There were supposed browser interpretations of the 'col-resize' cursor style as 4 way arrows such as  IE which was replaced by 'ew-resize' which seems to show the horizontal 2 way arrow always.

How to Test: 

Choose an image with regions (on cowfish) and try setting the stroke width and font size to values that were not allowed prior to this PR. For the mouse cursor (4) you need to move the mouse over the the edge of the left hand or right hand panel and the 2 arrow cursor should appear. 
Note: the cursor when hovering over the collapse icon will look different.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -7,7 +7,7 @@
 
 	<property name="version.major" value="1" />
 	<property name="version.minor" value="0" />
-    <property name="ol3-version" value="3.20.1" />
+    <property name="ol3-version" value="3.19.1" />
 
 	<property name="closure-compiler.jar" value="${basedir}/closure/compiler.jar" />
 	<property name="closure-calcdeps.py" value="${basedir}/closure/calcdeps.py" />
--- a/build.xml
+++ b/build.xml
@@ -7,7 +7,7 @@
 
 	<property name="version.major" value="1" />
 	<property name="version.minor" value="0" />
-    <property name="ol3-version" value="3.19.1" />
+    <property name="ol3-version" value="3.20.1" />
 
 	<property name="closure-compiler.jar" value="${basedir}/closure/compiler.jar" />
 	<property name="closure-calcdeps.py" value="${basedir}/closure/calcdeps.py" />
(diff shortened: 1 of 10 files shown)


bramalingam:
The testing instructions are clear and all of them work. But identified one bug and one RFE (related to this PR).

1) BUG:  Change the stroke width for any ROI (increase to any value other than one), and then change the stroke color/fill color. This will reset the value of the stroke width on the viewer to 1 but the right hand-side panel will still show the previous value (that was set by the user). 

2) RFE: Visually am not able to see the difference between a stroke width 1 and 0 (unless I click away on another ROI). This is not a very intuitive workflow, when a user sets stroke width to 0, can you throw a tooltip, suggesting a requirement for a fill color to view the ROI appropriately? Given that the region of pixels could be any color and selecting a default fill color might in some cases not show the ROI at all. Please let me know if this explanation is not clear. @jburel @waxenegger

waxenegger:
I'm going to fix 1 soon.

Regarding 2: that's a question of whether selected shapes should be marked as selected in some way and I'd argue that they should. for translates, multiples selects and modifies of geometry it's quite essential that the user know which ones s/he has selected. The discrepancy arises because when I change the stroke width the "selection style" reflects that generally so therefore you'd expect to see no line when you set it to 0 (I fully understand the reasoning). Problem is then I loose my way of knowing what is selected. Insight I believe does not change the selection style at all so at least there is consistency and perhaps that's a possible way to go. Sure, a user could potentially "hide" a shape by making both fill fully transparent and setting the stroke 0. I mean there is always the listing in the right hand panel by which means one can locate and select the shape because unless one forbids full transparency that is always an option. Insight lets you do so for instance.

bramalingam:
Works as expected (The change in stroke color or fill color does not reset the stroke-width value to 1 in the viewer). Looks good to merge.

will-moore:
I agree: I think when a shape with stoke: 0 is selected, you have to show a line of 1 since the dashed line is the only way of telling what's selected.
In Insight or Figure, we have handles to show what's selected but we're using a different model for iViewer.
NB: not for this PR, but we might want to explore other options, for example, when the line is thicker it's hard to see that it's dashed (selected):

screen shot 2017-01-30 at 11 21 12

jburel:
https://trello.com/c/yyI19Pus/265-shapes-handling for further discussion

PR #21, opened 2017-01-06
bugfix: undo of the deletion of newly drawn shapes

waxenegger:

The deletion of newly drawn shapes was handled slightly differently which lead to a bug in the undo functionality (part of history). In particular a newly created shape that was modified (spatially/geometry/style) and then deleted would be recreated in its initial state on clicking "undo" (as opposed to its most recent/modified state before deletion).

Test: Create a new shape and modify it by moving it or manipulating its vertices/edges (right panel edit dropdown menu offers switch between translate/modify) and changing its fill/stroke color for instance. Then delete it via the delete button in the right hand panel (you could use either delete selected or all, doesn't matter). After clicking the undo button, everything should go back to the state before deletion and  you should be able to undo the individual changes as well

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1457,7 +1457,9 @@ ome.ol3.Viewer.prototype.getSmallestViewExtent = function() {
     if (ome.ol3.utils.Misc.isArray(ids) && ids.length > 0 && this.regions_.idIndex_) {
         col = new ol.Collection();
         for (var id in ids)
-            if (this.regions_.idIndex_[ids[id]] instanceof ol.Feature)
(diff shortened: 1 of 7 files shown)


snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#63](https://ci.openmicroscopy.org/job/iviewer-merge/63/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/63/consoleText) for more details.
Possible conflicts:
  - PR #15 waxenegger 'remain in draw mode until deselect or mode change'
    - ol3-viewer/src/ome/ol3/source/Regions.js
    - src/regions/regions-drawing-mode.js

waxenegger:
--exclude until conflicting PR was merged

jburel:
The conflicting PR has been merged

dominikl:
Works like expected 👍

PR #22, opened 2017-01-07
Fix path

jburel:
Fix the path to the folder to check during build process

test 1:
 * delete the ``plugin/omero_iviewer/static``
 * Run ``python setup.py sdist`` from the plugin directory
 * This should run the full build
 * Run it again. It should not build

test 2:
Install from testpypi
``pip install --upgrade -i https://testpypi.python.org/pypi omero-iviewer``
This should install and not build anything (the version is 0.0.7 on testpypi, this is not important)


dominikl:
Works like expected 👍

PR #23, opened 2017-01-10
improved storage routine

waxenegger:
Some improvements to the routine that marshals and stores the regions json:

-) fix bug when multiple shapes were within one roi
-) better check for roi removal when empty
-) exception mid-way returns successfully processed ids as well as error

On an aside: The storage routine that for historic reason resided in plugin/ol3-viewer as well was removed.

TEST: cowfish.openmicroscopy.org/webtrial/

Select and modify shapes, create new ones and delete existing ones (all via right hand panel titled Rois), then save. 
Check that the list of shapes updates appropriately and, just to be absolutely sure, reload the same image and double check if the changes have been persisted.

Update on testing: 
-) there will be a message box coming up now when you hit save. 
-) if you click on a thumbnail in the left-hand panel (and you have unsaved rois changes) you'll receive a confirmation dialog if you want to save.
-) if you want to close the window and you have unsaved rois changes the default browser onbeforeunload alert will pop up


snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#71](https://ci.openmicroscopy.org/job/iviewer-merge/71/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/71/consoleText) for more details.
Possible conflicts:
  - PR #20 waxenegger 'trello requested fixes for region edit constraints and cursor appearance'
    - src/regions/regions.html

will-moore:
I created a bunch of shapes and clicked the Save button (saw no reaction) and clicked it again. It was still enabled so everything got saved again and now I've got [duplicates of every shape](https://cowfish.openmicroscopy.org/webtrial/omero_iviewer/56754/?dataset_id=17803)

Deleting shapes works fine and I didn't see any problems with multiple shapes in an ROI (although see other comments below).

I selected a Rectangle (already saved) and chose "Attach to all Z". The "Propagate Selected Shapes" button is not very obvious (maybe move it up) and when clicked I don't see any effect (since the added shapes are at the bottom of the table and not visible). Clicking the button again repeats the propagation, so after Save I now have 2 shapes on every plane (except the original plane of the shape).
This can be repeated many times with any shape leading to massive proliferation of shapes. Maybe need to discuss strategy for this since it's quite different from the "Propagate" dialog in Insight that only lets you propagate to planes that don't yet have any shape from that ROI.
Also, I don't see anything to indicate that all the shapes are under the same ROI?
It's hard to "undo" a propagation after Saving because I can't select multiple shapes in the table (e.g. shift-click to select range) to delete multiple at once.


Minor issues (not this PR?):
 - With the Arrow tool, I'd prefer if this defaulted to drawing a single line, not a Polyline. I didn't support PolyLine Arrows in webclient image viewer (for tech reasons) and we decided to disable them in Insight for consistency and mostly because the use of PolyLine Arrows is very edge case (never seen this). In iViewer I keep drawing PolyLine Arrows when I really just want a singe line Arrow.
 - Clicking on a Point in the table zooms the image in to 4000%. Maybe 200% would be better?
 - In fact, for all shapes, when zooming into to show the shape, maybe the zoom should be 1/3 off the current zoom, so instead of the shape filling the whole viewport, it fills 1/3 or 1/4. This puts the shape in "context" of it's surrounds and is a more useful working distance. 
 - After saving a point, I click to select it, then draw some more points and they show up bigger on the image. After saving them and refreshing the image they shrink to the normal size again.
screen shot 2017-02-09 at 11 21 07
 - When I'm in the "draw new shape" mode, I find that I can't select a shape from the table (took me a while to work out why this wasn't working). I see that when I've selected a shape *then* select a drawing tool, it automatically switches from drawing mode to selection mode, so it would be nice if this worked the other way around: Clicking on a shape in the table selects it and switches to "selection" mode.
 - The "Propagate" and "Attach to viewed Z and T" are check boxes but I can't uncheck both of them at the same time (to allow a shape to be Unattached) which would be a useful feature and is supported by the model (Z and T null for shape).
 - It would be nice if the shapes table scrolled to automatically show the selected shape (when selected by clicking on the image).

waxenegger:
@will-moore can I please ask to put observations that do not relate to this PR (storing of rois) in the respective trello card. These comments span from storing (one remark about being able to click 2 times which clearly needs fixing) over the roi listing to to propagation, drawing and drawing of polylines with arrows.

There are suggestions I can make against the double save which is to disable the button immediately, pop up a modal dialog, or both.

waxenegger:
With the need to have a modal dialog/message box as well as refactoring needed and done by https://github.com/ome/omero-iviewer/pull/35 I'm going to cherry-pick the commits and integrate them into this PR.

will-moore:
I don't think you need a dialog/message to prevent the double save. Just disable the Save button as soon as it's clicked.

Apologies for all the comments. When seeing a bunch of issues it's sometimes hard to sort them until after writing them all down somewhere first. 
Also wasn't sure exactly what was relevant to this PR.
I'll move comments to a (new or existing?) trello card.

waxenegger:
Disabling the button would prevent the double save but the feedback message is more user-friendly and removes all doubt as to what is happening and when it's done. 

The notification dialog is a feature that is potentially useful on other occasions.

I believe all of the comments would be rois/shape related so https://trello.com/c/yyI19Pus/265-shapes-handling is probably a good location. If in doubt create a new card under iViewer I guess.

I made some comments to address some of what you said.

will-moore:
Modal popup "Saving Regions. Please Wait..." is working fine.
But it is quite "in your face". Is it essential to prevent the user doing *anything* while Save is in progress? Or would it be sufficient simply to prevent them clicking Save a second time? That's what we did in figure to prevent the double save: https://github.com/ome/omero-figure/pull/161

waxenegger:
Having revisited the workflow I would recommend the dialog. 

Due to the asynchronous nature of storage and the necessary resync of ui - we are in a one-page app- there are actually places where not waiting could potentially hurt us in the future. The listing on the right-hand is such a case if the user continues to do things and in between the listing is updated. At the moment such a possibility is rather slim for technical reasons/implementation but taking into account a more complex widget in the future (i.e. filter, sort, right-click context, hierarchy) it can more easily effect it. 

I had to also adjust some code in the case of saving and switching images since the async code execution can be anywhere, that is before the success handler of the saving fires, or after in which case we must not have an update happen since we are in a new image context already.

will-moore:
OK, fair enough.
I think this is good to merge now.
I'll move comments to trello...

jburel:
Merging
@will-moore will take care of migrating the comments to a card

PR #25, opened 2017-01-24
retrieve untiled images in one go if dimensions don't exceed 1kx1k

waxenegger:
So far every image (with the exception of the split view images) has been retrieved in a tiled fashion.
This is not necessary and probably overhead for images that do not exceed a certain size.

Therefore, any image with dimensions x times y that fall under a pixel limit of 1000x1000 will be fetched in its entirety and with a single request (tile if you will).

Test: on cowfish (https://cowfish.openmicroscopy.org/webtrial)
 
1) Open up a pyramidal image for example and check that it is retrieved as tiles still. 
2) Open another image (untiled format) and check that it is fetched in its entirety (if under the limit)

Note: Chrome / Firefox / IE / Opera have developers tools that can be opened by pressing F12 (once the viewer window is open!). The tools have network (requests) tabs that list the type of requests made. You can filter the view to show images only. In case 1 you should see the individual tiles of the mosaic i.e. a list of images, in case 2 it should be only 1 request. Of course this relates to the actual image, not any other images in the page such as icon requests.

image

image


ol3-viewer/src/ome/ol3/source/Image.js
```diff
@@ -142,7 +142,21 @@ ome.ol3.source.Image = function(options) {
  	 * @type {boolean}
  	 * @private
  	 */
-     this.tiled_ = this.resolutions_.length > 1;
+     this.tiled_ = opts.tiled;
+
+     /**
+  	 * should we use tiled retrieval methods?
```
jburel: some indent issue

jburel:
@waxenegger  is this PR ready for review?

waxenegger:
I think it makes sense to wait with this PR a bit longer, one reason is that I suspect that the "1 tile" approach causes a blank center panel on rare occasions (only initial load). This seems to be more of an openlayers related thing and I believe it has to do with the size of the viewport as well. Either way I'd like to make sure that it does not happen and I have never seen it happen with the tiled approach.

waxenegger:
I believe my latest commit fixed the issue but I'd still like to have this one open for a bit longer. If the blank load behavior does not repeat in the next view days I'm going to list the PR for review.

waxenegger:
--exclude

will-moore:
Seems to be working fine.
Loading tiles / single planes as expected and no issues with image failing to load.
Looks good to merge.

jburel:
One RFE discussed with @waxenegger today: collapse the bird eye view for untiled image
i will add a note to a trello card.
Merging this PR

PR #26, opened 2017-01-24
argb -> rgba

waxenegger:
This PR follows suit with https://github.com/openmicroscopy/openmicroscopy/pull/4948 and https://github.com/openmicroscopy/openmicroscopy/pull/4995 which got merged more recently.

Instead of using ARGB color encoding we want to use RGBA.

Along with the corresponding changes for the js-unit tests, I adjusted the assert syntax in those tests.

Test: While there are js-unit tests (see run below) and you can rerun them locally, the best way to test this is on cowfish, against the present web viewer and against insight.

image


joshmoore:
NB: nice unit test feature

waxenegger:
Thank you. Unfortunately phantom.js also allowed me to run the tests command line style, well the output was there as well I should say, and not just html (though the report is nice enough) but when I shifted things around that stopped working. Annoys me a bit, and I hope I can find the reason and fix it.

jburel:
The color is correctly displayed,
We will need to run the unit tests as part of "travis build"
This should be done in another PR.
Merging

PR #28, opened 2017-01-31
left-hand thumbnail panel empty when opened image belongs to multiple datasets

waxenegger:
I changed the description of this PR to narrow down its scope and avoid confusion:

It was noted in trello ( https://trello.com/c/HYnRRhE8/243-review-first-round ) that an image that was opened with the iviewer AND belongs to more than 1 dataset does not show the thumbnails of the other images which is due to the dataset ids not being supplied in the imgData when there are multiple.

It has become apparent to me that this is not an easy thing to remedy since in the case of multiple datasets one needs to know which of the image/dataset combination in the tree was clicked on, not to mention multiple selections as @jburel points outs in his comment below.

Therefore the only way for this to work is to steal the idea in https://github.com/ome/omero-iviewer/pull/7 and use open_with.

TEST: cowfish.openmicroscopy.org/webtrial/

Use an image that belongs to 2 datasets, or link an image to another. Then right-click on the image in the tree and open with iviewer. The iviewer window should show a left hand panel that displays the images from the dataset that was the parent of the opened image. Make sure you repeat this procedure for the other dataset/parent.

### DIFF
--- a/src/utils/constants.js
+++ b/src/utils/constants.js
@@ -40,7 +40,7 @@ export const WEBCLIENT = "WEBCLIENT";
  * @type {Object}
  */
 export const REQUEST_PARAMS = {
-    DATASET : 'DATASET',
+    DATASET_ID : 'DATASET_ID',
     SERVER : 'SERVER',
     IMAGE_ID : 'IMAGE_ID',
     CHANNELS : 'C',
--- a/src/utils/constants.js
+++ b/src/utils/constants.js
@@ -40,6 +40,7 @@ export const WEBCLIENT = "WEBCLIENT";
  * @type {Object}
  */
 export const REQUEST_PARAMS = {
+    DATASET : 'DATASET',
     SERVER : 'SERVER',
     IMAGE_ID : 'IMAGE_ID',
     CHANNELS : 'C',
--- a/src/utils/constants.js
+++ b/src/utils/constants.js
@@ -40,7 +40,7 @@ export const WEBCLIENT = "WEBCLIENT";
  * @type {Object}
  */
 export const REQUEST_PARAMS = {
-    DATASET_ID : 'DATASET_ID',
+    DATASET_ID : 'DATASET',
     SERVER : 'SERVER',
     IMAGE_ID : 'IMAGE_ID',
     CHANNELS : 'C',
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -240,10 +240,11 @@ export default class Context {
      * the same image to be used in multiple ImageConfigs.
      *
      * @memberof Context
-     * @param {number} image_id the image id of the server
+     * @param {number} image_id the image id
+     * @param {number} dataset_id an optional dataset_id
      * @return {ImageConfig} an ImageConfig object
      */
-    addImageConfig(image_id) {
+    addImageConfig(image_id, dataset_id) {
         if (typeof image_id !== 'number' || image_id < 0)
             return null;
 
@@ -252,7 +253,7 @@ export default class Context {
             for (let [id, conf] of this.image_configs)
                 this.removeImageConfig(id,conf)
 
-        let image_config = new ImageConfig(this, image_id);
+        let image_config = new ImageConfig(this, image_id, dataset_id);
         // store the image config in the map and make it the selected one
         this.image_configs.set(image_config.id, image_config);
         this.selectConfig(image_config.id);
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -95,7 +95,50 @@ export default class Context {
         if (typeof eventbus instanceof EventAggregator)
             throw "Invalid EventAggregator given!"
 
-        let server = optParams[REQUEST_PARAMS.SERVER];
+        // process request params and assign members
+        this.processServerParameter(optParams);
+        this.readPrefixedURIs(optParams);
+        this.eventbus = eventbus;
+        this.initParams = optParams;
+
+        // we set the initial image as the default (if given)
+        let initial_dataset_id =
+            parseInt(
+                this.getInitialRequestParam(REQUEST_PARAMS.DATASET_ID));
+        let initial_image_config =
+            this.addImageConfig(initial_image_id, initial_dataset_id);
+        this.selected_config = initial_image_config.id;
+
+        // set up key listener
+        this.establishKeyDownListener();
+
+        if (this.hasHTML5HistoryFeatures()) {
+            window.onpopstate = (e) => {
+                if (e.state === null) window.history.go(0);
+                this.addImageConfig(e.state.image_id, e.state.dataset_id);
+            };
+        }
+    }
+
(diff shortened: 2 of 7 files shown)


waxenegger:
--exclude

src/utils/constants.js
```diff
@@ -40,6 +40,7 @@ export const WEBCLIENT = "WEBCLIENT";
  * @type {Object}
  */
 export const REQUEST_PARAMS = {
+    DATASET : 'DATASET',
```
joshmoore: `DATASET_ID` maybe?

joshmoore:
I don't know much about what's going on here, but I have a slight reaction that this is overly specific for the long-term. I.e. what about SPW? Folders? Etc.

src/utils/constants.js
waxenegger: I guess you are saying I'm not correct enough in my naming there ;-) 

sure, no problem, I can change the constant name only or both, in the former case there is only one other adjustment, in the later I'd have to manipulate the open_with script as well. It's the one that directly adds the query parameter as "dataset=..." and the constant value has to match that.

waxenegger:
well, this is merely for datasets. open_with enters into it because it does me the favor of supplying exactly that, a dataset id. I have no other good way since I bootstrap from an image. it's really all about images that belong to more than one dataset and the thumbnails of that associated dataset are then displayed in the left hand panel of the iviewer.

jburel:
it might be time to introduce a type.
Also we may want to think of adding support for multiple objects i.e. select 2 images.
I don't want to have the full context e.g. dataset in the viewer 
Let's try to flash that out before we move to far down the line.

waxenegger:
I think, maybe, too much weight is put into this PR. It is not so much a different/enhanced way of bootstrapping as a possible 'fix' to put the right thumbnails on the left side.

I do admit that I have no good way of determining that if I don't use open_with but at least for that case it would hand me in the right dataset. 

The opening of two images in one iviewer window would necessitate quite a bit more thought and work, the ui arranging of things being a problem but not nearly as much as the technical side. one would have to switch to a kind of multi window within the window in that case unless one tried to pop up 2 windows simply...

waxenegger:
I have used ever so slightly modified code from: https://github.com/ome/omero-iviewer/pull/7

This means that, using open_width the panel won't be empty any more. Given that the iviewer, in the short run, will be used in that configuration I think it's at least a fix.

@joshmoore, @jburel 

I see where you are coming from and I realize that this is not the solution to selecting and opening multiple things. 

The latter is going to require quite a bit more discussion and substantial refactoring though if, in fact, multiple images should be shown in one window, kind of like mdi. If not either an opening method would have to pop up more than one window or the iviewer window would have to do so. That raises questions about interaction issues between them (unless there should be none) which can become tricky in browsers. It also goes into urls in the address bar. Sure, HTML5 history does give us improved ways of modifying it as opposed to the old hash tag solution which was not so great and anything else in the path being updated would trigger a reload.

In short, I'm not so clear on how all of this can fit together in the web. I believe the idea should come first, then one can check for technical solutions to make that possible. Concretely, I guess, the UI should be contemplated and desired interaction. If, say, people want an mdi kind of thing, then I can look into that. If they'd want multiple windows communicating with each other one can investigate that. 

The design work by gus was very much in the spirit of everything within one window, sure there were multiple bits like tools and panels but they did relate to one image. To move toward multiple images in one window (I don't mean just thumbnails) with tools and panels relating to them is a bit of a technical paradigm shift. The implications for things like multiple selection and tool use, or even single selection and things such as history and saving are enormous. Do I lump them together and differentiate by artificial ids, do I keep a separate copy in memory. Which actions can be done on multiple images, which can't, just to list some.

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#95](https://ci.openmicroscopy.org/job/iviewer-merge/95/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/95/consoleText) for more details.
Possible conflicts:
  - PR #7 will-moore 'open with parent'
    - src/openwith.js

waxenegger:
ah, missed that the other PR is not on hold any more. ok, no problem, I can --exclude this one until the other one has been merged.

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#96](https://ci.openmicroscopy.org/job/iviewer-merge/96/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/96/consoleText) for more details.
Possible conflicts:
  - PR #7 will-moore 'open with parent'
    - src/openwith.js

waxenegger:
--exclude

will-moore:
This is working pretty well for me with ```http://web-dev-merge.openmicroscopy.org/omero_iviewer/9470/?dataset_id=1460``` even when I open iViewer without dataset in the query. Clicking on thumbnails of images in 2 Datasets doesn't break the list of thumbnails. I wonder if you want to use ```?dataset=1460``` instead of ```?dataset_id```. This is the convention we're using in /api/.
NB: Sometimes when I click through the thumbnails in iViewer, the selected image doesn't appear in the viewer (right panel, Z/T sliders etc all load but no image plane). Clicking the thumbnail to re-load the image seems to fix it.

jburel:
Discussed with @waxenegger yesterday, we may want to introduce ``type and id`` 
so we can then load for example all fields of a well in the full viewer.
I think it will make sense to follow the convention introduced in the api

waxenegger:
@jburel, @will-moore 

I'm going to change back to the dataset parameter. Initially I changed it after josh's code review question. I'm going to use DATASET_ID internally (as a constant) since it's more to the point but dataset for the query string since that's consistent with api usage.

@will-moore: the initial load of the image problem did happen to me too BUT it's not this because of this PR. I'm 99.9% sure it's https://github.com/ome/omero-iviewer/pull/25.

I'm going to look into it. It happens very sporadically, never happend for "tiled" retrieval/pyramids and I "believe" that it happens more so when the server's busy, and/or traffic is slow. Also, another reason could be the size of the window and internal loading of tiles. Even more so I suspect that due to the bigger size of the 1-tile approach and the smaller window size, openlayers may wrongly lump it into the off-screen category. If, for instance, you try and drag the image  (that you don't yet see) with the mouse  (instead of clicking the thumbnail) it does show (as part of the next rerender cycle).

will-moore:
The 'dataset' parameter looks fine and everything works when images are in multiple datasets. E.g. https://cowfish.openmicroscopy.org/webtrial/omero_iviewer/56755/?dataset=17803
Good to merge.

waxenegger:
@jburel it would be great if this one could be merged as soon as possible. unfortunately it clashes with on e other already and will be having a conflict itself if I make a change in another PR.

jburel:
merging, I will add a card to capture the discussion i.e. support other types e.g. wells

waxenegger:
Thanks.

PR #30, opened 2017-02-02
let user enter channel start/end values outside of min/max but inside of range limits

waxenegger:
Before this PR iViewer, when in min/max "mode", start/end values for channels outside of the set min/max were not allowed. To go below/above one had to switch to "full range" with the slider spanning the entire range.

To exhibit behavior like in the present web viewer as well as figure these restrictions were removed.

TEST:

Deployed on cowfish, e.g: https://cowfish.openmicroscopy.org/webtrial/omero_iviewer/9442/?dataset_id=1457 (user-1) 

Choose an appropriate test image, ideally with a larger range (16 bit say), and some smaller min/max -values. By default iViewer is in min/max 'mode'.

You should now be able to enter start/end values for the respective channels that are outside the original min/max (but still within the full range limits). Check that the range slider adjusts correspondingly after expanding the range and using it gives sensible values within the range. Likewise if the range is compacted the slider adjusts, yet it should not get "smaller" than the min/max that's set.

### DIFF
--- a/src/settings/channel-range.js
+++ b/src/settings/channel-range.js
@@ -122,18 +122,8 @@ export default class ChannelRange  {
                     let imgInf =
                         this.context.getSelectedImageConfig().image_info;
                     imgInf.initial_values = true;
-                    // we need to distinguish for cases that exceed the
-                    // min/max view
-                    let fRange = imgInf.needsFullRange();
-                    if (fRange) {
-                        this.mode = CHANNEL_SETTINGS_MODE.FULL_RANGE;
-                        this.change_mode({mode: this.mode, fullrange: true});
-                    } else if (!fRange &&
-                        this.mode !== CHANNEL_SETTINGS_MODE.MIN_MAX) {
-                        this.mode = CHANNEL_SETTINGS_MODE.MIN_MAX;
-                        this.change_mode({mode: this.mode, fullrange: true});
-                    }
-                    this.updateUI();}));
+                    this.updateUI();
+                }));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


will-moore:
The sliders seem to be behaving as expected but when I stop sliding the image is not re-rendered and when I Save the settings and refresh the page, the change in levels is ignored.
Also it's quite nice to see the input numbers updating while sliding the slider, so you know where you're sliding to (doesn't have to be fixed in this PR, but it seems very closely related).

waxenegger:
That's the codomain url code that's deployed. I see that the request takes place in the network tab but I use the old format and it doesn't seem to like that because the moment I take out the -r/r it gives me the adjusted image.

see: https://cowfish.openmicroscopy.org/webtrial/webgateway/render_image/9443/14/0/?c=1|158:197$0000FF,2|96:124$FF0000&m=c&p=normal&q=0.9

as opposed to: https://cowfish.openmicroscopy.org/webtrial/webgateway/render_image/9443/14/0/?c=1|158:197-r$0000FF,2|96:124-r$FF0000&m=c&p=normal&q=0.9

waxenegger:
Ok, the input value should now update with the slider position, and images should update as well.

will-moore:
This is working nicely.
Minor issue: I noticed that the channel slider updates and image is re-rendered on every key-stroke in the text inputs, which is quite aggressive. E.g. If I type in "3000" the image will be rendered 4 times. Maybe only update on 'enter' or when the input loses focus ('blur')?

waxenegger:
@will-moore  yes that's been bugging me a bit too. I'm going to commit some code for it to only do it on blur and enter.

for review and completeness sake I should mention that I https://github.com/ome/omero-iviewer/pull/25 is excluded at the moment so that smaller images will be retrieved in a tiled fashion. That confused me the other day since I'd gotten used to only seeing 1 request on cowfish.

will-moore:
Looks good thanks.
Good to merge.

src/settings/channel-range.js
```diff
@@ -267,51 +257,71 @@ export default class ChannelRange  {
...
+                    "value", ui.values[0]);
+                $(this.element).find(".channel-end").spinner(
+                    "value", ui.values[1]);
+
+                let imgConf = this.context.getSelectedImageConfig();
+                if (!imgConf.image_info.has_histogram) return true;
+
+                // we want to update the histogram on slide so we
+                // need a separate event. we throttle so that
+                // we send only the lastest slider value within a 100ms
```
jburel: shouldn't be ``latest`` or just write  ``last``

jburel:
I can end up with an empty text field. This should not be allowed.
This results in no image displayed in the viewer
see screenshot
iviewer-slider

waxenegger:
@jburel could you please detail how you got an empty input...

was it the slider movement or you deleting the value and then losing focus?

jburel:
will try again today

waxenegger:
in case we cannot reproduce it, I do have a hunch what else it could have been. It has to do with the session expiring, and I have seen it when I work in the webdev setup or any setup where I connect directly to the iviewer opening the file in the url so that omero web is not there (which sends the keep-alive pings). then changing channel settings or anything that will result in a rerender of the image gives a blank because the image request fails. I wonder if you had closed down omero web while still having the iviewer window open, and then after sufficient inactivity tried changing start/end values which led to the issue.

jburel:
I cannot reproduce it. I did not close OMERO.web
I will keep an eye
The last commit reset correctly the value when the focus is lost
Merging

PR #31, opened 2017-02-02
Run unit test

jburel:
Run the unit test in travis

Check the output of the travis run. Tests should have been run and travis should be green

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,4 @@ script:
   - python setup.py sdist install --user || exit 1
   - if [ ! -d "omero_iviewer/static" ]; then exit 1 ;fi
   - if [ ! -d "omero_iviewer/templates" ]; then exit 1 ;fi
+  - ant unit-tests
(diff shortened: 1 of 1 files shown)


waxenegger:
Just a remark because I thought of it:

I'm probably going to do some work in regards to not necessarily stretch the travis run. As is the ant task "unit-tests" should not depend on build-all which is overkill. Thatincludes the build of the compressed ol3-viewer yet again which is something that "npm run prod" already does and, really, the tests use the debug version only. So I'm thinking that a depend on "build-debug" then "build-all" should actually be much better but I'll still need to test that fully.

jburel:
I agree that it is a bit overkill in its current form.
I can close this PR and we can include it in Travis after refactoring

waxenegger:
@jburel we don't really need to close this PR. I have opened a new one: https://github.com/ome/omero-iviewer/pull/33 with just that tiny build change.

jburel:
Closing this PR 
The change will be made in
https://github.com/ome/omero-iviewer/pull/33

PR #32, opened 2017-02-03
fix reversed start vs end marker interpretation

waxenegger:
Direction of arrows were reversed compared to insight. This PR aims at fixing that.

Test: 

Create arrows (aka lines with either end or start marker) in both insight and the iviewer, save them, then compare respectively. 

Both tools have a short cut for drawing arrows: insight: dashed arrow in measurment tools, iviewer: arrow in roi panel. If you'd like to create a start decoration you can do so in insight via the measurment tools again and changing the start decoration in the Inspector tab. In the iviewer the corresponding action is a dropdown menu in the edit section denoted <-->. see screenshots for where to find functionality: 

image

image


will-moore:
Arrows are now a single Line by default and appear same in webgateway viewer as in iViewer.
I'm trying to think of a nicer way to toggle the arrow heads since the ```<-->``` button is not really intuitive at present (you need screenshots and instructions above to know about it).
Maybe an improved icon (like the Arrow tool icon but with 2 heads?) and a tooltip on the button "Toggle arrow heads" would be enough. Even nicer if the 'Start' and 'End' options could give some clue as to the current state of the toggle, but not essential.

will-moore:
Last commit works fine, but it's a bit confusing to have Start/End scored out to indicate they don't have an arrow head. It's more normal for 'checkbox menus' to have a tick beside enabled items. E.g. can be unicode:
✓ Start 
✓ End
Any luck with icon and tooltip?

waxenegger:
Sounds good, the icon I'd have to hunt for though...

will-moore:
I would just photoshop your draw_arrow.png so it has arrow heads both ends. I can do it in a sec if you want? In fact, here's one I just did...

draw_double_arrow

will-moore:
Looks great. Working as expected. Good to merge.

will-moore:
@waxenegger In case you're wondering what to work on next, for me the thing that bothers me the most is you can't click on ROIs tab - have to click 'show ROIs' first. Clicking on the ROIs tab should auto-show ROIs. And within the ROIs tab you need a "Hide ROIs ```X```" somewhere that does the same as currently unchecking the "show ROIs". Maybe a ```X``` in the top-right of the panel (similar to github comment panels above) with tooltip "Hide ROIs". Or maybe a whole new row at the top with a clickable ```Hide ROIs```?

waxenegger:
Well, showing rois by default is not a problem at all. So, about "Hide ROI X", you mean shapes that are grouped in a roi I guess because toggling individual shape visibility is possible already (in the table listing):

not showing shape

PR #33, opened 2017-02-06
unit tests do not require a build-all

waxenegger:
The jsunit tests are at the moment depending on the target "build-all" which is overkill, especially for travis runs.

As is the scope of the tests rely on the target "build-debug" only.

TEST:  

Clone the project locally, cd to the root, then execute ant unit-tests. There is no need to wipe the entire git project if you have it already cloned. Instead delete the directory _node_modules_ (rm -rf node_modules) so that the dependency fetching is tested as well (the reason why "build-all" was listed a dependency to begin with - build-debug does so as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,3 +18,5 @@ script:
   - python setup.py sdist install --user || exit 1
   - if [ ! -d "omero_iviewer/static" ]; then exit 1 ;fi
   - if [ ! -d "omero_iviewer/templates" ]; then exit 1 ;fi
+  - cd ..
+  - ant unit-tests-only
--- a/build.xml
+++ b/build.xml
@@ -174,7 +174,7 @@
         </exec>
 	</target>
 
-	<target name="unit-tests" depends="build-all" description="runs unit tests" >
+	<target name="unit-tests" depends="build-debug" description="runs unit tests" >
         <echo>Fetching test libraries ....</echo>
         <exec dir="${basedir}" executable="npm">
 			<arg line="install mocha chai mocha-phantomjs"/>
--- a/build.xml
+++ b/build.xml
@@ -209,9 +209,9 @@
     </target>
 
     <target name="build" depends="build-all, copy-to-lib" description="builds production and debug version, copying it to omero_iviewer's lib directory" />
-    <target name="build-all" depends="clean,init, compile,compile-debug" description="builds both, production and debug version" />
+    <target name="build-all" depends="clean, init, compile, compile-debug" description="builds both, production and debug version" />
     <target name="build-prod" depends="clean, init, compile" description="builds production version" />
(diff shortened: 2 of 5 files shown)


jburel:
Run locally the build
w/o the change:

``build-all`` is invoked ``Total time: 2 minutes 45 seconds``

with the change:
``build-debug`` is invoked ``Total time:  2 minutes 30 seconds``

This reduces the time but not by a significant amount of time

waxenegger:
Well, depending on the machine and runs (npm init call which is variable), I see differences of in between 25-45 seconds usually, but I failed to also point out another gain which comes from the fact, that contrary to the test instructions I gave for this PR the dependencies would not be wiped (rm -rf node_modules).

Realistically, starting with a clean build, that is project github fetch, then "npm run prod" which fetches the main project dependencies (which takes a long time), a "build-debug" will be much shorter then, actually I have no more than 20 sec usually because the dependencies for the debug build are already there from the prod step, there is no need to redo them. And that's then a save of at least 2 minutes if I look at the runs you did.

waxenegger:
@jburel 

 I pushed a commit for what I described in the last paragraph of my previous comment.

"npm run prod" is now be optimized -in a sense- for unit tests as it builds the debug along with the production version (additional cost  it's virtually 0) which can then be used afterwards to run "ant unit-tests-only" as part of the travis workflow which limits itself to fetching test libraries, stripping whitespace and running the tests. Effectively I have eliminated a redundant build with dependency checking for main build dependencies that exist already (otherwise the main build would have failed).

I have cleaned up the build slightly, and ant-p will now only show the actual targets that make sense for developers locally, not the dependent targets. As an example, it will show the "unit-tests" target as opposed to also list "unit-tests-only" which is not what should be called for a local test run. I believe that reduces potential test fails and confusion when working locally.

Give it a try. If you think this solution causes problems in some way and/or there is something that I didn't think of that is affected by that change, let me know. I can always roll back to the previous solution.

jburel:
if the ``node_modules`` is not deleted, then yes, we have a major gain.
Will review today again

jburel:
I ran the ``npm build-debug`` followed by ``ant unit-tests-only``
Error:
``
/Users/jmarie/Documents/git-repo/omero-iviewer/build.xml:193: Execute failed: java.io.FileNotFoundException: /Users/jmarie/Documents/git-repo/omero-iviewer/test/build/ol3-viewer-test.js (No such file or directory)
``

I works if I run ``npm run prod``
so might have to clear about the command to run or not to run

waxenegger:
Yes "unit-tests-only" is not intended to be the default target for unit test runs. It's kind of a shortcut for prod/travis. The old target "unit-tests" is still the preferable one.

` ant -p `  shows only "unit-tests" and I left the README as is.

jburel:
Looks good merging

PR #36, opened 2017-02-09
use ajax setup to avoid redundancy

waxenegger:
in the spirit of https://github.com/openmicroscopy/openmicroscopy/pull/5091 this PR makes use of the global ajax setup to eliminate cache: false and dataType: json/p.

TEST: test on cowfish.openmicroscopy.org/webtrial/webclient/, any user.

Open image with iviewer, then hit F12 in your favorite browser and navigate to the network tab to view the traffic. In fact it's best to reload/refresh the page straight away so that you get the initial requests as well. The only requests that matter are the ajax requests. So, in the chrome dev tools you can reduce the list by filtering for XHR (see screenshot). Make sure all those requests have the timestamp appended.

image


waxenegger:
--exclude

This is going to clash with: https://github.com/ome/omero-iviewer/pull/28

waxenegger:
Force pushed again after merge.

will-moore:
All looks good. All ajax requests have timestamp.
Good to merge.

PR #37, opened 2017-02-14
allow modify and translate at the same time

waxenegger:
Geometric modification of shapes used to be mutually exclusive with translation. 

Now both can be done without switching modes. 

TEST: use https://cowfish.openmicroscopy.org/webtrial/, select shapes, then modify and translate them.

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -697,9 +697,11 @@ ome.ol3.Viewer.prototype.addRegions = function(options) {
 	// add a vector layer with the regions
 	if (this.regions_) {
 		this.viewer_.addLayer(new ol.layer.Vector({source : this.regions_}));
-		// enable roi selection by default
+		// enable roi selection by default,
+        // as well as modify and translate
 		this.regions_.setModes(
             [ome.ol3.REGIONS_MODE['SELECT'],
+             ome.ol3.REGIONS_MODE['MODIFY'],
              ome.ol3.REGIONS_MODE['TRANSLATE']]);
 
 		this.onViewRotationListener =
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -56,474 +56,475 @@ goog.require('ol.Map');
  * @param {Object.<string, *>=} options additional properties (optional)
  */
 ome.ol3.Viewer = function(id, options) {
-	goog.base(this);
+    goog.base(this);
 
-	var opts = options || {};
-
-	/**
-	 * the image id
-	 *
-	 * @type {number}
-	 * @private
-	 */
-	this.id_ = id || -1;
-	try {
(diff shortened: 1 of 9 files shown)


will-moore:
This is working nicely now - much improved workflow.
Good to merge.

ol3-viewer/src/ome/ol3/Viewer.js
```diff
@@ -697,9 +697,11 @@ ome.ol3.Viewer.prototype.addRegions = function(options) {
 	// add a vector layer with the regions
 	if (this.regions_) {
 		this.viewer_.addLayer(new ol.layer.Vector({source : this.regions_}));
-		// enable roi selection by default
+		// enable roi selection by default,
```
jburel: minor point: could you fix the indentation?
it is a mix of space and tab in this file

src/regions/regions-edit.js
```diff
@@ -5,8 +5,8 @@ import {Utils} from '../utils/regions';
 import {Converters} from '../utils/converters';
 import {REGIONS_MODE} from '../utils/constants';
 import {
-    REGIONS_CHANGE_MODES, REGIONS_SET_PROPERTY,
-    REGIONS_GENERATE_SHAPES,REGIONS_MODIFY_SHAPES, REGIONS_COPY_SHAPES
+    REGIONS_SET_PROPERTY, REGIONS_GENERATE_SHAPES,
```
jburel: maybe alphabetical order to make it easier to read

ol3-viewer/src/ome/ol3/Viewer.js
waxenegger: sure, as a matter of fact -and I have started this already- if I find myself making changes to a file where there are tabs I replace them from now on. this is only relevant for files that reside in ol3-viewer. it's however somewhat tedious work (like with big files) unless I just leave it at the automatic conversion...

waxenegger:
ok, had to redo all commits and force push because of https://github.com/ome/omero-iviewer/pull/23 merge into master. also added changes suggested by @jburel.

will-moore:
@waxenegger I usually find that merging of 'master' into your branch and fixing of conflicts is usually nicer than rebasing because you don't have to force push (preserves history better).

waxenegger:
well I cherry pick the commits so apart from the ones where I need to merge the changes history will be ok but might just merge master next time...

jburel:
This might a side effect of this work
If I draw 2 rectangles on the same plane
Then click "Select All" in the Edit button
I have a rectangle with white border that does not match the 2 rectangles I have initially on display

waxenegger:
Hmm, I actually have an issue with the "Select All" as well on cowfish but I don't see a white border ....
I'm going to look into it.

jburel:
you are correct 
I was in the drawing mode.
We will need to look in another PR when the ``select all`` is enabled or not

jburel:
The overlapping of shapes will also need to be investigated in another PR. 
I will add it to the card
Otherwise the PR tackles the problem "modify & translate"
Merging

PR #40, opened 2017-02-22
Shape Handling - misc changes

waxenegger:
see https://trello.com/c/yyI19Pus/265-shapes-handling

1. choose stroke/fill color before drawing
2. don't zoom close when selecting a shape in the right hand listing
3. open rois tab when clicked


waxenegger:
--exclude

PR #41, opened 2017-02-23
birdseye view: start with collapsed view unless bigger/tiled image

waxenegger:
see: https://trello.com/c/Zbpqg5MR/281-rfe

When opening the iviewer or another image within the iviewer it is not necessary to start with the birdseye view in the expanded state for smaller images (takes up space too). It can always opened on demand when zooming is done and it makes sense to use it. 

Choose small images as well as big images/pyramids and check that the birdseye view is closed in the former case and open in the latter.


jburel:
works both for small (collapsed) and big (expended) images

PR #42, opened 2017-02-23
Shape Handling - Misc

waxenegger:
This PR addresses the following trello remarks (https://trello.com/c/yyI19Pus/265-shapes-handling):

1. be able to choose stroke/fill color before entering drawing mode
2. don't zoom close when selecting a shape in the right hand listing (only center but leave zoom)
3. open rois tab when clicked (no explicit "show rois" checking is needed)
4. allow drawing 'unattached' shape (shows for all Z/T)
5. right-hand shape table: scroll to shape if clicked/selected (on the image)

Can be tested here: https://cowfish.openmicroscopy.org/webtrial/


waxenegger:
--exclude

will-moore:
- Choosing stroke/fill color before drawing is working nicely.
 - No change in zoom when selecting shape. Works fine and is a good improvement.
 - Opening ROIs tab directly works fine. I wonder if we want to move the "Show ROIs" checkbox somewhere else (either within the ROIs tab itself, or maybe simply move it to the far right of the toolbar it's currently in.
 - I can create new shapes, either attached to Z/T or unattached. But I can't seem to change the attached status of a shape *after* it's been created. The whole Drawing mode/propagation panel is kinda confusing, because it partly looks like it's trying to show the current attached/unattached state of a shape, and partly is for performing the propagate action. E.g. if I create an attached shape and then I create an Unattached shape, I would expect that when I toggle selection between these 2 shapes that the attached/unattached state would be updated in the panel, but I don't see any change.
Here's how it looks in ICY (they don't have the propagate option, but this kinda UI makes sense for the attached/unattached control).
screen shot 2017-03-15 at 10 15 00
 - Scrolling of table to selected shape is working fine.

will-moore:
For managing 'attachment' I think we could separate the attachment and propagate into different panels. Here's a photoshop mock-up:
I think it would be nice when in the 'unattached' state for the greyed-out Z/T option to track the current Z/T in the viewer. That way, you can easily switch to lock the shape on the current plane.
Or, user can enter the Z/T index if they know what index they want.
screen shot 2017-03-15 at 11 13 16

waxenegger:
I'll see how I can improve things for propagation/attachment. It might not happen in that one PR because I have another one for rois/shapes and potentially I won't get it in and merge-deployed since this one touches so many files already.

I need to point out -one more time- that the propagate button was a means to do what -according to the    design docs- was for drawing, well I mean the UI bits defined it to be for drawing. So I guess there is a slight misunderstanding in what that button that I added is supposed to achieve. I basically made it do what happens after a new shape is drawn (and propagated) only that is uses existing ones as template. In that it's never been intended to change/modify attachments. Of course it's obvious that such functionality should be there is the future.

Maybe it's best I started with that in a fresh PR. It's better because it enables me to include https://github.com/ome/omero-iviewer/pull/47 in the merge-build after this one has been merged. The propagation/change bits are mostly in different files anyhow.

jburel:
The initial propagation mock-up was there to facilitate the migration of users from insight to web
in its current form I think it takes too much space 
we will probably need to review it. 
I will create a card with the various comments.
Merging so we can move forward

PR #43, opened 2017-02-27
showing only 10 dataset thumbnails by default

waxenegger:
see: https://trello.com/c/6bKLBqO5/280-todos

It's best to limit the number of thumbnails (dataset images) shown in the left hand panel, just in case one happens to have lots of images. Therefore the limit has been set to 10 with the option to show more on demand (by clicking the expand symbol/downward arrows).

Choose an image from a dataset that has fewer than 10 images associated. The left hand panel should not display the double down arrorw. Then choose an image from a dataset that has lots of images. 10 should be loaded by default, then 10 more each time you click the arrow.

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -4,8 +4,8 @@ import {IMAGE_CONFIG_SELECT} from '../events/events';
 import Misc from '../utils/misc';
 import ImageConfig from '../model/image_config';
 import {
-    REQUEST_PARAMS,
-    WEBGATEWAY, WEBCLIENT, PLUGIN_NAME, URI_PREFIX, IVIEWER
+    API_PREFIX, IVIEWER, PLUGIN_NAME, REQUEST_PARAMS,
+    WEBCLIENT, WEBGATEWAY, URI_PREFIX
 } from '../utils/constants';
 
 /**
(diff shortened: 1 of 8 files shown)


waxenegger:
--exclude

will-moore:
This works fine.
The only problem is when I open an image that is not in the first 10 images in the Dataset. Then, the thumbnail for the viewed image is not present on the screen, so you can't move to next/previous and back to current image again etc.

I could make a small change to the ```paths_to_object``` support in webclient (where we already have the functionality to tell you which page an image is on) so that it supports a specified page_size/limit.
E.g.
```
/webclient/api/paths_to_object/?image=56797&limit=10
```
would then give you a path that included:
```
{
  "childIndex": 53,
  "type": "dataset",
  "id": 17752,
  "childPage": 6,
  "childCount": 61
},
```

waxenegger:
ah, I see. yes that would be very useful. It's an additional call and I'd also have to implement loading more thumbnails in the "upwards" direction (before that is I suppose... - in terms of order less than) but that's not really a problem.

will-moore:
Working nicely. Opening the correct 'page' of 10 thumbnails and allowing me to load the previous or next page.
Good to merge.

PR #44, opened 2017-02-28
using maps parameter for reverse intensity

waxenegger:
Following suit with https://github.com/openmicroscopy/openmicroscopy/pull/5082.

Check that the &maps= parameter is part of the image requests and that setting reverse is applied to the channels in question. Also check that the copy and paste, as well as the save functionality is working.

### DIFF
--- a/src/utils/constants.js
+++ b/src/utils/constants.js
@@ -40,17 +40,18 @@ export const WEBCLIENT = "WEBCLIENT";
  * @type {Object}
  */
 export const REQUEST_PARAMS = {
-    DATASET_ID : 'DATASET',
-    SERVER : 'SERVER',
-    IMAGE_ID : 'IMAGE_ID',
-    CHANNELS : 'C',
-    PLANE : 'Z',
-    TIME : 'T',
-    PROJECTION : 'P',
-    MODEL : 'M',
-    CENTER_X : 'X',
-    CENTER_Y : 'Y',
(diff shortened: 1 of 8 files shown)


waxenegger:
--exclude

will-moore:
URLs look good. Copy, paste & save working fine. Good to merge.

src/utils/constants.js
```diff
@@ -40,61 +40,62 @@ export const WEBCLIENT = "WEBCLIENT";
...
-    IMAGE_ID : 'IMAGE_ID',
-    CHANNELS : 'C',
-    PLANE : 'Z',
-    TIME : 'T',
-    PROJECTION : 'P',
-    MODEL : 'M',
-    CENTER_X : 'X',
-    CENTER_Y : 'Y',
-    ZOOM : 'ZM'
+    CHANNELS: 'C',
```
jburel: Not possible to use the constants defined in ol3-viewer/src/ome/ol3/globals.js
only diff is ``DATASET_ID``?

waxenegger: @jburel one is the internal ol3-viewer which is compiled by closure and can be run by itself. so they also have different namespacing, ecma script standards and even code. I'm not super happy about the duplication in some places but it's the nature of the beast.

jburel: Thanks for the clarification

PR #45, opened 2017-03-02
openlayer patching mechanism

waxenegger:
Potential bugs in openlayers (see: https://trello.com/c/6bKLBqO5/280-todos) may require a patching mechanism until, ideally, the problem has been fixed with the next openlayers release.

While using not stable releases could sometimes also be a solution to the issue it's safer to apply just the specific code change needed.

This PR enables such patching by simply putting the patch into PROJECT_ROOT/ol3-viewer/patches.
It will then be picked up and applied to the ol3 source tree before compilation.

TEST: 

not entirely sure what test instructions I should give in general. it's on cowfish and was building fine.
you could test malformed patches locally which will cause the build to fail, an empty patch directory should not.

as for the patch that's part of the PR, the image https://cowfish.openmicroscopy.org/webtrial/omero_iviewer/56752/ (user: root) would ordinarily have not shown entirely (cut off height/y at a length of x/width). it's now shown fully.

### DIFF
--- a/ol3-viewer/apply_patches.sh
+++ b/ol3-viewer/apply_patches.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 for p in *.patch
 do
+    [ -f "$p" ] || continue
     patch -p0 -N < "$p"
 done
--- a/ol3-viewer/apply_patches.sh
+++ b/ol3-viewer/apply_patches.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+for p in *.patch
+do
+    patch -p0 -N < "$p"
+done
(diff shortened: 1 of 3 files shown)


ol3-viewer/apply_patches.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
mtbc: Out of curiosity would `sh` suffice or are we happy to assume `bash` everywhere? (E.g., pretty safe bet on Linux I'd think.)

waxenegger: @mtbc I'd agree ;-)

will-moore:
Image at http://web-dev-merge.openmicroscopy.org/omero_iviewer/56752/ appears not to be cut-off.

PR #46, opened 2017-03-06
grayscale trello remarks

waxenegger:
see: https://trello.com/c/uGRf1IrD/9-discussion

With this PR the behavior when in grayscale should be the same as is in the web presently, i.e. only one channel is allowed to be active at a time


will-moore:
Works great. Good to merge.

PR #47, opened 2017-03-13
regions table: shows shapes within roi

waxenegger:
see: https://trello.com/c/yyI19Pus/6-shapes-handling

While there is probably going to be a redesign of the region listing/table in the future it makes sense to address the shortcoming of not showing hierarchy/relationship of rois/shapes now with this PR, especially with drawing and propagating shapes which creates such a relationships (see: https://github.com/ome/omero-iviewer/pull/48).

TEST: http://web-dev-merge.openmicroscopy.org/webclient/ (unless cowfish is back online)

Either choose an image that has a roi with more than 1 shapes in it already, or, use the iviewer to create one by  selecting one of the options in the section "drawing mode & propagation" such as "Attach to all Z and/or T", then drawing a shape. (Note on the side: don't expect the propagation button to be a 1:1 equivalent of what Insight does...)

Check ~~that the table includes the roi id as the first column and, more importantly,~~ that with multiple shapes in one roi you see a triangle icon ~~next to its id~~ which collapses/expands the shape rows (on click). 

UPDATE: the layout looks now closer to what is in the web at the moment, in that it has a first column with a toggle for collapse/expand only. I use the ID column then for the roi/shape id, on top I display the number of shapes in the "type" column. That should help with the "clickability" of the toggle icon.
Also I took out the propagation button.

### DIFF
--- a/ol3-viewer/src/ome/ol3/utils/Conversion.js
+++ b/ol3-viewer/src/ome/ol3/utils/Conversion.js
@@ -19,45 +19,44 @@ goog.provide('ome.ol3.utils.Conversion');
  * @param {number=} alpha an optional alpha channel for rgb strings
  * @return {object|null} returns the color as an object or null if parse errors
  */
- ome.ol3.utils.Conversion.convertRgbaColorFormatToObject = function(rgba, alpha) {
-		if (typeof(rgba) !== 'string' || rgba.length === 0) return null;
-		if (typeof(alpha) !== 'number')
-			alpha = 1.0;
+ome.ol3.utils.Conversion.convertRgbaColorFormatToObject = function(rgba, alpha) {
+    if (typeof(rgba) !== 'string' || rgba.length === 0) return null;
+    if (typeof(alpha) !== 'number') alpha = 1.0;
 
-		try {
-			var strippedRgba = rgba.replace(/\(rgba|\(|rgba|rgb|\)/g, "");
-			var tokens = strippedRgba.split(",");
-			if (tokens.length < 3) return null; // at a minimum we need 3 channels
+    try {
+        var strippedRgba = rgba.replace(/\(rgba|\(|rgba|rgb|\)/g, "");
+        var tokens = strippedRgba.split(",");
+        if (tokens.length < 3) return null; // at a minimum we need 3 channels
 
-			// prepare return object
-			var ret = {'red' : 255, 'green' : 255, 'blue' : 255, 'alpha' : alpha};
+        // prepare return object
+        var ret = {'red' : 255, 'green' : 255, 'blue' : 255, 'alpha' : alpha};
 
-			ret['red'] = parseInt(tokens[0], 10);
-			ret['green'] = parseInt(tokens[1], 10);
-			ret['blue'] = parseInt(tokens[2], 10);
-			if (tokens.length > 3) // optional alpha
-				ret['alpha'] = parseFloat(tokens[3], 10);
-		} catch (parse_error) {
-			return null;
-		}
+        ret['red'] = parseInt(tokens[0], 10);
+        ret['green'] = parseInt(tokens[1], 10);
+        ret['blue'] = parseInt(tokens[2], 10);
+        if (tokens.length > 3) // optional alpha
+        	ret['alpha'] = parseFloat(tokens[3], 10);
+    } catch (parse_error) {
+        return null;
+    }
 
-	// some final asserts to verify bounds
-	if (ret['red'] < 0 || ret['red'] > 255 ||
-				ret['green'] < 0 || ret['green'] > 255 ||
-				ret['blue'] < 0 || ret['blue'] > 255)
-		console.error('RGB values need to range in between 0 and 255!');
-		if (ret['alpha'] < 0 || ret['alpha'] > 1)
-			console.error('Alpha values need to range in between 0 and 1!');
+    // some final asserts to verify bounds
+    if (ret['red'] < 0 || ret['red'] > 255 ||
+        ret['green'] < 0 || ret['green'] > 255 ||
+        ret['blue'] < 0 || ret['blue'] > 255)
+            console.error('RGB values need to range in between 0 and 255!');
+    if (ret['alpha'] < 0 || ret['alpha'] > 1)
+        console.error('Alpha values need to range in between 0 and 1!');
 
-		return ret;
+    return ret;
 }
 
 /**
  * Converts a color given as a rgb(a) array, e.g. [244,112,444, 0.5]
  * into an internal color object looking like this:
- *<pre>
+ * <pre>
  * { red: 255, green: 255, blue: 255, alpha: 0.75 }
- *</pre>
+ * </pre>
  *
  * If the string is only rgb, you can either hand in an alpha value as a
  * second (optional) argument, otherwise it will default to : 1.0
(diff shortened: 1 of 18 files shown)


waxenegger:
--exclude

will-moore:
Collapsing and expanding works fine, but I wonder if this should look more like a 'tree' hierarchy as in Insight and webclient, with the arrow to the left of the ROI row that is ```>``` when collapsed and ```V``` when expanded. The arrow is also quite small (especially when expanded) which makes it hard to click.

Insight lets you select the ROI row (instead of auto-selecting shapes when you click on ROI) and I think this could be useful (e.g. for copying or deleting the ROI). I *think* that when you delete all the shapes in a ROI then you're also deleting the ROI too(?) but this is not so clear as being able to select the ROI itself.
Also (maybe not this PR) it's annoying that when I select a shape, the table scrolls that shape to be at the top (even if it's already visible and doesn't need to scroll). This makes it hard to click through a number of shapes, since the table scrolls on every click.

Re: selecting multiple shapes - this is potentially useful for updating/deleting multiple shapes, and you can do it in the image itself, but it would be good if you could also do this in the table (so you could select multiple shapes on different planes).

It might be nice to show the type of Shape that each ROI contains if all the Shapes in an ROI are the same type (which they are in the vast majority of cases). Makes it easier to browse when ROIs are collapsed. Showing the number of shapes in an ROI also helps this (as in Insight & webclient).

Could you sort the shapes in an ROI by Z and T instead of ID.

Propagation needs discussion (separate PR) about whether to allow duplicate shapes on the same plane etc. When I propagated, saved then propagated again, I get the new shapes with T saved as -1 (single timepoint image).
http://web-dev-merge.openmicroscopy.org/omero_iviewer/56754/?dataset=17803 (user-3).

waxenegger:
@will-moore 

the scrolling when selecting the list directly is a bug that I'm going to deal with in this PR. It is annoying and I noticed it after the merge-deploy took place so I figured I would do it right after the weekend.

I'm also going to look into the storage procedure and the -1 dimension value for a roi. That should not happen.

I think it's best that I commented out the propagation button for now, not because of that storage bug but it has continued to be a source of confusion. As I mentioned in the PR description and in another PR it is not a 1:1 in terms of behavior to insights propagation (with visual grid). I'll bring it back in a separate PR with the aim of making it the same.

I'll also try and improve clickability of the expand/collapse within this PR. 

As I said in the description this PR is not going to be the end of the work on the table. A lot of your suggestions while trivial to implement when looked at them individually pose somewhat of a challenges in combination because technically they have slightly conflicting needs, alse UI wise. Sorting for instance is not rocket science, to do it efficiently is a different story, so is filtering. It impacts on the data structure behind it which is a handy js Map for now for easier look up by id which is needed a lot as is, yet for sort and filtering an array is better. For hierarchy a tree is better, yet to display the information a table with columns is useful, otherwise one needs a separate popup for info. yet with a separate popup and a mere tree sorting of any other content columns within that structure is not possible. Long columns such as comments should maybe be resizable which poses a bit of a challenge when additional roi info does not follow that column layout. Also what do you do if you go more flexible and your roi folder is not just unique to a propagation container. So what should you reveal in the roi about more than 1 type of shape or a range that has in fact gaps. in insight if I leave out a dimension step the roi header still indicates the full range which is strictly speaking not correct. 

A right-click context menu for actions on rois is probably highly desirable regardless of tree/table layout.

So the bottom line is: I'm going to try and improve it gradually, anything else would lead to a PR that is going to be open for a very long time...

will-moore:
Scrolling on selection bug is definitely improved (table doesn't scroll when I click a row in the table) but the table still scrolls unnecessarily when I click on a shape on the image. This is not a blocker, but would be "nice" to fix it sometime if possible.

 Expand/collapse of ROI is much clearer, apart from one thing. Before your last changes, you showed the ROI ID and Shape ID for ROIs with a single shape. This was quite nice because it was clearer what the ID was (instead of it being Shape OR ROI ID in the column) but also because you could see the hierarchy clearer. Now, you can't tell which shapes are nested under an ROI. In this screenshot, the selected Shape is the last child of parent ROI, but the next shape is an ROI with single Shape.
The same shape is selected in webclient table on right - NB: the IDs for this shape are different because one viewer is showing ROI ID and one is showing Shape ID, but I don't know which is which!

screen shot 2017-03-27 at 12 59 56

I appreciate that space is limited for an extra ROI ID and Shape ID column, but it makes the hierarchy (and our trick of collapsing a ROI with single Shape into 1 row) a lot clearer.
Column headings should be ROI and Shape instead of "Roi" and "ID" as before. Here's a screenshot from last week for comparison:

screen shot 2017-03-24 at 10 56 26

Sorting of shapes in an ROI is working.
Also count of Shapes in an ROI is shown fine.

Summary: if you could show all ROI IDs and Shape IDs, that would be great.

pwalczysko:
Agree with comment of @will-moore - having the shapes which have no ROIs displayed mixed with shapes which do does not work - it is not possible to keep overview, you do need to have a separate column for ROIs.

Selection of shapes on different Z planes is possible when in "Draw" mode (the z section in the image changes when you click), but the tooltip suggests the "no parking
screen shot 2017-03-27 at 17 03 21
" sign, which is confusing.

waxenegger:
@pwalczysko let me clarify: no mixing of any structures/objects/containers/types is going on. Looking at the present web solution I saw it had only one id column. I thought -for space considerations- that this was not a bad idea and used it. the only difference is that I happened to display the roi id in the "roi header row" since it is not a shape and then for the shapes within I displayed their shape id. I thought it made sense from the context but can see that it might be confusing despite my bold claim that the average user won't care much for either id when they draw and edit. 

@will-moore So if I understand you correctly and both ids ought to be displayed then -by analogy- the web suffers from the same shortcomings and ought to be rectified as well...

I'm going to go back to the layout that displays both ids (at all times) with the toggle being in its own column altogether.

will-moore:
Yes, webclient could be improved. Not sure if/when we'll get around to it.

Now, you've got ROI IDs in every shape row, whereas before you didn't have ROI IDs in shapes that were nested under an ROI and the hierarchy was much clearer then.
In this screenshot it's not instantly clear which shapes are under the parent ROI (307803):
screen shot 2017-03-28 at 09 25 18

The column now named ID should be "Shape" and "Roi" should be "ROI" I think.

Table doesn't scroll at all when I select a shape that's already visible. Great!

waxenegger:
ok, will make the name changes.

the rationale behind the roi id in every shape was that I thought if one had a long list of shapes in a roi, one might not see the header row and therefore its id without scrolling. from the previous comments it sounded as if both ids are quite important to be displayed so I thought if that's so and one wanted to know, it'd be annoying to have to scroll to get to it.

ol3-viewer/src/ome/ol3/utils/Conversion.js
```diff
@@ -527,84 +482,81 @@ ome.ol3.utils.Conversion.LOOKUP = {
...
+                presentStrokeStyle);
+    }
+    var presentStrokeWidth =
+        (typeof(feature['oldStrokeStyle']) === 'object' &&
+        feature['oldStrokeStyle'] !== null &&
+        typeof(feature['oldStrokeStyle']['width']) === 'number') ?
+            feature['oldStrokeStyle']['width'] : 1;
+
+    jsonObject['StrokeWidth'] = {
+        '@type': 'TBD#LengthI',
```
mtbc: this TBD is okay for now?

will-moore: That's what omero-marshal is currently using I guess. There are a few TBD types at the moment.

mtbc: Aha, ta.

will-moore:
Columns and IDs look good to me.
Sorting of shapes by Z within an ROI is working (for real this time).
The mouse cursor when hovering over the table (when a drawing tool is selected and table is disabled) is no-longer showing a 'no entry' / "no parking" sign.

I think this is good to merge now.

Just a summary of other points from this PR, to consider for later:
 - It would be nice if you could select shapes from the table while you're in drawing-tool mode (and auto-switch to selection mode) in the same way that you can currently switch from selection mode to drawing-mode by clicking on toolbar.
 - Multiple-selection of shapes in the table (useful for shifting several shapes at once on different layers)
 - Shapes table always has a scroll-bar across the bottom because it's always too wide (even when I resize the right panel).

screen shot 2017-03-29 at 09 40 33

ol3-viewer/src/ome/ol3/utils/Conversion.js
waxenegger: yes, that's omero marshal. I have found it wants that giving it the stroke width for marshalling ;-)

jburel:
Points added to https://trello.com/c/ulcpDFTo/13-shapes-manipulations-from-tables

PR #48, opened 2017-03-16
Attachment of Shapes Revisited (incl. modification of existing shape attachment)

waxenegger:
see trello (https://trello.com/c/yyI19Pus/6-shapes-handling) and https://github.com/ome/omero-iviewer/pull/42

This PR aims at providing a means to edit Z/T attachment for existing shapes. I have oriented myself at @will-moore's suggestion/mockup and put an input and unattached option (for Z/T) into the edit section of the rois tab: 

NOTE: the propagation button/part is not there as it has been disabled and will remain so until exact behavior and potential implications are determined (rois of more then one type, do we just override, etc.). I ask all potential reviewers to not start a discussion on this in this PR and stay focused on the purpose. Thank you!

Also the attachment options for drawing were refactored to become a dropup menu and combined with the drawing palette which is where they really belong contextually.


waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#179](https://ci.openmicroscopy.org/job/iviewer-merge/179/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/179/consoleText) for more details.
Possible conflicts:
  - PR #47 waxenegger 'regions table: shows shapes within roi'
    - src/model/regions_history.js
    - src/regions/regions-edit.js

waxenegger:
--exclude

will-moore:
This is generally working pretty well: Toggling Z attachment for a shape behaves as expected. When unattached the Z: input updates with current Z-index, allowing you to snap shape back to current Z (although it could use a lighter grey text to indicate you can't edit this).
Also creation of new shape with specified attachment is working OK.
However I think the UI could do with some tweaks....

I like that the Z attachment control is with the other shape tools (color etc) at the top of the right panel. This makes sense since they are effectively editing a shape in a similar way. But I think you only need a single control for editing existing shapes AND picking an attachment for new shapes. In the same way that you have a single colour picker for new/existing shapes.

It might be clearer to always include Z: and T: attachment widgets, even for single Z/T images, since the presence of both Z and T makes it clearer what they're doing. Also, I think we don't need to use radio buttons, since this is really just a boolean.

So here's a revised screenshot. Apologies for this being a moving target, but I think this iteration is needed to get this right.
Here, the 'lock' icon indicates attached state. Padlock closed is 'attached'. A tooltip could say "Lock shape to Z-index" and this widget should be used to pick the attached state of newly created shapes.

screen shot 2017-04-03 at 11 59 11


NB: When I'm in the drawing mode (e.g. Rectangle tool is selected) I am able to pick a new "Attachment" setting for the shape I am about to draw, BUT I cannot select a new colour. I have to switch out of drawing mode to select a new colour then back to Rectangle tool each time I want to change colour.

waxenegger:
@will-moore I'm going to look into some of these things, especially the lock symbol and showing both Z/T always. To that end it would be good if I could get the lock icon from you.

As for the other suggestions, particularly combining drawing/editing attachment I would almost recommend against it. For one it makes sense to give more options for drawing (as opposed to quickly change slice or put a shape in unattached) which can then be also used to copy (the previous -now commented out- misunderstood "propagate" button) but most of all the 2 actions go absolutely different code paths, one a property modification such as styling, the other is part of the draw interaction code. Also custom selection inputs allowing ranges/lists are more complicated than the modification input and I'd not want to extend the latter to be bigger, more complicated (parsing checks, different rules such as present T/Z exclusions), yet more complicated selection makes sense for drawing and copying.

Let me just quickly explain the technical issue with the draw interaction (in ol3): it is mutually exclusive with others. That makes sense because people don't select or edit shapes on an image/map when they want to use the cursor to draw. But what I'm trying to get at is that it is in a sense a state that lingers on until it is broken (draw finished or interaction removed). So any change that I want to bring to that state requires me to communicate it to the interaction prior to drawing or programmatically breaking it to set new params and then start it again (which is what I do when the drawing options change). That code goes back and forth between ol3 viewer and rest of iviewer in an async fashion (state changes have to be acknowledged - can not react on good faith) and async observers trigger them on property changes (e.g.drawing mode change). In summation one can always hack in more parameters to make the necessary distinctions to condense the UI elements even further but the code won't become easier to maintain that way. 

I'll see what I can do to make some of the changes you suggested. if I find solutions that work well within what I have I'm going to make the changes, if the code gets too hacky I'll probably suggest we don't do it this way.

will-moore:
Color change in drawing mode is great.
The lock icons are working well.
It would be better to have Z; T: instead of T: Z: since that's the order in the table and we generally use that order elsewhere (e.g. in webclient right panel "Z-sections/Timepoints:	4 x 18").

The attachment of new shapes is generally working OK.
I would expect that if I use "4-9" this would be inclusive and include 9 instead of just 4,5,6,7,8.
It can be confusing when the Custom Z I've specified doesn't include the current plane, in which case each shape I draw vanishes instantly (but logically)!
The handling of invalid inputs is logical but if you make a mistake there's no feedback. e.g. index outside Z range or 11-10 etc.
The attachment options when I have multi-Z and T image are quite complex, and even then you don't have the option of "Attach to all T but don't attach to Z".

screen shot 2017-04-05 at 11 39 51

In summary, need to fix Z/T order for lock inputs and off-by-one for Z/T custom range.

However, (as discussed) I would prefer that the options for creating new shapes are simply "attach Z?" and "attach T?". I don't see much use case for other options, and if they really are needed they can be provided by a propagate option.

will-moore:
I noticed a few other issues that do not need to be fixed in this PR: put them at https://trello.com/c/LFZZAzep/19-shape-creation

waxenegger:
Have changed order and included upper bound for custom input.

I think waiting with the drawing options is not a bad thing, they can always be easily removed, if needed to no option (just present Z/T). Other improvements can be made as well at a later stage. For the upcoming release it's better to get the permissions in as soon as possible and have those changes tested for as long/much as possible. As is I cannot even get https://github.com/ome/omero-iviewer/pull/49 into "web-dev-merge" territory because of conflicts. There should be sufficient time to address the other things after.

waxenegger:
Since they were not major changes I addressed some of the trello remarks:
- stroke width can now be selected before and during drawing
- no reset of colors/width after drawing
- ambiguous select arrow was replaced by same icon for editing arrows/markers + tooltips for drawing icons

will-moore:
All those changes seem to be working well.
This looks good to merge, and we can remove the Attachment drawing widget later, unless we find that it's something people really want to keep.

jburel:
Feedback from users added to https://trello.com/c/uCx4W1bv/20-session-with-users
Some points need to be fixed in this PR before we merge it

waxenegger:
@jburel Can I ask to be more specific with some of the points in trello, otherwise this PR is going to be open for very long and I foresee a constant back and forth, e.g. it's been mentioned that the selection style is not clear but then I'd like to ask for what a clear selection style is, also what an unambiguous selection style is since color can potentially not used as the color itself could be chosen.

will-moore:
As I understood it from @jburel today, for this PR we just need to disable the "Attachment" widget for drawing new shapes.
Maybe removing IDs in the table too, but I would suggest we leave that until we're sure we know how the table should look. I've just removed the IDs from ROI/Shapes table in OMERO.figure (it's not only iViewer that gets mixed feedback!) and it certainly makes the hierarchy less clear.
We've displayed ROI/Shape IDs in webclient and Insight for years, so it's not going to hurt to leave them in place for now and don't block this PR.

I think all the other points on the trello card are general feedback not related to this PR.

waxenegger:
Thank you for the feedback @will-moore.

I'm in the process of removing ids and adding the selection icon that was there to make it a true selection icon (not drawing arrow) so as to respond to the user feedback and be more in line with other tools that require to click the select icon to toggle draw/select. These are quick and easy changes.

Then I'm going to remove the attach options and since they are gone I'm going to see if I can combine them with the edit section widget in this first step so that unattached is possible at least when using the lock symbol.

I'll hold off on the selection style until a decision is reached and other things can be addressed afterwards.

jburel:
@waxenegger: sorry for the misunderstanding, I was not suggesting that all points need to be addressed in this PR.
only the "drawing propagation" point needed to be addressed in this PR. Thanks for doing it.
I think removing it for now from the display is fine since we need for feedback on that feature

jburel:
The drawing propagation is still present. As indicated earlier that's the sole point really need in this PR
The rest should be tackled in follow up PR.

Working:

* ID no longer present in table. ROI and Shape ID display in tooltip.
* Select "arrow" works as expected

Last point to address in this PR:

* Remove ``Attachment`` from Drawing section

Also It is currently not possible to draw on big images. This should not be addressed in this PR.  I will create a card.

waxenegger:
@jburel I didn't list it for review because the attachment removal was not "working". it was working locally in the webdev-server setup but I noticed that the build ignores the commented out template bit (I'm pretty sure that's causing the discrapencies).

I couldn't fix it any more before standup (and would have needed to merge-deploy as well).

jburel:
@waxenegger no problem thanks for the update

jburel:
@waxenegger Thanks it works. I have re-deployed it.
Merging this PR, we will come back to the propagation (all planes or not) in another PR

PR #49, opened 2017-03-29
Regions of Intererest - Permissions

waxenegger:
This PR aims at enforcing permissions through the UI incl. a refactoring to receive the associated permissions via omero-marshal and a changed json format (compared to what was used before).

TEST: a first test can be done on web-dev-merge.openmicroscopy.org.

NOTE: There is no visual identification at the moment for rois/shapes that cannot be modified/deleted, and that's intentional for now!!! Given the amount of changes it's probably best to put off special "visual effects" and make them part of a new PR.

To roughly summarize the existing behavior as it has been implemented for now: 
For images that have canAnnotate=true one can select all shapes but moving shapes around that have a canEdit=false on them will not work. They'll remain in place and in the same vein modifying their geometry won't work either and not give the circle for the vertices along the outline. Editing their style should not affect any changes and deletion will fail silently if they have a canDelete=false on them.That said one can draw new shapes if not the owner, shapes drawn/owned should of course be editable/deletable by their owner and these should be persisted when clicking save.

If canAnnotate=false for the image then the save button will be disabled altogether and the drawing tools won't show. Editing/deleting existing shapes will not do anything like in the scenario described above. 

Note on the side: I'm aware that there can be situations (and I have tested that locally) such that one has once created shapes on an image that was not owned (annotate was allowed at that point in time) but that then the privilege of annotating was taken away. So effectively I had rois/shapes that were mine and had canEdit=true and canDelete=true with canAnnotate=false for the image as well as the rois/shapes. At the moment I treat that just as rigidly as the canAnnotate=false case for the simple reason that UpdateService will throw an error if I try to save at any rate, even if I just give it the shapes that are mine and have canEdit=true. It seems to need a canAnnotate=true on the image level regardless so to enable modification/editing/deleting to then get an error when saving seemed non-sensical to me.


waxenegger:
--exclude

will-moore:
This seems to be working pretty well and could be considered "Good to merge".
The UI correctly reflects the "canAnnotate" status of the image and editing of shapes respects the "canEdit" permissions.
As mentioned above, other UI clues to indicate which shapes are 'canEdit' or showing ROI/Shape ownership are tricky and can wait for future PRs.

More minor tweaks that might help could be included in this PR or deferred:
 - Disable colour-pickers (in the same way as stroke-width is disabled) when non-editable shape selected.
 - Same for Delete option and Z/T 'Lock' icons.

waxenegger:
Thank you for your input @will-moore.

All of your suggestions I can do provided canAnnotate=false. Then I can literally disable everything or maybe better hide it. Might make sense to hide edit section altogether and I think I would have done that already hadn't it been for that one scenario that I mentioned (owner of editable shapes but canAnnotate=false) where I wasn't sure if that will be/ought to be a possibility. 

If there is a "mix" of permissions (canAnnotate=true) I'll have to leave the options open for drawing as well as editing and deleting of rois/shapes that have the corresponding permission on them.

will-moore:
@waxenegger I was only suggesting that when canAnnotate is True and non-editable shapes are selected, you should disable colour-picker, delete and lock icons in the same way that you're already doing for stroke-width.
Can also do the same when canAnnotate is false too.
Probably best to disable controls instead of hide them since it is a bit disturbing if the UI is very different, and you can also use tooltips on disabled widgets to indicate why it's disabled. Also you want to leave 'Edit' button available and hiding other controls on the same row would look wrong.

waxenegger:
@will-moore for canAnnotate=false I'm going to disable the edit section. It won't look too different that way and gives a good indication.

the canAnnotate=true is a bit tricky. For one you would have to go by selection as you said. with multi-selection possible I have inhomogeneous selections in terms of permission as well. I'd have to loop until I hit the first one at a minimum to make the call to disable and I'd have to do that also in cases of select All and the box-drag select. So if I did a select all or a drag-box select and by accident or because of proximity I happen to select ones I don't have permissions for  I'd have to manually deselect the ones that I'm not allowed to edit before I can modify their properties. Or I'd have to manually select the lot that I want to do so.

For completeness sake I'd like to mention that when I started out with the permissions work I thought about disabling selection (on the ol-viewer) level. That would have been very neat code-wise with permissions control only in one place but there are 2 problems with that:
First if it was ever possible to have mutually exclusive canEdit and canDelete in the future I'd have a problem because on that level I do not know about the intentions of the select, it's just a select. Moreover, if there ever was an action that would need selection of shapes regardless of permission (say information querying/aggregation, e.g. stats,etc) I'd close the door to that as well. So I had to take a different route and put the checks in the respective places...

waxenegger:
@will-moore: 
For the canAnnotate=false on the image I have disabled the corresponding edit sections as well as the delete dropdown. 

I revisited the code for the canAnnotate=true case, thought about what you said in terms of selection and disabling and decided to make changes such that if all selected shapes are canEdit=false then the edit section will reflect that now.

There is also a tooltip on hover that tells if something in the edit section was disabled because of lack of permission.

Unfortunately I was not able to commit in time for today's merge-deploy.

will-moore:
Looking good.
Color-picker and other controls are all disabled if none of the selected shapes are canEdit.
If a mix of canEdit shapes are selected, controls are enabled and only the editable shapes are updated.
Good to merge from me.

jburel:
Merging. We will prepare a full permissions testing in the coming days

PR #50, opened 2017-04-03
Luts - png follow up

waxenegger:
With https://github.com/openmicroscopy/openmicroscopy/pull/5210 the luts.png that is used for slider and button backgrounds has changed as it got another gradient row added to it. That led to "shifted/overlapping" luts rows in the background.

NOTE: Despite https://github.com/openmicroscopy/openmicroscopy/pull/5210 being the reason for this PR, it can/should be handled independently that is to say that the (not) merging of https://github.com/openmicroscopy/openmicroscopy/pull/5210 does not impact on this PR. It is for all intents and purposes a better handling of the luts.png dimensions.

This PR changes the rather rigid/bad strategy of fixing the png's height and sets it instead using the image's actual height (after retrieval). That way the described "shift/overlap" won't happen any more.

TEST: Check that the colors/luts backdrops are displaying properly for channel button and slider as well as in the color/luts change dialog.


snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#199](https://ci.openmicroscopy.org/job/iviewer-merge/199/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/199/consoleText) for more details.
Possible conflicts:
  - PR #48 waxenegger 'Attachment of Shapes Revisited (incl. modification of existing shape attachment)'
    - src/settings/channel-range.js

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#200](https://ci.openmicroscopy.org/job/iviewer-merge/200/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/200/consoleText) for more details.
Possible conflicts:
  - PR #48 waxenegger 'Attachment of Shapes Revisited (incl. modification of existing shape attachment)'
    - src/settings/channel-range.js

waxenegger:
--exclude

attempted conflict resolution does not cut it -> wait until after merge...

will-moore:
Looking at your code, I see that you're assuming that the list of LUTs provided by the server is the same as what's in the ```luts_10.png```, but this is not the case. The LUTs provided by the server are dynamic.
E.g. If I make a copy of an existing LUT
```
$ cd /dist/lib/scripts
$ cp luts/red-green.lut luts/red-green-blue.lut
```
and then upload to eel.openmicroscopy
```
$ omero script upload luts/red-green-blue.lut --official
```
Now your LUTs from ```red-green-blue``` onwards are off-by-one:

screen shot 2017-04-25 at 10 34 51

In webclient, I have hard-coded a list of these https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroWeb/omeroweb/webgateway/static/webgateway/js/ome.viewport.js#L27 and I lookup the index of each LUT in this list. If it's not in the list, we just show nothing:

screen shot 2017-04-25 at 10 42 15

I also noticed a minor issue and found the same problem in webclient, which I've just fixed in https://github.com/openmicroscopy/openmicroscopy/pull/5210/commits/e6843250ea6aef4f51795e9f05e763012fbc2826....
In the iViewer screenshot above, many of the LUTs have a thin 1px stripe along the top, caused by the blurring of the adjacent LUT when the background image is zoomed.
I first tried fixing this by zooming more and then applying an additional few pixels offset, to move the blur out of the way. Then I found that applying a css style of ```image-rendering: pixelated;``` to the background image means that you don't get any blur. 

I just tried this in iViewer:
Before:
screen shot 2017-04-25 at 10 34 02

After ```image-rendering: pixelated;```
screen shot 2017-04-25 at 10 33 29

Maybe a little extra offset of ```1px``` might make this even cleaner?

With the new background png from https://github.com/openmicroscopy/openmicroscopy/pull/5210/ you could also show the gradient for the channel sliders when NO lut is picked, using the greyscale gradient at the bottom of the png with a coloured background (see screenshot in the PR). This means that you can see which channels have reverse intensity even when no lut is picked for them.

waxenegger:
@will-moore I don't think that it would be advisable to have them not match (aside from an additional of the black-white gradient at the end which doesn't hurt). If anything at all I only see disadvantages in hard-coding a long list of luts and potential problems which we have already run into.... 

After all one wants to have a visual to the luts files otherwise why embark on this visualization of the list to begin with. So I think for flexibility's sake I'd rather recommend that the two match which should not be a problem given the files, a simple convention/agreement on ordering would suffice, right?

waxenegger:
@will-moore Apart from the linear array cross-scan (O=n^2) there is one other thing that should be mentioned: if I create a same name luts in a different directory it's first come, first serve and I assign a png offset to a custom luts since all I have so far as a means of synchronization is the name. At a minimum  there should be another means of disambiguation (either luts in a 'predefined' directory, that takes precedence or some other flag to weed out the luts that are not part of the pre-generated png)

waxenegger:
I'm going to put in the pixelated setting despite it not being well supported: https://developer.mozilla.org/en/docs/Web/CSS/image-rendering#Browser_compatibility

The truth is I don't know what else I should do since for the world I cannot reproduce what your screenshots show. I tried locally and web-dev, with 3 browsers (Chrome, FF and IE) and zoom up to 300%, for no lookup row I saw that blurring effect.

Regardless, the css setting won't hurt...

will-moore:
The ```pixelated``` css has certainly improved the LUT picker for me. But I still see the same issue in the channel button and slider:

screen shot 2017-04-26 at 09 51 30
(strangely this was more noticeable in the slider when the slider range was smaller).

compare with applying the ```pixelated``` css in the browser:
screen shot 2017-04-26 at 09 52 09

Handling of unknown LUTs is working fine:

screen shot 2017-04-26 at 09 58 11

And when NO lut is picked, the sliders show the plain colour gradient to black as expected. And these are correctly reversed with "Reverse Intensity" selected.

Minor point: When I click the "Reverse Intensity" label beside the checkbox, I would expect this to check the checkbox, since that's fairly standard behavior. I think you can put the text in a ```<label>``` with attribute ```for="id_of_input"``` or I see that bootstrap example simply wraps the checkbox and text with a ```<label>```. 
```
<label>
  <input type="checkbox"> Check me out
 </label>
``` 
I don't know if aurelia has/needs a way of specifically handling the click on the label?

waxenegger:
@will-moore As I said in my remark earlier: I could not yet reproduce that blur issue. Nevertheless I have added +1 to the offset to shift it away from the edge. I did that actually for all three occurrences (1 js, 2 template). Is there some caching issue here or something, I'm running out of ideas...

will-moore:
It looks good in Chrome now. To get the 'pixelated' behaviour in other browsers see http://phrogz.net/tmp/canvas_image_zoom.html but that's not needed in this PR because I also don't see the same issue in Firefox.
The "Reverse Intensity" label is clickable and everything is working nicely so this is good to merge.

will-moore:
NB: We use the 'pixelated' behaviour in webclient viewer for rendering the image when the user un-checks the "Interpolated" checkbox. It would be good to look into whether this is possible in OL3, although it will require canvas technique rather than css, so would probably be quite tricky/impossible?

waxenegger:
@will-moore: there are ways in ol3 to set global canvas context properties. where it gets more hairy/hacky is when one has to hook into the vector drawing since all high level shapes get translated to instructions basically.

by default ol3 uses: https://developer.mozilla.org/en/docs/Web/API/CanvasRenderingContext2D/imageSmoothingEnabled

this can be turned off. it comes with similar compatibility issues as the css pixelated setting itself: https://developer.mozilla.org/en/docs/Web/CSS/image-rendering

so for instance the setting (algorithm) applied in FF is: moz-crisp-edges and I suspect that will work for IE would be ms-interpolation-mode (nearest neighbor):
pixelated_dev_tools

I can use that setup (like pixelated class in web) for the luts related images if that's the effect that you are looking for. the interpolation checkbox is better for another PR.

PR #52, opened 2017-04-24
openlayers upgrade to 4.1.0

waxenegger:
This PR brings the openlayers version up to date and makes a patch that was included in https://github.com/ome/omero-iviewer/pull/45 obsolete.

TEST: not much to do since this has been on web-dev for a week. To check that the patch that has been removed is no longer needed, follow the link in the test instructions for: https://github.com/ome/omero-iviewer/pull/45


will-moore:
OL3 viewer seems to be working fine today for tiled and non-tiled images.

PR #53, opened 2017-04-24
Histogram - behave like present web

waxenegger:
When updating range values for channel intensity the histogram was not redrawn. (as is the current behavior in web). This PR addresses that shortcoming.

TEST: When adjusting the channel intensity sliders, the histogram should be updated accordingly.


will-moore:
This seems to be working pretty well except that I don't see the histogram loaded for a channel when I simply turn the channel ON.
Also, it looks like the histogram json is being loaded when the image is first loaded, even if I never choose to show the histogram. Would be nice to remove this extra call if it's not needed.

waxenegger:
Yes the load is intentional. I regard the overhead (given that it is ajax) negligible as opposed to the wait if it's done on demand.

waxenegger:
@will-moore you made me think: I do see an awful lot of inefficiency here now that the slider interaction goes through plotHistogram and requests a lot more often. Really I should have done that before...

will-moore:
Caching of histograms is working nicely.
But I still don't see the histogram switch to a new channel when I toggle that channel ON (it does switch when I drag the slider or increment the rendering inputs).
Histograms will be supported for all images that are not "tiled" (big) images. So you can use the ```"tiles": true,``` attribute of the ```imgData``` json to tell whether histograms are supported without having to try and load a histogram.

waxenegger:
that's strange, I could have sworn when I tried the preview at the beginning of this PR it had the first one selected policy always. I must have missed that clicking it to turn it off and then on again affects the histogram...

waxenegger:
@will-moore the support check was for support in terms of size but, more so, server-side support in general. That, however, has become irrelevant since many more things now require a minimum version (5.3.0 so far unless we incl. regions request from api) so, yes, it is not needed any more.

will-moore:
The switching of histogram to the correct channel when I toggle channels on/off is working great.
However, I noticed now that when I'm sliding the channel sliders, the histogram doesn't update until I stop dragging. This means that the correct channel isn't shown while sliding and also the min/max bars on the histogram don't move as I'm sliding, so I can't see where I'm at until I let go. I'm pretty sure this was working perfectly earlier?

will-moore:
Oh, and the histogram is only being loaded when needed.

will-moore:
All working nicely now. Good to merge.

src/model/image_info.js
```diff
@@ -34,13 +34,11 @@ export default class ImageInfo {
     tiled = false;
 
     /**
-     * a flag that signals whether the histogram is enabled or not
-     * this will be set accordingly by the histogram but due to its
-     * more global nature we want the flag here.
+     * a flag that signals whether we a pixe_size and hence a scalebar
```
jburel: typo ``pixel_size``

src/settings/histogram.js
```diff
@@ -147,29 +145,30 @@ export default class Histogram extends EventSubscriber {
...
-        let channel = 0;
-        if (Misc.isArray(this.image_info.channels))
+        // use handed channel index or the last active channel that was used
+        let prop =
+            typeof params.prop === 'string' ? params.prop : null;
+        let channel =
+            typeof params.channel === 'number' ? params.channel :
+                Misc.isArray(params.ranges) ?
+                    params.ranges[0].index : this.last_active_channel;
+        // should we not have a last lactive channel or our channel was set
```
jburel: typo ``lactive`` => ``active``

src/model/image_info.js
waxenegger: @jburel I actually remove all scalebar code (incl. comments) apart from the ol3-viewer bits in another PR (https://github.com/ome/omero-iviewer/pull/57) since we have said that the scalebar is always displayed (if possible).

jburel:
Thanks merging

PR #54, opened 2017-04-25
scalebar - increase spectrum (finer unit differentiation)

waxenegger:
So far the scalebar showed a lesser degree of differentiation. It now covers -like the present web- from angstrom to meters.

see also: https://trello.com/c/jzBkvdaC/24-units-of-pixels-size-missing-from-json

### DIFF
--- a/ol3-viewer/src/ome/ol3/controls/ScaleBar.js
+++ b/ol3-viewer/src/ome/ol3/controls/ScaleBar.js
@@ -14,15 +14,25 @@ ome.ol3.controls.ScaleBar = function(opt_options) {
     if (typeof(opt_options) !== 'object') opt_options = {};
 
     /**
-	 * list available units for display incl. symbol and
-     * micron multiplicator
+	 * list available units for display incl. symbol, threshold for usage
+     * and micron multiplication factor
 	 *
 	 * @type {Array.<objects>}
 	 * @private
 	 */
     this.UNITS = [
-        { unit: 'micron', multiplier: 1, symbol: 'µm'},
-        { unit: 'millimeter', multiplier: 1000, symbol: 'mm'}];
+        { unit: 'angstrom',
+          threshold: 0.1, multiplier: 10000, symbol: '\u212B'},
+        { unit: 'nanometer',
+          threshold: 1, multiplier: 1000, symbol: 'nm'},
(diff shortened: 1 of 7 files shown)


ol3-viewer/src/ome/ol3/controls/ScaleBar.js
```diff
@@ -14,21 +14,31 @@ ome.ol3.controls.ScaleBar = function(opt_options) {
...
+     * list available units for display incl. symbol, threshold for usage
+     * and micron multiplication factor
+     *
+     * @type {Array.<objects>}
+     * @private
+     */
     this.UNITS = [
-        { unit: 'micron', multiplier: 1, symbol: 'µm'},
-        { unit: 'millimeter', multiplier: 1000, symbol: 'mm'}];
+        { unit: 'angstrom',
```
jburel: This should be moved to the utils package since we will need to use the units in the new info tab (I will open a PR)

waxenegger: Unfortunately this is part of the ol3-viewer and there is no shared code as such, only eventbus communication. Also I'm not too sure how it will be reused since the default unit is micron (as it seems from the code). At most I'd get the symbol out of a shared definition, right?

jburel:
The change of cursor is not obvious when dragging the scale bar: switch to hand cursor maybe in that case?
To consider (not in this PR) during the UI refactoring is:
 * remember scale bar location when selecting a new image i.e. back to bottom right each time. I can see that being annoying for users

waxenegger:
no cursor icon yet but I'm going to put in one.

As discussed, I'm going to display the scalebar by default (if possibe - pixel size).
Keeping the location can be done, the one thing that needs to be handled is if the window size is not the same and the scalebar would be off (in the not visible part). I'm not entirely sure if we should keep the location across images but at least in between on/off toggles of the scalebar.

jburel:
if you keep the scale bar always visible, make sure you do not change header.js or header.html

This will conflict with  #55

alternatively change could be made in #55 if it makes things easier

jburel:
It works as expected, cursor is now changed. Easier to follow I think
Remembering the scale bar location could be handled in another PR

jburel:
Merging

PR #55, opened 2017-04-25
Info tab

jburel:
Initial work on info tab
layout to be adjusted when the mock-up is finished
and so methods might be removed too

Not handle in this PR is the physical size for X/Y/Z dimension.
This will require either an adjustment in the webgateway call or use the new api
The second option is preferable, this will be handled in a follow up PR.
The aim of this PR is to add support for more than 2 tabs and to do some initial UI clean up

This info will probably grow closer to the right-hand pane of the web client
option to download, run script script, annotation etc.
This will not be covered in this PR


src/info/info.js
```diff
@@ -0,0 +1,124 @@
...
+     *
+     * @memberof Header
+     * @param {Object} params the event notification parameters
+     */
+    onImageConfigChange(params = {}) {
+        let conf = this.context.getImageConfig(params.config_id);
+
+        if (conf === null) return;
+        this.image_info = conf.image_info;
+        console.log(this.image_info);
```
waxenegger: @jburel: this console.log should come out I think.

src/utils/misc.js
```diff
@@ -232,7 +232,7 @@ export default class Misc {
         channels.map(
             (c) => {
                 url+= (i !== 0 ? ',' : '') + (!c.active ? '-' : '') + (++i) +
-                 "|" + c.window.start + ":" + c.window.end + "$" + c.color;
+                 "|" + c.start + ":" + c.end + "$" + c.color;
```
waxenegger: @jburel the property access was right as it was. this method is used for saving and copy both of which don't work at the moment.

src/app/header.html
```diff
@@ -1,40 +1,5 @@
...
-            <label class="${image_info === null ||
-                 image_info.projection === 'split' ? 'disabled-color' : ''}">
-                 <input type="checkbox"
-                        disabled.bind="image_info === null ||
-                            image_info.projection === 'split'"
-               checked.bind="context.show_regions"
-               change.delegate="toggleRegions($event)">&nbsp;Show ROIs</label>
-
-        </div>
+        <div class="fixed-header-text"><strong>${image_info.image_name}</strong></div>
```
waxenegger: That's a minor thing but it's recommended that some tags (that relate to formatting) be better replaced by css equivalents. I guess the idea is to separate container/layout tags from formatting and not increase the html tag nesting, not sure:
https://www.w3schools.com/tags/tag_small.asp and https://www.w3schools.com/tags/tag_strong.asp so it's probably better to put it on a class or use the style property on the tag.

waxenegger: actually with long paths in the file name it doesn't fit the header any more and if I resize the window it tries to wrap/break the line. Problem is that the header height remains fixed (which it should in order to not push down the content below) such that the wrapped text will be cut off anyhow. so I think we could as well avoid the wrapping/line break altogether. and without strong the text may grow less long.

jburel: I have not really worked on layout (since we are reviewing that section) this was mainly to get the action in place and allowing additional tabs

src/utils/misc.js
jburel: not sure we will still need it but I will revert the change

src/app/header.html
jburel: I will adjust the handling of long name e.g.
http://web-dev-merge.openmicroscopy.org/omero_iviewer/9443/?dataset=1457

waxenegger:
--exclude

sorry @jburel , the scalebar toggle is gone and without it the other PR cannot be tested

waxenegger:
@jburel the scalebar is shown by default now so I took off the excluded.

Should you want to put in the checkbox somewhere here is the template bit that's needed:
``` 
<label class="${context.show_scalebar === null ? 'disabled-color' : ''}">
    <input type="checkbox"
               disabled.bind="context.show_scalebar === null"
               checked.bind="context.show_scalebar"
               change.delegate="toggleScalebar($event)">
    &nbsp;Show Scalebar
</label>
```
There used to be some enable/disable code in header.js which was among the things you took out and it is no longer needed. The show_scalebar (on context) has either values null (scalebar not supported) or true/false for show/hide respectively.

jburel:
@waxenegger: I had the code snippet in the info.html but after the discussion 
I removed it

waxenegger:
@jburel: does it make sense for me to clean up even more then? I mean if we show it by default with no ability to show/hide I could as well take out some more code in the aurelia parts. I'd just handle it in the ol3 viewer in that case, one event handling less...

jburel:
@waxenegger I was thinking to do that in a second round i.e. was waiting for feedback first.

will-moore:
I wonder if we want to show the "Settings" tab by default, since most users will want to use that (and not really need the Info tab so much). This is what we do in Figure.
Also, I noticed that when you switch between images (by clicking thumbnails on left) we always revert back to the Info tab being selected. This is kinda annoying if you want to move between images, tweaking settings or ROIs on each.
Minor point - Need comma separator between channel names.

src/info/info.html
```diff
@@ -0,0 +1,19 @@
...
+        <template repeat.for="column of columns">
+            <div class="col-sm-6 text-weight-bold">${column.label}</div>
+            <div class="col-sm-6">${column.value}</div>
+        </template>
+    </div>
+    <div class="row">
+        <div class="col-sm-6 text-weight-bold">Channels:</div>
+        <div class="col-sm-6">
+            <template repeat.for="channel of image_info.channels">
+                <span if.bind="!$last">${channel.label},</span>
```
waxenegger: This is most definitely a way of doing it in aurelia, another more concise way without using if.bind is: 

```
<span>${channel.label + !$last ? ',' : ''}</span> 
```

I tend to use it because the if creates comment placeholders in the html and the html template itself is a bit leaner, at the expense of having more complicated conditions I agree. So I usually use the if.bind method for large sections of "more different" html, e.g. the rois table where I have more different looking rows with event hooks that exist in one type of row but not the other. in such cases I do not know how else to handle such situations.

src/app/right-hand-panel.js
```diff
@@ -36,20 +81,20 @@ export class RightHandPanel {
      * @memberof RightHandPanel
      */
     attached() {
+
         $(this.element).find("a").click((e) => {
             e.preventDefault();
-
             let hash = e.currentTarget.hash;
-
+            this.selected_tab = e.currentTarget.hash;
```
waxenegger: I think we should combine the hash variable (above) and the selected_tab one. with the selected_tab, the local hash becomes kind of obsolete.

src/app/right-hand-panel.js
```diff
@@ -13,20 +31,47 @@ import 'bootstrap';
 @customElement('right-hand-panel')
 @inject(Context, Element)
 export class RightHandPanel {
+
+    tabs = [
```
waxenegger: I think what you are doing here is not a bad idea per se, I'm just thinking that at the moment to list and keep the tabs in the js (as well as the selected/default tab) is a bit too much. This method would make sense especially if it is used in the template to construct the tabs flexibly and lots of its properties are integrated into the corresponding html and these properties come from sources such as requests, i.e. based on data. As is we do reduce the template code with this loop which is nice I admit but we set up additional bindings with the repeat that have to be processed when the only thing that changes is the active class.

The other, bigger problem with this approach is that we still need a tabcontent for each tab item, so in a sense we achieve only half the work with the repeat, the rest is still static in the template and for it to work in any case one would have to adjust the code in the template regardless.

I know it doesn't look as nice in the template but I'd maybe go for an approach where I list the tabs explicitly and solely in html (without binding to the tab member in the js and repeat) but compare to the default/selected member in each tag to determine whether to put the active class there:
```
class="tab-pane ${selected_tab === 'settings' ? 'active' : ''}"
```

Has the nice side-effect that the loop you use is not necessary for active either and if in the class you set selected_tab to 'info' it should use that as a default.

src/info/info.html
waxenegger: There is one other thing on the side: I is recommended that the repeat.for be on the element that it repeats. I noticed you use an extra <template> whereas you can/should put the repeat on the <span>.

I know I have used template in the regions table, the reason being that I have double nesting in that repeat and I cannot produce the html I need in such a case so that I have to rely on this crutch of using <template> as a construct to be able to loop, yet not produce an actual html element. Essentially <template> is a pseudo element that does not reflect in the produced html, however most of the time you want the iteration such that an element is repeatedly produced in which case the repear.for should be on the element.

jburel: @waxenegger thanks for the hint

src/app/right-hand-panel.js
jburel: yep, i was trying other things there and left it like that

src/app/right-hand-panel.js
jburel: actually I hit the problem last night trying to sort the tab-content and wanted to discuss that today with you
I was using ``<compose>`` and had some "strange" side-effects
Let's discuss later on

waxenegger: btw. I found some good resource online which can maybe help to explain a bit more what I was trying to say: https://www.danyow.net/aurelia-custom-element-vs-compose/

jburel:
Points suggested by @will-moore should now all be implemented.

src/app/right-hand-panel.js
```diff
@@ -63,22 +57,23 @@ export class RightHandPanel {
...
                  img_conf.image_info.projection === 'split')) return;
 
-            if (!this.context.show_regions && this.selected_tab === '#rois') {
+            if (!this.context.show_regions && this.context.selected_tab === '#rois') {
                 this.context.show_regions = true;
                 this.context.publish(
                     REGIONS_SET_PROPERTY, {property: "visible", value: true});
             }
             $(e.currentTarget).tab('show');
+            console.log(this.context.selected_tab);
```
waxenegger: Think you forgot about the console.log. btw: with 'npm run debug' you can also debug locally.

waxenegger:
Apart from that one code comment everything's looking good.

jburel:
@waxenegger removed the ouput and add an extra commit if no info available.
This is to handle image for example generated from figure.
I think that's it for this PR.
Will modify the data loading in another PR so we do not conflict

PR #56, opened 2017-04-28
OL3 style changes for controls (esp. zoom)

waxenegger:
see: https://github.com/openmicroscopy/design/issues/75

If there are any requests for style changes, please be as specific as possible. The css colors and styles I cannot really fully deduce from looking at the mockups only. So if the there is a preferred style and it is already in the css just point me toward it, if not just give me the color in any notation or the respective css property that should be applied. That way we avoid a guessing and back and forth.

PR #60, opened 2017-05-04
Sliders

jburel:
Change the layout of the sliders

Ref https://github.com/openmicroscopy/design/issues/75

Not cover in this PR. This won't be included in the first version:
 * controls for projections
 * display of time info. Required to adjust a call.

Also no modification of the css to avoid conflict


jburel:
To fix: 
 * play button visible even if slider is not
 * T slider layout

jburel:
@waxenegger the text on the right hand side of the T slider is always wrapped up when I move the splitpane.
I have tried various things but still not doing what I want. any tip?

waxenegger:
@jburel I'll fetched the pr locally and I'm going to have a look if I notice something that might go wrong. 

Given that you are using glyphicons you'd need to merge https://github.com/ome/omero-iviewer/pull/56 into your branch but I'm actually guessing that you did so already (and just haven't commited).

For now I can just give you this general info: If any of the panels to the left or right are adjusted (user-interaction or window resize to rather small) the size of the center frame (ol3 viewer) will have a new width since it has now more or less space available: 
layout

For the sliders that should not be of much interest to be honest other than the fact that that's the available space you end up with.

I see a few things that slipped into the slider template that are probably not needed or need adjustment. the css class 'slider-corner' is a mere placholder that's used for in between the 2 sliders, I don't think it should be needed inside the slider html/template. With glyphicons instead of background-icons the class 'dim-arrows' might probably be somewhat smaller (if not omitted) and is not a good choice to be put on anything but the arrows for the dimensions. I'll play around with that as well and see if I can help.

jburel:
I will try the span w/o div and see if it improves the layout. I did not merge #56. I made a change locally to the index-dev so I could see the icons (not committed)

waxenegger:
@jburel sorry, I don't think I could articulate well what I meant: I have fetched your local branch and tried out things that might work for you: https://github.com/waxenegger/omero-iviewer/commits/sliders_jmarie

the last commit is the only one that matters.

I have overridden the style definitions in the template (for convenience and to not have so many more files to look at for diff) and added one style. If you chose to make use (some) of my changes you'd have to adjust the settings in the app.css file. Also the dim-arrows can be reduced (which I didn't do). I believe that all that is needed for the css settings/classes on the divs/spans in the custom dimension control is the desired width/margin. The slider div with/height (dep. horizontal/vertical) will be adjusted automatically on resize to fill the space that's left.

jburel:
@waxenegger Thanks going to cherry-pick your commit
I don't want to touch the css in this PR since it will conflict with yours I reckon

jburel:
Closing this PR, @will-moore will open a new PR

PR #61, opened 2017-05-08
Sliders

will-moore:
Continue work on @jburel's PR #60.

screen shot 2017-05-08 at 16 35 33

To test:
 - Try with images with multi-Z, single-Z, multi-T, single-T.
 - Sliders should be hidden if not needed and Z/T labels should be able to handle 3 digits:

screen shot 2017-05-08 at 16 38 32


jburel:
* Track of the slider could be a bit wider I think
 * z-slider: the max value should be at the top (not bottom) and current value at the bottom (not top)

jburel:
The values are now correctly displayed. 
The z-slider and t-slider are only visible when sizeZ > 1 and sizeT >1
I still think the slider track is too small, everything is "big" expect the track
Few more pixels will look better

will-moore:
Also need to test that the extra label shows the Z/T index while sliding the slider & tracks alongside.

@waxenegger I tried to clean up the now unused ```onViewerResize()``` (see last commits) but this seems to break the display of Dataset thumbnails and the Z slider (not sure why)? so I reverted it.

waxenegger:
@will-moore I'll have look, it sounds like some other side-effect. Might be worth trying it again making sure that the build was clean. this is an event listener so just passive receiving, no publishing (in which case there could be a potential connection).

waxenegger:
@will-moore nevermind the rebuild ... the method is still called in the class: https://github.com/will-moore/omero-iviewer/blob/2b6ebd55dd344615b78c6fa331bc0f0726923da0/src/controls/dimension-slider.js#L225

no need for that either, now that the resize is not needed.

jburel:
Image ID 9445 (user-1) has 200 timepoints

jburel:
small point: the track and arrows do not seem to be centred

src/controls/dimension-slider.js
```diff
@@ -311,6 +308,15 @@ export default class DimensionSlider extends EventSubscriber {
     }
 
     /**
+     * Projects along z or t
+     *
+     * @memberof DimensionSlider
+     */
+    projection(value) {
+        console.log(value);
+    }
```
waxenegger: one left-over from @jburel's original PR. I think we should take out the entire method for now.

will-moore:
Example images (user-3):
 - Z (not T): http://web-dev-merge.openmicroscopy.org/omero_iviewer/56754/
 - Z and T: http://web-dev-merge.openmicroscopy.org/omero_iviewer/71353/?dataset=17803
 - T (not Z): http://web-dev-merge.openmicroscopy.org/omero_iviewer/9472/?dataset=1460
 - no Z, no T: http://web-dev-merge.openmicroscopy.org/omero_iviewer/22684

src/css/app.css
```diff
@@ -194,14 +194,17 @@ a.ui-button:active,
...
 
 .center-row2 {
-    max-height: 25px;
-    margin-top: -25px
+    height: 35px;
+    margin-top: -35px;
+    position: relative;
 }
 
 .slider-corner {
```
waxenegger: I think there is still the some .slider-corner css left

waxenegger:
Apart from the few code left-overs I could spot, and a possible inclusion of conditional rendering of the dimension slider element (via if.bind) everything's fine.

src/controls/dimension-slider.js
```diff
@@ -59,9 +76,7 @@ export default class DimensionSlider extends EventSubscriber {
      * @type {Array.<string,function>}
      */
     sub_list = [[IMAGE_CONFIG_UPDATE,
-                    (params = {}) => this.onImageConfigChange(params)],
-                [IMAGE_VIEWER_RESIZE,
```
waxenegger: I believe there is also a an import of the constant left still (IMAGE_VIEWER_RESIZE)

jburel:
Something is not right now
slider: no more track or knob

slider

src/controls/dimension-slider.html
```diff
@@ -1,5 +1,5 @@
 <template>
-    <template if.bind="dim === 'z'">
+    <template if.bind="dim === 'z' && image_config.image_info.dimensions.max_z > 1">
```
waxenegger: @will-moore: please no nested template tags, they are in 9/10 cases not needed. They are sort of crutches if you don't want the expression/condition to be on the element itself. I use it in the slider because I don't want to introduce another level of element there and in the rois table because of the row-layout and nestedness of the html.

for the purpose of conditionally rendering the slider it's good enough to put that if expression that you have in the template tag on the dimension-slider element in the ol3-viewer.html (one level up so to speak). 

If you want revert it and I'm going to prepare a commit for either cherry-picking or you can use the code.

will-moore:
@waxenegger I've reverted that commit - Don't really understand why it caused the jQuery slider plugin not to initialise (as seen by @jburel) since the elements were there when needed.
I guess I need to actually learn aurelia at some point!

waxenegger:
@will-moore I have had a brief look and we can talk about it if you want. for now reverting is the best thing  to do.

waxenegger:
I think that's good to be merged.

jburel:
Thanks for sorting out the alignment, looks good now merging

PR #62, opened 2017-05-08
Units

jburel:
This PR displays physical size of pixels using units

To test, check images with units other than micrometer e.g. log as user-3. 
Check that the value displayed matches what is displayed in the web client
 * image ID: 70104 (mm) has units mm
 * image ID: 24401 has micrometer
 * Image ID: 56662 no units

see https://trello.com/c/jzBkvdaC/24-units-of-pixels-size-missing-from-json

### DIFF
--- a/plugin/omero_iviewer/urls.py
+++ b/plugin/omero_iviewer/urls.py
@@ -29,4 +29,6 @@ urlpatterns = patterns(
     url(r'^persist_rois/?$', views.persist_rois,
         name='omero_iviewer_persist_rois'),
     url(r'^request_rois/(?P<iid>[0-9]+)?/?$', views.request_rois,
-        name='omero_iviewer_request_rois'))
+        name='omero_iviewer_request_rois'),
+    url(r'^image_data/(?P<image_id>[0-9]+)/$', views.image_data,
+        name='omero_iviewer_image_data'))
--- a/plugin/omero_iviewer/urls.py
+++ b/plugin/omero_iviewer/urls.py
@@ -20,11 +20,13 @@ from django.conf.urls import url
 
 import views
 
-urlpatterns = patterns('django.views.generic.simple',
-                       url(r'^(?P<iid>[0-9]+)?/?$', views.index,
-                           name='omero_iviewer_index'),
-                       url(r'^persist_rois/?$', views.persist_rois,
-                           name='omero_iviewer_persist_rois'),
-                       url(r'^request_rois/(?P<iid>[0-9]+)?/?$',
-                           views.request_rois,
-                           name='omero_iviewer_request_rois'))
+urlpatterns = patterns(
+
+    'django.views.generic.simple',
+
+    # index 'home page' of the iviewer app
+    url(r'^(?P<iid>[0-9]+)?/?$', views.index, name='omero_iviewer_index'),
+    url(r'^persist_rois/?$', views.persist_rois,
+        name='omero_iviewer_persist_rois'),
+    url(r'^request_rois/(?P<iid>[0-9]+)?/?$', views.request_rois,
+        name='omero_iviewer_request_rois'))
--- a/plugin/omero_iviewer/urls.py
+++ b/plugin/omero_iviewer/urls.py
@@ -1,3 +1,20 @@
+#
+# Copyright (c) 2017 University of Dundee.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+
 from django.conf.urls import patterns
 from django.conf.urls import url
 
--- a/plugin/omero_iviewer/urls.py
+++ b/plugin/omero_iviewer/urls.py
@@ -30,5 +30,5 @@ urlpatterns = patterns(
         name='omero_iviewer_persist_rois'),
     url(r'^request_rois/(?P<iid>[0-9]+)?/?$', views.request_rois,
         name='omero_iviewer_request_rois'),
-    url(r'^image_data/(?P<image_id>[0-9]+)/$', views.image_data,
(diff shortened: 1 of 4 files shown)


waxenegger:
@jburel do you think somthing like this will find it's way into the original webgateway method I used?
I'm asking because in the long run I was hoping to to restore the webpack dev setup after aurelia update, switching to regions request via api and cors. It is still the most productive dev enviroment I find, I have seen so again working on the slider -> button grp conversion.

jburel:
@waxenegger:  I think it will be replaced by the new method in the api and/or it can make its way to the webgateway (I did not want to create a dependency on 5.3.2) + figure has similar code.
We will also need the time information cf. mockup when it is stabilized we can push downstream

jburel:
I have adjusted the call to load information required to display the "time values"
The display of the values will not be done in this PR.

src/info/info.js
```diff
@@ -97,32 +97,51 @@ export class Info extends EventSubscriber {
             acquisition_date = new Date(this.image_info.image_timestamp * 1000).toISOString().slice(-25, -14);
         }
         let pixels_size = "";
+        let pixels_size_label = "Pixels Size (XYZ)";
         if (typeof this.image_info.image_pixels_size === 'object') {
+            let symbol = '\xB5m'; // microns by default
+            let unit = 'MICROMETER';
             let count = 0;
+            let set = false;
             if (typeof this.image_info.image_pixels_size.x == 'number') {
```
waxenegger: given that you checked for typeof number you can safely call .toFixed on the variable itself. if it was a string it's preferable to use string.parseFloat or string.parseInt (depending on what you have/want) instead of constructing a Number object but in that case that's not necessary.

jburel: I have not touched that in this PR but can modify the data handling

waxenegger: yes, I missed that in https://github.com/ome/omero-iviewer/pull/55, sorry for bringing in up in this one.

waxenegger:
Tested all 3 images and got the expected results. Css will need some adjustment but the js code is good apart from that one remark.

jburel:
I won't touch the css in this PR.

src/info/info.js
jburel: I am not actually checking on the same value (I need to fix that!) since that part has changed

jburel:
@waxenegger: comment is now fixed

plugin/omero_iviewer/urls.py
```diff
@@ -1,13 +1,34 @@
...
+
+    'django.views.generic.simple',
+
+    # index 'home page' of the iviewer app
+    url(r'^(?P<iid>[0-9]+)?/?$', views.index, name='omero_iviewer_index'),
+    url(r'^persist_rois/?$', views.persist_rois,
+        name='omero_iviewer_persist_rois'),
+    url(r'^request_rois/(?P<iid>[0-9]+)?/?$', views.request_rois,
+        name='omero_iviewer_request_rois'),
+    url(r'^image_data/(?P<image_id>[0-9]+)?/?$', views.image_data,
```
will-moore: This should be ``` url(r'^image_data/(?P<image_id>[0-9]+)/$' ``` since the ```image_id``` is not optional.
The same can be applied to a few other urls that have used the same pattern.

plugin/omero_iviewer/views.py
```diff
@@ -188,3 +208,87 @@ def request_rois(request, iid, conn=None, **kwargs):
         return JsonResponse(ret, safe=False)
     except Exception as someException:
         return JsonResponse({"error": repr(someException)})
+
+
+@login_required()
+def image_data(request, image_id, conn=None, **kwargs):
+    if image_id is None:
```
will-moore: Using the regex suggested above in urls.py, it shouldn't be necessary to check if None here.

plugin/omero_iviewer/views.py
```diff
@@ -188,3 +208,87 @@ def request_rois(request, iid, conn=None, **kwargs):
...
+        return JsonResponse({"error":
+                            "no image id supplied for image data request"})
+    image = conn.getObject("Image", image_id)
+
+    # if image is None:
+    #    return JsonResponse({"error": "Image not found"})
```
will-moore: I think we should handle Image not found with status of 404:
``` return JsonResponse({"error": "Image not found"}, status=404) ```

File: plugin/omero_iviewer/views.py [hunk 5]
```diff
@@ -188,3 +208,87 @@ def request_rois(request, iid, conn=None, **kwargs):
...
+
+    # Test if we have Units support (OMERO 5.1)
+    px = image.getPrimaryPixels().getPhysicalSizeX()
+    pix_size_x = str(px)  # As string e.g. "0.13262 MICROMETER"
+    units_support = " " in pix_size_x
+
+    rv = imageMarshal(image)
+
+    if units_support:
```
will-moore: I think we can remove all of the ```units_support``` logic since we know we're using OMERO 5.3.x.

jburel:
@will-moore: adjusted as suggested 
I will review the other calls in another PR

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#292](https://ci.openmicroscopy.org/job/iviewer-merge/292/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/292/consoleText) for more details.
Possible conflicts:
  - PR #4 waxenegger 'cache buster'

plugin/omero_iviewer/views.py
```diff
@@ -188,3 +208,81 @@ def request_rois(request, iid, conn=None, **kwargs):
...
+    if image is None:
+        return JsonResponse({"error": "Image not found"}, status=404)
+
+    px = image.getPrimaryPixels().getPhysicalSizeX()
+
+    rv = imageMarshal(image)
+
+    # Add extra parameters with units data
+    # Note ['pixel_size']['x'] will have size in MICROMETER
+    px = image.getPrimaryPixels().getPhysicalSizeX()
```
will-moore: This line is duplicated above.

will-moore:
This approach is working fine for display of pixel sizes in the client, but we might want to consider a different approach when pushing to webgateway imgData: With this ```image_data``` json, you can't tell whether the value in the DB is in e.g. ```mm``` or if the conversion is happening in the Python code.
We might want to make the JSON data match what's in the DB, to give the client more choice over whether to display what's in the DB or do it's own conversion for display.

jburel:
I think we have a card for changing units.
As I mentioned before, this call should be replaced by a call to web api 
I did not want to do it now since it will imply changing the way the json returned is handled and we have enough moving part at the moment

will-moore:
Fixes look good.
Good to merge.

jburel:
Merging so @waxenegger can fix the conflict

PR #14, opened 2016-12-17
apps name

jburel:
Add missing "apps" file

PR #63, opened 2017-05-09
Apps Name

waxenegger:
continuation of https://github.com/ome/omero-iviewer/pull/14

TEST: 
- check that the application now 'lives' under /iviewer (browser address bar)
- open the browser dev tools, reload and check that there are no requests failing


waxenegger:
-- exclude

until further notice

waxenegger:
@jburel when looking for a good constant name I discovered the existing APP_NAME and remembered that it was used to set the browser's window title. I checked and noticed that it got removed with: https://github.com/ome/omero-iviewer/pull/55

Was that intentional or should we bring it back in?

jburel:
I think you mentioned it at the time and forgot to write something down
Since the short name is now in the header, I am not sure we really need in the title

src/utils/constants.js
```diff
@@ -3,7 +3,7 @@ import {noView} from 'aurelia-framework';
 /**
  * @type {string}
  */
-export const APP_NAME = 'iViewer';
+export const APP_NAME = 'iviewer';
```
jburel: couldn't we use the label from IviewerAppConfig?

jburel:
no requests failures

src/utils/constants.js
waxenegger: we could, I didn't do so because the main reason for this PR was to keep the app name fixed (url). So I didn't bother passing yet another thing from the plugin.

will-moore:
Url is fine and no requests errors.

PR #64, opened 2017-05-09
Info panel tidy

will-moore:
Use a table to display the image info in Info tab:

screen shot 2017-05-09 at 13 42 17


jburel:
example of long name 
http://web-dev-merge.openmicroscopy.org/omero_iviewer/9470/?dataset=1460

Using the full name in that case is not great

jburel:
let's get a bit more feedback before changing the layout
Merging

jburel:
the more I look at the long name, the more I think it is too much in our face
I will be inclined to only show a short name and have the full name in tooltip

PR #68, opened 2017-05-12
Timestamp display T slider

waxenegger:
Show the timestamp info (if available).

@jburel or @will-moore sorry, forgot: how fast are we here with the deltaTs ms or s?

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -265,7 +265,7 @@ def image_data(request, image_id, conn=None, **kwargs):
             t_index = info.theT.getValue()
             if info.deltaT is not None:
                 value = format_value_with_units(info.deltaT)
-                timemap[t_index] = round(value[0])
+                timemap[t_index] = value[0]
                 if delta_t_unit_symbol is None:
                     delta_t_unit_symbol = value[1]
         for t in range(image.getSizeT()):
--- a/src/controls/dimension-slider.html
+++ b/src/controls/dimension-slider.html
@@ -39,7 +39,7 @@
         </span>
         <span show.bind="image_config.image_info.image_exposure.length > 0"
               class="timestamp">
-            ${image_config.image_info.dimensions[dim] <
+            ${image_config.image_info.dimensions[dim] &lt;
                 image_config.image_info.image_exposure.length ?
                     image_config.image_info.image_exposure[
                         image_config.image_info.dimensions[dim]] : '&nbsp;'}
--- a/src/controls/dimension-slider.html
+++ b/src/controls/dimension-slider.html
@@ -34,6 +34,15 @@
         <div class="dim-arrows glyphicon glyphicon-chevron-right"
              click.delegate="onArrowClick(1)">
         </div>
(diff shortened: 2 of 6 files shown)


jburel:
The default is second but I just realised that there is bit missing in the call since we did not check the DeltaTUnit. I will look into that and open a PR against your branch
cf. schema
```
<xsd:attribute name="DeltaT" use="optional" type="xsd:float">
<xsd:annotation>
<xsd:documentation>
Time since the beginning of the experiment. Units are set by DeltaTUnit.
</xsd:documentation>
</xsd:annotation>
</xsd:attribute>
<xsd:attribute name="DeltaTUnit" use="optional" default="s" type="UnitsTime">
<xsd:annotation>
<xsd:documentation>The units of the DeltaT - default:seconds[s].</xsd:documentation>
</xsd:annotation>
</xsd:attribute>

```

jburel:
PR opened https://github.com/waxenegger/omero-iviewer/pull/3

jburel:
The display implemented matches the implementation in figure but I think I prefer the display in the mock-up https://github.com/openmicroscopy/design/issues/75

jburel:
If I open http://web-dev-merge.openmicroscopy.org/iviewer/9446/?dataset=1457
images with a delta T of 0s always 
I don't think we should display the time but that might too much to check

will-moore:
Sorry I didn't get around to this earlier (various issues with import on eel etc).

I eventually imported some FRAP images e.g. (user-3)
http://web-dev-merge.openmicroscopy.org/iviewer/76324/
which have a rapid rate of acquisition and you need to view quite accurate timestamps.
It looks like the data you're loading in /iviewer/image_data/76324/ is rounded to the integer seconds (same as in figure) but we really need this to be more precise.
e.g. if I remove the rounding in figure, the json I get is
```
-0.8709999322891235,
-0.6009999513626099,
-0.2710000276565552,
0.060000061988830566,
0.33000004291534424,
0.7200000286102295,
0.9890000820159912,
1.2690000534057617,
1.600000023841858,
...
```
I guess we could round this to 2 or 3 decimal places easily enough, but the harder issue is formatting this for display.
It's tricky to try and guess the best timestamp format for different timestamps / movies. So, either we stick with the current e.g. ```00:00:00.3``` and simply fix the rounding of data in the loading. Or maybe we add an extra decimal place to show hundredths of a second. This maybe slightly too accurate for a 1hour movie, but I guess it's better to show this unnecessarily sometimes than it is to lose this accuracy for those who need it? Can we find enough space for ```00:00:00.33```?

waxenegger:
@will-moore I have to go back to the python code that I merged from @jburel. The front-end code displays full millis (if they are there - 3 digits). The json is supposed to be seconds. I do expect floats with the decimals being the millis but in that case they must have been truncated/rounded.

jburel:
rounding after 3 decimals seems reasonable

will-moore:
Looks good.
Happy for this to be merged....
Just a thought (maybe for another PR) It would be nice if the T-label that shows while you're dragging could also show timestamp info when we have it, instead of just T index.

jburel:
Timestamp displayed along the T index. Merging

PR #69, opened 2017-05-12
Thumbnails size 96

will-moore:
We now use thumbnails of size ```96``` for the thumbnail slider and for rdef settings thumbnails.

This means that after saving rendering settings, or save-to-all, new thumbnails of size 96 will be requested from the server. These are cached, allowing other users in a read-only group to access thumbnails (when they can't generate their own).

See: https://trello.com/c/I6XeC3Xz/33-bug-iviewer-does-not-refresh-thumbnails

cc @pwalczysko 

To test: save settings in iviewer and check that another user in read-only group can still get thumbnails for the image. Should work for Save-To-All when Dataset thumbnails are shown on the left, and also for SPW images (no Dataset thumbnails) using Save / Save To All (Save-To-All doesn't save to Plate).

### DIFF
--- a/src/app/thumbnail-slider.js
+++ b/src/app/thumbnail-slider.js
@@ -53,13 +53,6 @@ export default class ThumbnailSlider extends EventSubscriber {
     thumbnails = [];
 
     /**
-     * the default thumbnail length
-     * @memberof ThumbnailSlider
-     * @type {Array.<Object>}
-     */
-    thumbnail_length = 96;
-
-    /**
      * the number of thumnails we request in one go
      * @memberof ThumbnailSlider
      * @type {number}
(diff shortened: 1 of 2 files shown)


pwalczysko:
Behaves as expected. Ready to merge.

jburel:
We should probably move the value in one location only 
Probably to ``utils.constants.js`` so we only have one place to change if one day it is switched to another value

jburel:
Discussed with @will-moore, better to not specified the size so we will have the default i.e. 96
but this still raises an interesting problem.
If an app requests another size, this could lead to some unexpected behaviour

jburel:
Everything looks good. 
Thumbnails are loaded.

PR #70, opened 2017-05-14
Roi ordering

jburel:
In this PR
Order shapes in the tables by z and t

cc @pwalczysko @waxenegger 

https://trello.com/c/GqokGSWB/31-feedback

### DIFF
--- a/src/model/regions_info.js
+++ b/src/model/regions_info.js
@@ -59,7 +59,7 @@ export default class RegionsInfo extends EventSubscriber {
      * @memberof RegionsInfo
      * @type {number}
      */
-     number_of_shapes = 0;
+    number_of_shapes = 0;
 
     /**
      * @memberof RegionsInfo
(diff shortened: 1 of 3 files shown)


waxenegger:
@jburel we cannot just convert the map to an array for sorting purposes.

there is more code, the map is central to roi and shapes. it's a map based access to get by id (some hash map access underneath (browser specific), we'll lose the fast query  which is used in lots of places for select, delete, modify, etc. I doubt that as is everything is going to work....there are corresponding gets ....
with arrays we get a linear scan count performance for those operations, there has to be another solution ....

jburel:
I just realised that when checking the selection.
Closing for now

waxenegger:
To be honest, for now, the easiest way would be to get the sort into the query. the map should unlike hash  maps in java guarantee order so the cheapest option is to somehow add it to the sql. That said I'm not sure what options I have with marshal...

waxenegger:
hang on, what am I saying, we can query ourselves ....

jburel:
Use tmp map

jburel:
replaced https://github.com/ome/omero-iviewer/pull/71 since for some reasons I cannot re-open the PR

PR #71, opened 2017-05-14
Roi ordering

jburel:
In this PR
Order shapes in the tables by z and t

cc @pwalczysko @waxenegger

https://trello.com/c/GqokGSWB/31-feedback

I use this time a tmp array to sort the map
This should not affect the selection as the previous changes

Note the ordering is only happening in the ROI itself
and ROIs are ordered by ID
This implies that, for example,  ROI Id 1 with one shape on t=5 will be displayed before ROI id 2 with one time shape on t=1

### DIFF
--- a/src/model/regions_info.js
+++ b/src/model/regions_info.js
@@ -280,6 +280,16 @@ export default class RegionsInfo extends EventSubscriber {
 
                         // set shape properties and store the object
                         let roiId = roi['@id'];
+                        roi.shapes.sort(function(s1, s2) {
+                            var z1 = parseInt(s1['TheZ']);
+                            var z2 = parseInt(s2['TheZ']);
+                            var t1 = parseInt(s1['TheT']);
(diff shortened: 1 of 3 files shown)


jburel:
Discussed with @waxenegger via slack this am.
We will look at improving the sorting at the query level, probably in the next version

src/model/regions_info.js
```diff
@@ -294,6 +294,25 @@ export default class RegionsInfo extends EventSubscriber {
                             shapes.set(shapeId, newShape);
                             count++;
                         });
+                        //sort the shapes map
```
waxenegger: I think we can actually do it without the tmp array. sort should be doing it in place so that you could call the sort with comparator before the array.map line further up and that should do the trick.

pwalczysko:
The functionality works as described in the header of the PR.

waxenegger:
code is looking good as well.

PR #72, opened 2017-05-15
delete rois with dialog popup in case there are deleted

waxenegger:
UPDATE: It's been decided to revert to the old behavior of deletion with the need to save explicitly to persist them. No persistence request is sent when clicking delete on selected shapes !!!!!!

~~Instead of persisting the removal of shapes on explicit save a dialog they can now be deleted straight away.~~ When clicking save a dialog is displayed asking for confirmation of action if the action includes deleted rois. If confirmed the save continues, otherwise aborted to give the user a chance to undo their actions via the history (undo/redo).

_It was necessary to make the following changes to the history:_
- _with other shapes in an unsaved state a clearing of history cannot happen after delete (as does after save)_
- _any shape before deletion could have been modified (style,attachment or position/geometry) which would have led to a history entry for undo/redo. with the shape gone these entries had to be removed as well._

_NOTE: the changes above (done for the original purpose of immediate persistence of deletion) have no bearing on the reverted behavior since immediate delete is not used any more._

while working on the above changes I also came across other things that needed improving/fixing:
- the blue color in the rois table to indicate that these shapes have been modified did not revert after undoing all changes
- the input field behavior was not the same as with attachment input and the channel settings inputs in so far as that it worked on the event 'input' (which also generated unnecessary history entries for comment changes on every key press for example) instead of 'change' or hitting enter/return/tab. 
- the dropdown menu for editing arrow heads was too narrow

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1571,10 +1571,10 @@ ome.ol3.Viewer.prototype.getSmallestViewExtent = function() {
   }
 
 /**
- * Persists modified/added shapes
+ * Persists modified/added/deleted shapes
  * see: {@link ome.ol3.source.Regions.storeRegions}
  *
- * @param {boolean} selectedOnly if true only selected regions are considered
+ * @param {Array.<string>} selected an array of selected ids (of the form: roi_id:shape_id)
  * @param {boolean} useSeparateRoiForEachNewShape if false all new shapes are combined within one roi
  * @param {string} uri a server uri to post to for persistance
  * @param {boolean} omit_client_update an optional flag that's handed back to the client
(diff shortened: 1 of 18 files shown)


pwalczysko:
When ROIs are saved, all works as expected.
When I draw unsaved ROIs, then delete couple of them (not all), the "Save" button is disabled and I cannot Save the remaining unsaved ROIs. Would it be possible to change that ? (as long as some unsaved ROIs are present, keep the Save button enabled).

Otherwise works nicely.

will-moore:
Similar comment to @pwalczysko: Everything is working but unsaved changes are ignored after Delete (no confirm/save dialog when I try to navigate to a different image or refresh the page).

waxenegger:
that actually brings up a problem I failed to see before ... the history. as is the delete wipes the history (as does a complete save), which is the reason the button is disabled btw. it has to be rewritten, however, which will be a bit of a pain since I have to trace individual shapes through history entries and remove them, and not just the delete action entry itself. for instance I should delete the entry that has the initial creation of that shape as well, right? or if I create through copy and paste I have multiple rois in that entry do I remove only that deleted shape so that a redo will produce only all other shapes ....?

will-moore:
As I was suggesting in our chat yesterday, I'm not sure that these big changes to the workflow are needed to fix the previously confusing Delete behaviour.
For me, I don't see that Delete and other edits should be handled differently (deletes saved and other edits not saved during workflow). But the most problematic is the removal of deleted items from the Undo queue. This means that I can now Undo to a state that never existed before, not to a previous state.

I have just chatted with @pwalczysko to see if there is a nicer way to solve the issue reported at https://trello.com/c/GqokGSWB/31-feedback.
Part of the problem was that Deleted items were still shown in the table in red and that these disappeared without warning on Save.
These concerns can be addressed by removing ROIs & Shapes from the table on delete (instead of showing them in Red) and then showing a confirm dialog on Save *IF* the save will include deletion of existing ROIs/Shapes in the DB.

This way, the user can edit and delete during their normal workflow without having to be disrupted by confirm dialogs. It also means they can use Undo to revert any deletes before saving.

waxenegger:
@will-moore the altering of state in that context is a bit of a rhetorical question because realistically I cannot put back back in shapes from previous action that the user believes they deleted ....

in general, there are -roughly speaking- 2 approaches (regardless of minor things such as colors or row hiding or dialogues): do we persist, i.e. go from an in memory state -> database (in which case I can discard history) or do we allow persisting of individual actions before (such as delete). 

I have always preferred the former but if we allow for the latter and intermittent saving of some, then the state of history and its altering is unavoidable in my opinion or you have stuff happening in the case of delete that is just unexpected/unnatural. On the other hand you cannot wipe the history altogether since then I have lots of stuff that has happened to shapes that have still not been persisted themselves and I'd very well want to undo some of that ....

waxenegger:
@will-moore @jburel I checked and the good news is that while I had to make a whole lot of changes to code in order to enable the immediate delete, going without it is not much of a change. The delete, store and undo are all triggered in one place and with none of that happening everything behaves as was. 

Well, actually better, the modified color is now accurate and the one case where the modified flag returned a false value has been fixed, and some code was tidied up as well.

will-moore:
@waxenegger Thanks, the workflow is nicer like that.
The Delete is working as expected (no confirm at delete and not deleted until Save) but I still see a red shape in the table when I've deleted a shape that is saved in OMERO. This should be hidden, as it is with unsaved Shapes.
Also, as requested by @pwalczysko, *IF* Save is also deleting a ROI/Shape on the server, a confirm dialog should be shown (see  https://github.com/ome/omero-iviewer/pull/72#issuecomment-302363674) e.g. "Saving your changes will include the deleting of 3 ROIs.  OK/Cancel".

waxenegger:
@will-moore I cannot reproduce the case where you say that a deleted shape is still in the table after saving it. a delete does no longer save immediately, that's what I rolled back so if you see deleted shape as red in the table it just means it's been marked for deletion and nothing more. Regardless of any additional confirmation dialog it's an immediate visual marker to somebody who wanted to undo it, they have a clear idea of what they have deleted and when it's been undone (after the confirm message) as opposed to undo, hoping, clicking save, being told, no there is 2 more, undoing, etc...

will-moore:
@waxenegger Sorry, I wasn't very clear.
Re: ``` I still see a red shape in the table when I've deleted a shape that is saved in OMERO```
what I meant was "For a shape that is already saved in OMERO, when I delete it I see a red shape in the table".
Shape saved in OMERO:
screen shot 2017-05-19 at 10 35 19
After deleting it I see:
screen shot 2017-05-19 at 10 35 41

This was the previous behaviour that was confusing.
However, having discussed again with @jburel and @pwalczysko, I think we are OK with this for now (especially since the confirm-on-Save is working).
We all have slightly different ideas on how this should work, but don't have time to discuss and fine-tune this before the first release.
I'll test a bit more, but looking good so far....

will-moore:
Everything seems to be working nicely. Good to merge from me.

PR #73, opened 2017-05-15
Usability - tooltips, popup for warning/additional info

waxenegger:
Aims at addressing some usability remarks in: https://trello.com/c/GqokGSWB/31-feedback

-) popup/tooltip for "save to all" button on settings tab
-) tooltip for attachment lock

### DIFF
--- a/src/css/app.css
+++ b/src/css/app.css
@@ -160,10 +160,10 @@ a.ui-button:active,
 
 /** APP SPECIFIC STYLES SECTION **/
 .disabled-color {
-    filter:alpha(opacity=65);
+    filter:alpha(opacity=55);
     -webkit-box-shadow:none;
     box-shadow:none;
-    opacity:.65;
+    opacity:.55;
 }
 
 .fixed-header {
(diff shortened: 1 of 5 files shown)


will-moore:
Functionality is working fine.
Minor point: use Upper case for Dataset, Image etc in tooltip and confirm dialog. Also add a question mark to the dialog question. E.g. in webclient we have ```Are you sure you want to delete the selected Dataset?```.

For the lock tooltip, IF you know the number of selected shapes when creating the tooltip (and it's easy to use this info) then we could make the tooltip a little clearer, since users are not so familiar with this concept. E.g.
 - 0 shapes: "Lock new shapes to current Z-index"
 - 1 shape: "Lock shape to Z-index"
 - >1 shapes: "Lock shapes to Z-index"

Also, making the disabled text lighter grey for the Z and T inputs would highlight the difference when they're unlocked, since the text doesn't noticeably change colour just now.

jburel:
The tooltip of ``Save to Al`` has a question mark at the end, this should be removed

will-moore:
Sorry - I forgot to check the opacity changes to the Z/T lock inputs. It seems they didn't have any affect (I don't see the ```.disabled-color``` class is being used here, so it's defaulting to the bootstrap style
```.form-control[disabled] { opacity: 1; }```.

screen shot 2017-05-18 at 11 38 20

Anyway, this was never the main point from this PR, so no harm there - but it would be nice to see this fixed elsewhere, Thanks.

PR #75, opened 2017-05-16
increase untiled retrieval limit

waxenegger:
see: https://trello.com/c/stAlbSTI/32-bug-middle-sized-images-do-not-render-correctly

The new limit for retrieving images in an untiled fashion (unless they are tiled to begin with) is: 2K^2

### DIFF
--- a/ol3-viewer/src/ome/ol3/globals.js
+++ b/ol3-viewer/src/ome/ol3/globals.js
@@ -351,7 +351,7 @@ ome.ol3.RENDER_STATUS = {
 };
 
 /**
- * Limit for untiled image retrieval: 2K^2
+ * Limit for untiled image retrieval: 1K^2
  * @const
  * @enum {number}
  */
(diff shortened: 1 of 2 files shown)


will-moore:
Looks good. The h5 image now renders fine too.
Good to merge.

jburel:
Need to test from slow network

jburel:
Having a discussion with @sbesson,  it is something we will need to check at reader/rendering engine level to optimize the retrieval since it is format specific

jburel:
@waxenegger could you change to 1kx1k for tomorrow's am build? so I can test the diff from home

jburel:
Will test over the weekend

jburel:
Still not sure we need to change to 2k by 2k now
The refactoring of the code is better
Ideally this should be a property that we request from the server could be 1kx1k 2kx2k etc.

PR #78, opened 2017-05-18
roi - show checkbox - revisited

waxenegger:
does it fix the problem (in FF/mac)


pwalczysko:
This works as expected now.

PR #79, opened 2017-05-22
T slider with visibility: hidden blocks scalebar and birdeye view interaction

waxenegger:
see: https://trello.com/c/GqokGSWB/31-feedback

Due to the hidden t slider being in the dom layout still with absolute placement it 'obscured' the ui controls underneath (in terms of mouse events). Using css 'display' instead of 'visibility' should fix that.

TEST: Check that the birdeye collapse (toggle) button in the bottom right hand corner of the viewport can be clicked and toggles the birdeye visibility. Check that the scalbar can be grabbed and dragged anywhere in the viewport (before it was only moveable under the t-slider element)
Note: This problem actually only showed in images that had no t-slider/time points but it's recommended that you check all 3 possiblilies: no sliders (z/t), one z slider (no t), one 1-slider (no z), just to confirma that there are no layout side-effects.

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -540,11 +540,9 @@ ome.ol3.Viewer.prototype.bootstrapOpenLayers = function(postSuccessHook, initHoo
        view: view
     });
 
-    // expand bird's eye view for tiles sources
-    if (source.use_tiled_retrieval_ && this.viewerState_["birdseye"] &&
-       this.viewerState_["birdseye"]['ref']
-           instanceof ome.ol3.controls.BirdsEye)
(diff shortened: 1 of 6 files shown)


will-moore:
All working fine. Good to merge.

jburel:
This looks like a side effect
now we have "time" info on the z-slider tooltip when moving the slider's knob
e.g. user-1 http://web-dev-merge.openmicroscopy.org/iviewer/9442/?dataset=1457

waxenegger:
@jburel not a side-effect, just never discovered ;-)
same code for t/z, in that case there's a difference...

jburel:
so that's a feature :-)

waxenegger:
@jburel yes. it was a hidden feature of of the other PR so to speak ;-)

jburel:
Everything works as expected now.
Merging

PR #80, opened 2017-05-22
don't hide right hand panel on image switch

waxenegger:
instead of hiding and then showing again, the right hand panel remains visible with a loading message in the respective tab.


jburel:
Working nicely. That's better I think
Merging

PR #81, opened 2017-05-24
Viewer code refactoring

waxenegger:
Goals:
- right hand panel data binding to support flexible use of selected image/region data
- ol3 viewer instance sits within own div but more importantly does not include right hand panel so that the right hand panel is independent and not bound per ol3-viewer
- ol3 full screen control includes sliders now (see previous bullet point)
- context menu on viewport to allow for actions such as copy and paste (at location) and potentially other actions
- regions table is more 'responsive' (when panel resizes) instead of manual column resizing


jburel:
--exclude

jburel:
--exclude

jburel:
removed that the exclude so it can be included in the next build

jburel:
So far I have not noticed any regression
Now that the table "expands" when the right-hand panel is resized, we may want to review the columns layout e.g. show maybe before ``Comment``
This will require discussion, not to be done in this PR

jburel:
No problem noticed
we will need to look for possible side effects in the coming days
Merging so @waxenegger can move forward with other tasks

PR #82, opened 2017-05-24
Bump the size

jburel:
still being torn with size
We will look into improving the data loading speed in the next version e.g. load image details using new API instead of webgateway
Those changes have already been tested see https://github.com/ome/omero-iviewer/pull/75

### DIFF
--- a/ol3-viewer/src/ome/ol3/globals.js
+++ b/ol3-viewer/src/ome/ol3/globals.js
@@ -351,11 +351,11 @@ ome.ol3.RENDER_STATUS = {
 };
 
 /**
- * Limit for untiled image retrieval: 1K^2
+ * Limit for untiled image retrieval: 2K^2
  * @const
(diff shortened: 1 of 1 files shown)


jburel:
@waxenegger:
we will need to look at that for next release
```
/Users/jmarie/Documents/git-repo/omero-iviewer/ol3-viewer/src/ome/ol3/globals.js:358: WARNING - enum initializer must be an object literal or an enum
     [exec] ome.ol3.UNTILED_RETRIEVAL_LIMIT = 4000000;
     [exec] ^
     [exec] 
     [exec] 0 error(s), 1 warning(s)

```

waxenegger:
@jburel that's just a copy and paste mistake of the docs that I missed.

instead of  ``` @enum {number} ``` it should be ``` @type number```

will include it in https://github.com/ome/omero-iviewer/pull/81

PR #83, opened 2017-05-24
Update docs links in readme

hflynn:
Using 5.2 link was incorrect as this is 5.3.x supported, general omero5 links mean we won't have to keep updating them.


jburel:
discussed in chat few other adjustments:
changelog and readme for ol3-viewer

PR #88, opened 2017-05-31
copy & paste improvements

waxenegger:
- fix: copy and paste only in visible viewport
- paste via context menu drops shapes in click location
- copy and past across differently sized images uses scaling factor

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1383,97 +1383,110 @@ ome.ol3.Viewer.prototype.setRegionsModes = function(modes) {
 
 /**
  * This method generates shapes and adds them to the regions layer
- *
  * It uses {@link ome.ol3.utils.Regions.generateRegions} internally
  *
- * @param {Object} shape_info the roi shape information (in 'get_rois_json' format )
- * @param {number} number the number of shapes that should be generated
- * @param {boolean} random_placement should the shapes be generated in random places? default: false
- * @param {ol.Extent=} extent the portion of the image used for generation (bbox format), default: the entire image
- * @param {Array.<Array.<number,number>>=} theDims a list of dims to associate with
- * @param {boolean=} add_history an optional flag if we should consequently not add the generation to the history
- * @param {number=} hist_id a history id to pass through and return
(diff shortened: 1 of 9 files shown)


jburel:
Copy/paste and scaling seems to work 
The location is still random maybe I am missing a step

waxenegger:
for keyboard triggered and edit section button paste ... yes. nothing has changed there.
well, they should now all end up in the viewport which was not the case before. so if you zoomed in they should end up there.

for context menu paste, the left top corner of the context menu should define the left top corner of the bounding box of the shape. given that shapes could be quite 'strange' and to find their centroid would be more involved I use their bounding boxes.

also generally, copying form a smaller image and pasting into a large tiles image created barely visible shapes since their dimensions were so different. I'm scaling now in that case.

will-moore:
Pasting with context menu places the shape at the same location as expected.
Pasting the same shape multiple times at different zoom levels on the same image that it was copied from can result in different sized shapes pasted. It gets scaled down if I am zoomed in quite a lot when pasting:

screen shot 2017-06-05 at 22 26 55

Probably the use case of copying and pasting between images of different sizes is not so common, so it might be best to simply ensure that pasted shape is always the same size as the original. I think that the visibility of pasted shape is less of an issue than the assurance that shapes are all the same size.

waxenegger:
When copying from a larger into a smaller it will obscure the entire image and you end up moving till the edges of the geometry come into the viewport so that you can effectively resize it. 

With the opposite scenario the small shapes get sort of lost and could in turn be obscured by others.

So maybe it does not make sense to allow copy and pasting between images that are vastly different or only if confirmed. Maybe a tolerance value for deviation of more than xx % (area or longest side?) should be specified and if exceeded a dialog could come up and ask whether one wants to go ahead pasting...?

jburel:
I think scaling makes more sense. We will need more feedback
We can review/adjust in a follow-up PR if required
Merging it now since this PR is preventing others to be reviewed

jburel:
Discussed with Chris, pathviewer for example does not scale when the shape is copied but we could see case for both.
We probably need to look at tools like omnigraffle/photoshop for controls ideas
e.g. we do not scale by default but offer option to scale using a given ratio
This will require some mock-ups before changing the code

PR #90, opened 2017-06-01
drawing lines (instead of via polylines)

waxenegger:
https://trello.com/c/HAjiBZpk/11-shape%3A-support-for-line


jburel:
Looks good

PR #91, opened 2017-06-01
Refactoring, episode II (The Lookup Table Strikes Back)

waxenegger:
https://trello.com/c/dUmPoHhz/21-forgotten-luts

Puts luts back onto the page again.


jburel:
Works as expected
The luts are loaded and correctly displayed post refactoring

PR #93, opened 2017-06-02
add viewer method to turn off smoothing

waxenegger:
see https://trello.com/c/4ryhpngH/10-interpolation

Update: settings tab has checkbox to turn on/off interpolation

pixelated

interpolated


waxenegger:
--exclude

jburel:
looking at the quality of the image w/o interpolation I am not sure we will need to port the feature
I will need to go back to the initial requests on forum/user lists

waxenegger:
Bear in mind though that I zoomed in to show the effect more clearly.

jburel:
thread https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=7786&p=15601

waxenegger:
yes, so that would be what the user in that thread requested -if I interpret it correctly. while often hardly noticeable at 100% the pixelated look is more pronounced when zoomed in (hence my screenshots).

jburel:
@waxenegger: that's correct.
We can discuss where we can put it. probably on settings tab

waxenegger:
--exclude

jburel:
the default value needs be read using the property
``omero.client.viewer.interpolate_pixels``

waxenegger:
@jburel since I needed to touch the view anyhow I thought I'd add the image.getRoiCount to the image_data request. if that's something that you'd rather not have I can roll-back the last commit.

will-moore:
The ROI count commit was too late for the merge build I think.
Loading the data like that makes sense, but I was thinking the best way to display it would be to add it to the ROIs tab label, so that the user would see it there when they go to click on the tab, instead of having to check the Info tab. e.g. ```ROIs [200]```. 
However, I haven't mocked this up yet, so I hope it looks OK and I guess there's an issue with updating this when the user adds ROIs and saves?

jburel:
I will not include the ROI count this round since we still have few things to decide.

waxenegger:
ok then, I'm going to undo the commit. with the update comment I was about to change my solution slightly  anyhow in such that I would have taken the initial roi count from the request while after loading the rois I would have taken the internal map size that holds the rois (to be up-to-date always).

src/app/context.js
```diff
@@ -265,12 +265,13 @@ export default class Context {
     }
 
     /**
-     * Processes and sanitizes some of the server param
+     * Processes intial/handed in parameters,
+     * conducting chcks and setting defaults
```
jburel: typo ``checks``

PR #94, opened 2017-06-02
rendering settings - floating number images

waxenegger:
see: https://trello.com/c/2MKvXiqH/9-floating-point-images

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "bootstrap": "3.3.7",
     "bluebird": "3.4.7",
     "jquery": "2.2.4",
-    "jquery-ui" : "1.12.0",
+    "jquery-ui" : "1.12.1",
     "spectrum-colorpicker" : "1.8.0",
     "d3" : "4.2.7"
   },
--- a/src/model/image_info.js
+++ b/src/model/image_info.js
@@ -542,10 +542,11 @@ export default class ImageInfo {
     *
     * @param {number} mode the channel setting mode
     * @param {number} index the channel index
+    * @param {number} precision a floating point precision
     * @return {Object|null} returns object with the respective min,max properties or null
     * @memberof ChannelRange
     */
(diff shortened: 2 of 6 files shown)


jburel:
I have imported few floating images under user-1, project "float-images"
 * The text boxes next to the slider are usually too small to display the values
 * there are differences in the values displayed between insight and iviewer (see screenshot)
float-user1-77455

waxenegger:
@jburel I have to check that, at first I suspected that the exponential representation for rounding was the issue (which comes highly recommended) but when I downloaded and imported the file locally I got the same numbers as insight. I did see, however that web-dev-merge shows different ....
local_float

waxenegger:
ok the solutions seems straight-forward. I did a merge-iviewer to check for conflicts but a deploy was never done after that so web-dev is not up-to-date just yet.

jburel:
same user
http://web-dev-merge.openmicroscopy.org/iviewer/77509/?dataset=19203 
has a very small range i.e. -0.3 +0.3
This cannot be manipulated with slider
Compare with insight for expected behaviour

waxenegger:
@jburel not sure what the heck is going on. Downloaded original and, again, no problem locally. I'll try a redeploy again, maybe closing/reopening PR or something .....

waxenegger:
Ok something's not kosher, here, either locally (but then I fail to see how I can get the desired behavior - I'd rather be lead to believe I'd get the wrong), or on the server.

@jburel Perhaps you could try this locally just to get a better impression of what is wrong...

jburel:
It works much better now 
small issue
http://web-dev-merge.openmicroscopy.org/iviewer/77507/?dataset=19203
The interval is correct and matches insight
but when you move the left knob of the slider, the right-hand side value also changes i.e. initial ``0.03`` then moves to ``0.029``

jburel:
This looks much better. It works fine.

PR #95, opened 2017-06-05
projection z stack

waxenegger:
see: https://trello.com/c/rYsBHvMW/7-add-support-for-projection


waxenegger:
@jburel there is something wrong with the server deploy: this should have been deployed according to the iviewer merge. also the deploy takes 52 seconds. while I'm impressed with the speed-up I don't have the impression that things get updated...

jburel:
@waxenegger does not seem right for the deployment
I will have a look at it

jburel:
@waxenegger I found the problem
I have restarted the job this time the changes should be in when the build is done

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#355](https://ci.openmicroscopy.org/job/iviewer-merge/355/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/355/consoleText) for more details.
Possible conflicts:
  - PR #93 waxenegger 'add viewer method to turn off smoothing'
  - Upstream changes

jburel:
* it is clearer now when we are in projection mode
* top/bottom values are updated when the slider's knob is released
One thing that leads to dragging issue (same in figure) is when top-value==bottom-value.
for example if I move the top knob and overlap the bottom, the bottom gains focus and I cannot move the top one again to extend the range up

jburel:
Drawing of ROI and projection workflow needs to be reviewed
If I project an interval then start drawing the shapes is attached to the default Z 
this will be confusing.
This will require some discussion

waxenegger:
yes, under the hood the dimension has not changed so that the ol3 layer is still filtering based on that. if we want to hide the rois while in projection mode, that shouldn't be a big change...

jburel:
I think we need to keep the ability to draw regions of interest while projecting the image
In that case we should probably not assigned the shapes to any Z

jburel:
When in projection mode, if I draw a shape the value is not linked to any z (this is good) but there is still a value in the z-text field (see screenshot) this is misleading.
projection

ol3-viewer/src/ome/ol3/source/Image.js
```diff
@@ -273,7 +283,16 @@ ome.ol3.source.Image = function(options) {
             }
             url += "&maps=" + JSON.stringify(maps);
             url += '&m=' + this.image_model_;
-            url += '&p=' + (this.split_ ? 'split' : this.image_projection_);
+            url += '&p=' +
+                        (this.split_ ?
+                            ome.ol3.PROJECTION['SPLIT'] :
+                            this.image_projection_);
+            if (this.image_projection_ === ome.ol3.PROJECTION['INTMAX'] &&
```
jburel: how will you identify the projection along T when we introduce it?
it might be better to have INTMAX_Z for example
and nothing could stop us to have both at the same time

Support for T not to be added to this PR something to keep in mind also
we will need to support other projection algorithms

Basically this will need to scale

jburel:
- [ ] No more values displayed in the Z-field:

PR #96, opened 2017-06-07
luts janelia

waxenegger:
see: https://github.com/openmicroscopy/openmicroscopy/pull/5318

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -229,15 +229,15 @@ export default class Context {
                 let i=0;
                 response.luts.map(
                     (l) => {
-                        let idx = LUTS_NAMES.indexOf(l.name);
+                        let isInList = LUTS_NAMES.indexOf(l.name) !== -1;
                         let mapValue =
                             Object.assign({
                                 nice_name :
                                     l.name.replace(/.lut/g, "").replace(/_/g, " "),
-                                index : idx
+                                index : isInList ? i : -1
                             }, l);
                         this.luts.set(mapValue.name, mapValue);
-                        if (idx >= 0) i++;
+                        if (isInList) i++;
                     });
                 for (let [id, conf] of this.image_configs) conf.changed();
             }
--- a/src/utils/constants.js
+++ b/src/utils/constants.js
@@ -195,6 +195,8 @@ export const LUTS_NAMES = [
     'orange_hot.lut',
     'phase.lut',
     'physics.lut',
+    'pup_br.lut',
+    'pup_nr.lut',
     'rainbow_rgb.lut',
     'red-green.lut',
     'red_hot.lut',


will-moore:
@waxenegger I think it would be a good idea to include the ```luts_10.png``` in the iviewer code itself instead of loading from webgateway. If you use the webgateway png then the version of iviewer must exactly match the same version of webgateway and this is too fragile/restrictive. E.g. with iviewer 0.1.0 will be broken when using the webgateway with openmicroscopy/openmicroscopy#5318.

jburel:
i.e. it will be linked to 5.3.3+
until we can generate "color of the luts" depending on what is available on the server
it is safer to have a copy in the app

waxenegger:
@will-moore 

what you are saying has no bearing on the generally flawed out of sync state of png and luts additions. the fact that we have hard-coded lists in the apps and a dynamically retrieved luts list from the webgateway already renders the question of whether we 'hard-code' something else redundant.

The question of it needing to be in sync exists regardless of a copy of the png in the project. There'd be a discrepancy between what I have list wise and the hard-coded list as well as the png and offset issues at any rate.

waxenegger:
@jburel I don't pretend to have a solution for the luts/png generation but I'd think that a solution where both the luts and the png are gotten from the webgateway is preferrable as a solution to stay in sync.
To that end why don't we add a value to the luts request for whether it's included in the png or not (new additions) instead of having this hard-coded list in every app. This also opens the door to regeneration under the hood. So regardless of how long it takes to generate a new png once that's done the flag for the newly added lut will say something like generated: new. That way I can without hardcoding and just by server resources determine everything.

waxenegger:
this is just to illustrate the basic idea: https://github.com/waxenegger/openmicroscopy/commit/2ffccb83c10c2ea6f723c69232c9e1b75b3f3b9b

will-moore:
@waxenegger That solution would also seem to be server/version independent, although it would mean that iviewer 0.2.0 would require OMERO 5.3.3 at least. However, it would be nice to get rid of the hard-coded LUT lists and luts_10.png in figure/iviewer. I'll add something like https://github.com/waxenegger/openmicroscopy/commit/2ffccb83c10c2ea6f723c69232c9e1b75b3f3b9b to https://github.com/openmicroscopy/openmicroscopy/pull/5318

waxenegger:
if we can get rid of even more hard-codedness even better.

I can meanwhile put in the respective png  as a binary but I'd absolutely hate for it to become the norm. I'd rather avoid future code updates every time the luts change and I think that goal is achievable regardless of the png being up-to-date (regenerated).

waxenegger:
Ok, I have had another look and I'm not going to put in the binary png. The changes would be unproportionally high for something that's temporary and does not behave any different in case an older omero server version was used. It adds nothing in terms of stability. If I get a luts list with fewer entries then I have a corresponding png that's shorter, and the internal list of luts would not do have a negative impact as is.

btw: adding a b64 encoded data url to the json could also be helpful. it can readily be converted into  an image and we should be able to use it for a css background-image src as well.

joshmoore:
NB: I could imagine treating lut pngs like scripts: something the server generates and caches (likely on startup). If that would be of help, happy to outline specifics.

jburel:
At the time we introduce the luts, the ".png" was put as a tmp solution since we did not have time to look at generating something dynamically
I don't disagree with you @waxenegger, we just did not have time to go back to it unfortunately and we won't have the bandwidth to do it for 5.3.3
This is a problem we know we have to solve since we have already copy in 3 different locations (insight/web/figure)
Maybe we will find time over the summer to go back to it

waxenegger:
@joshmoore sure, the dynamic (re)generation of the pngs would be the ultimate goal (with the json request for the luts list being in sync). whether under the hood a regeneration is happening is not so important, only that the json list of the luts and the png match at the point of query. some kind of 'mutex' (for lack of a better word) or internal versioning should take care of that problem. a version of the png is associated with a set of luts for that version and before the new lut has been integrated into the png the new lut is not available, that would be the simplest mechanism I could think of.

jburel:
It is currently possible to upload lut similar to script so we will need the dynamic approach.
Need to put some time aside to work on it.
This has not been top priority due to more urgent things to tackle first

will-moore:
@waxenegger I have added 'png_index' to each LUT in the JSON. This should be more useful and robust than a boolean, since it won't break e.g. if LUTs are dynamically deleted from the server. https://github.com/openmicroscopy/openmicroscopy/pull/5318/commits/a22406bde400d8adcfda2fc1885351a0081dcf5f

waxenegger:
That sounds great. And I can use that index straight away instead of setting my own internally.

jburel:
merging this PR so we can review other PR
I will update the readme accordingly

jburel:
the card capturing the LUT improvements
https://trello.com/c/SrcTVbw3/135-lut-improvements

PR #98, opened 2017-06-09
no scaling will be done for pasting

waxenegger:
Prior to this PR scaling has been performed to fit larger shapes into smaller extents and make small shapes not tiny in bigger extents.

While the bulk of code for scaling was left intact, the default behavior is that no scaling for pasting takes place. No distinction is made between same/different image pastes either.

### DIFF
--- a/ol3-viewer/src/ome/ol3/utils/Regions.js
+++ b/ol3-viewer/src/ome/ol3/utils/Regions.js
@@ -251,11 +251,11 @@ ome.ol3.utils.Regions.generateRegions =
         var ret = []; // our return array
 
         // if we have a scale factor, apply it now
-        if (typeof scale_factor === 'number') {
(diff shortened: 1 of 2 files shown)


will-moore:
This is working fine for copying and pasting within the same image or between similar sized images.
No scaling is seen regardless of zoom level etc.

Copying and pasting a large ROI onto a small image doesn't work, but I think this is OK for now (could maybe support this with a warning/scaling option later, but this is an edge-case functionality so not needed for now).

jburel:
@will-moore: Could you record your comment on the copy/paste card? since there are still few things to adjust in next release

jburel:
card created https://trello.com/c/dTIUqGJZ/5-copy-paste-large-to-small

will-moore:
Added comment to https://trello.com/c/QvOrDbUu/12-copy-paste-of-shapes-random-location

PR #102, opened 2017-06-14
roi count

waxenegger:
a first version of showing the rois (in the tab). 

intitial count is gotten via image.getRoiCount, once the regions response is processed that value is 'updated' and after that whenever changes are persisted (clicking save button).


waxenegger:
--exclude

waxenegger:
--exclude

waxenegger:
--exclude

jburel:
Do you want to re-include this PR?

waxenegger:
I think it will still conflict with the aurelia upgrade.

jburel:
* The unit is now correctly displayed in excel, but not in text edit so to avoid confusion we could use 
``Export measurements to Excel"
 * RFE:  the columns Z and T should be exported
 * Issue: for an image with pixels size e.g. 20985 user-1, the area/length is in micron in the table. This should be pixels but it is in pixels in the exported file (correct)

waxenegger:
@jburel I think this comment slipped into the wrong PR ;-)

jburel:
oops sorry moving it the correct one now

jburel:
This works as expected and matches the value displayed in the other clients
We can always review the way the value is displayed. I am not a big fan of ``[x]``
but this can be done in another PR.

Merging so we fix start the review of other PRs

PR #103, opened 2017-06-15
use separate gradient png for luts

waxenegger:
Given future changes (gradient being removed from luts png) and to avoid a separate offset calculation the gradient row has now been added as its own image and the code been adjusted to use it instead of the last row of the luts png.


will-moore:
Working fine. Good to merge.

PR #104, opened 2017-06-18
open with - images: preserve selection list order, display first in list by default

waxenegger:
see: https://trello.com/c/OOdPOerR/24-order-of-images-selected-in-web-is-illogical-in-iviewer

### DIFF
--- a/src/openwith.js
+++ b/src/openwith.js
@@ -39,7 +39,8 @@ OME.setOpenWithUrlProvider("omero_iviewer", function(selected, url) {
 
     // append image list if more than one image was selected
     url += '?images=';
-    selected.map((img) => url += img.id + ',');
+    for (var i=0;i<selected_count;i++)
+        url += selected[i].id + ',';
 
     return url.substring(0, url.length-1);
 });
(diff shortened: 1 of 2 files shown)


pwalczysko:
Works as expected. Ready to merge.

src/openwith.js
```diff
@@ -39,8 +39,7 @@ OME.setOpenWithUrlProvider("omero_iviewer", function(selected, url) {
 
     // append image list if more than one image was selected
     url += '?images=';
-    for (var i=0;i<selected_count-1;i++)
-        url += selected[i].id + ',';
+    selected.map((img) => url += img.id + ',');
```
will-moore: Minor point: It's a little confusing to see ```map``` when you're not doing anything with the list returned. You could use ```forEach``` if you simply want to iterate.
Even nicer is to do something like this (which doesn't add the trailing comma).
```url += selected.map((img) => img.id).join(",");```

However, we should probably avoid arrow functions in ```openwith.js``` since that is not transpiled to ES5, so it will break in IE etc.

will-moore:
Looks better, thanks. Working fine for me too. Good to merge.

PR #105, opened 2017-06-19
Projection interval

jburel:
Change the way the projection interval is set
The approach matches the one in figure

To test:

Test 1
 * Open an image with more than 10 z-sections
 * move the slider in the middle of example
 * Click the Projection button
 * The interval should be set to value-5, value+5

Test 2:
 * Move the slider to a value < 5
 * The interval should be 0, value+5

Test 3:
 * Move the slider to a value > max-5
 * The interval should be value-5,  max

https://trello.com/c/DxUKov4y/7-intensity-projection-range-selection
cc @pwalczysko


pwalczysko:
Works as expected. Ready to merge.

PR #106, opened 2017-06-20
Update licenses

jburel:
Update licenses 

cc @waxenegger 

The second part of the list was generated using https://www.npmjs.com/package/license-checker
``license-checker  --out original.txt`` 
Then by cleaning up the original output using 
https://gist.github.com/jburel/c20f821eb3d735599a0ec01f0b1f90cf

### DIFF
--- a/libraries.txt
+++ b/libraries.txt
@@ -1,4 +1,5 @@
-Third party libraries
+Managed manually
+-----------------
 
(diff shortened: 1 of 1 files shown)


will-moore:
Assuming the webpack dependencies are right (I didn't check them manually!) looks good to merge.

PR #108, opened 2017-07-17
calculate length and area measures in ol3

waxenegger:
see https://trello.com/c/iBe0GYo0/12-analysis-of-shapes

this is the js/ol3/front-end approach of calculating area/length for shapes. it won't give more than that.
the info is shown in the roi table in the right panel. note that for now the display of area/length or comments is mutually exclusive and can be chosen by clicking on the 3 dot icon in the table header.

This can be tested now on web-dev-merge. 
There is very likely to be a follow up soon that tries to address the layout dilemma to achieve a flexible column selection...

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1461,6 +1461,8 @@ ome.ol3.Viewer.prototype.generateShapes = function(shape_info, options) {
                 !this.getRegions().rotate_text_)
             f.getGeometry().rotate(-rot);
             ome.ol3.utils.Style.updateStyleFunction(f, this.regions_, true);
+        // calculate measurements
+        this.getRegions().getLengthAndAreaForShape(f, true);
     }
     this.getRegions().addFeatures(generatedShapes);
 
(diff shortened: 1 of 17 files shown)


waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#410](https://ci.openmicroscopy.org/job/iviewer-merge/410/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/410/consoleText) for more details.
Possible conflicts:
  - PR #102 waxenegger 'roi count'
    - src/model/regions_info.js

waxenegger:
--exclude

will-moore:
This looks nice.
It would be good to see the units, either in the column name ```Length (um)``` or on a tooltip if there really isn't room.

Minor comment: when I originally draw and save new shapes, the length and area is not rounded (many decimal places). Only after refreshing the image do I see the number rounding to match existing shapes.

As an aside, most of the shapes that I had previously drawn on http://web-dev-merge.openmicroscopy.org/iviewer/56756/ (user-3) are not showing up on the image now. Any idea what's up here?

waxenegger:
Thanks for the review. I'll integrate the units of the image in some way and address the drawing display (decimals).

Now, the rois that you see or not see, there was no change, basically any logic/retrieval is as was. I display, by default, what the web api gives me. The only other possibility is that the construction of shapes failed and I'll check that because in that case I'd see an error in the dev console but the only changes I made that could have caused that were -I believe- not yet in that build.

waxenegger:
ok, I see what's going on here. for some reason, it's not changing the z index. I'll look into it.

jburel:
For units display we will need to synchronize things
(This can be done in a followup PR if it blocks)
e.g. 
http://web-dev-merge.openmicroscopy.org/iviewer/77456/?dataset=19203

``
info tab: angstrom
scale bar: nm
roi tab: micron
``

The scale bad and roi tab units should match

waxenegger:
The scale bar is a bit of its own in some sense. it adjusts depending on the zoom level.

as for the info tab and the roi table I fully agree, and I think I must have misread a code comment because, for some reason, I still assumed the default was micron.

I'll see if I can combine the code in the population of the image_info...

waxenegger:
@jburel I cannot solve the space requirement by simply fiddling with column length. there is always a case where the space will not suffice. in this case it's the angstrom that create the large numbers, in other cases it might be the sheer size of it. same for the comments.

A possible solution for this PR is to show a tooltip. 

Another approach for another PR is to redo the underlaying css layout again to go back to a fixed/manually resizable width without relative percent, perhaps we'd then have to give it a horizontal scrollbar and we'd not see all columns to the right straight away because there happen to be columns that have larger values. Yet another measure is to increase the size of the entire right-hand panel but that also only scales up to a certain point.

jburel:
don't redo the layout at this stage.

We might have to rethink the display of the units in the header of the table since the values could become very big especially in area if kept in the initial unit
Need to review the discussion we had at the time for insight since we were facing a similar issue

For now making the unit "square" in this PR is sufficient.

jburel:
Looking good now
thanks

PR #112, opened 2017-07-25
Capture Viewport/Canvas Data

waxenegger:
This PR adds 2 possible ways of saving/downloading the viewer's content (only viewport/what is seen/canvas content). Not sure if both are needed/wanted but, for now, here are both.

TEST: can be tested on web-dev-merge.openmicroscopy.org
After opening an image with iviewer you can right-click on the image shown to get the context menu which offers the following 2 possibilities: 
screenshot


waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#421](https://ci.openmicroscopy.org/job/iviewer-merge/421/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/421/consoleText) for more details.
Possible conflicts:
  - PR #108 waxenegger 'calculate length and area measures in ol3'

waxenegger:
--exclude

jburel:
conflicting PR has been merged if you want to rebase

waxenegger:
--exclude

package.json
```diff
@@ -27,7 +27,8 @@
     "jquery": "2.2.4",
     "jquery-ui" : "1.12.1",
     "spectrum-colorpicker" : "1.8.0",
-    "d3" : "4.2.7"
+    "d3" : "4.2.7",
+    "file-saver" : "1.3.3"
```
jburel: we will need to update licenses.txt so it includes the new dependency

jburel:
The feature works as expected. A png is generated
I am confused by the major difference by the name used 
in one case the original name (``/`` replaced by ``-``)+png is used, for the attachment the term ``screen-date+png`` is used. This is confusing.
I do not think that the "save as attachment" offers much.
I would think that the option to capture viewport as png will be enough
The attachment option indicates that we probably need to look at adding more data to the info panel i.e. display the annotation. Not something to do in this PR

waxenegger:
@jburel regarding the licenses, I think the file you mean is libraries.txt.

You have to tell me what tool you used to generate that. it seems to have been licence-checker but the output is different with the version I downloaded (more info) and I cannot find any options to make it look the same. 

In general I don't know if it is a fruitful approach to list recursively all dependencies. Without out knowing it with an upgrade of our, relatively few, main dependencies the list is going to look different.

Diffing with the present output I have already differences (outside of format). Is it necessary to descend into all the child dependencies (and their own...) of the main libraries we use?

Another thing I noticed is that, by default, it does not use the production dependencies only, if I use that flag I get what's in the package json under production which is what is bundled when built  for deployment. Does it make sense to limit ourselves to that maybe as part of a generated dependency list?

jburel:
I think I will do in another PR and add step to readme

jburel:
Works as expected
Did you try to save as jpeg, tif?
That might be an option to offer later on when we review the menu so it matches the options available in other viewer

For the license point, I can update the file before the release.
I will create a card on the board.

Merging

PR #113, opened 2017-08-01
image - save projection as new image

waxenegger:
see https://trello.com/c/8LFAARBr/8-rfe-projection-saving and related (but different) efforts: https://github.com/ome/omero-iviewer/pull/112 

TEST on http://web-dev-merge.openmicroscopy.org:

Choose an image with more than one z section and turn on projection by clicking on the z-stack icon under the slider. adjust the lower/upper bound to your liking, then right-click on the image viewport which should give you the context menu with an option to "Save as new Image"

save_as_new_image

On successful execution, you should be presented with a pop-up (and link) that points you to the webclient and the new image.

Notes on the side: 
- the projected image applies the last saved rendering settings as well, I can take that out if that's not wanted/needed and/or for things to be closer to insight behavior.
- if multiple T points exist all of them will be used and be part of the new image. this is analog to default settings in insight (which offers a dialog). also, selecting t ranges different to start:end had a bug: https://github.com/openmicroscopy/openmicroscopy/pull/5410


waxenegger:
--exclude

jburel:
One thing to consider is the "origin" of the projection
cf. https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/insight/SRC/org/openmicroscopy/shoola/agents/imviewer/view/ImViewerModel.java#L2155

This will need to be discussed but can be done in another PR

jburel:
With the current implementation:

 * probably need a location other than the context menu for save option. TODO Mockup to be prepared
 * projecting an image could take a long time. After selecting the ``Save new image``, the user has no feedback indicating that something is happening.
 * Click on View image when the projection is done, did not open the viewer.
 * See comment above for description or not.

waxenegger:
As is I don't think there is a service call in place that can provide me with progress. I could, at best, notify the user that the action has been triggered.

pwalczysko:
The link to the created image works for me. I think that the problem is that in the dialog (see below), there are two options to click on, one is "OK" - more pronounced, which does nothing. The other option does something, but it is more hidden (click onto the words "Vew projected image." which does work, but is not intuitive. I guess a box with non-clickable "View projected image" and two buttons, "OK" (which will bring you to the image) and "Cancel" would solve the problem.


screen shot 2017-08-07 at 14 59 59

pwalczysko:
Agree with @jburel : the option of saving the projection cannot be intuitively got to, if you do not know that the context menu must be used, you will not find this possiblity.

jburel:
* The text in the dialog should be "iviewer" instead of "iViewer"
* The name of the projected image should match what is done in insight to make the transition from insight to web smoother.
* Open in iviewer should probably open the image in a new tab

I think we have a workflow issue to fix, not to be done in this PR
if I project an image, decide not to open it or view it in web. I have way to access it
We will probably need a "reload" dataset button when we are viewing the dataset.

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#435](https://ci.openmicroscopy.org/job/iviewer-merge/435/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/435/consoleText) for more details.
Possible conflicts:
  - PR #108 waxenegger 'calculate length and area measures in ol3'
    - src/css/app.css

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#438](https://ci.openmicroscopy.org/job/iviewer-merge/438/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/438/consoleText) for more details.
Possible conflicts:
  - PR #112 waxenegger 'Capture Viewport/Canvas Data'
    - plugin/omero_iviewer/urls.py
    - src/viewers/viewer-context-menu.html
    - src/viewers/viewer-context-menu.js

jburel:
minor difference with insight is the display of the projection type
i.e. ``Maximum`` vs ``intmax`` Both should be improved to ``Maximum Intensity``
Adjustment to be done in another PR since this PR prevents other to be reviewed

PR #115, opened 2017-08-15
save area/length measurements as csv

waxenegger:
This PR enables the download of area/length information of selected shapes as a csv.

TEST (http://web-dev-merge.openmicroscopy.org):
You need to select the roi(s) first before you can right-click and save the measurements, otherwise the option will be disabled.

save_measurements


waxenegger:
--exclude

jburel:
Sorry I have not been able to test it today and web-dev-merge is currently not accessible

src/app/header.js
```diff
@@ -89,39 +75,15 @@ export class Header {
...
-        this.unregisterObservers(true);
-        this.image_info_ready_observer =
-            this.bindingEngine.propertyObserver(
-                this.image_info, 'ready').subscribe(
-                    (newValue, oldValue) => imageInfoReady());
      }
 
      /**
-      * Unregisters the the observers (image selection and image data ready)
+      * Unregisters the the observer (image selection)
```
jburel: double ``the``

src/model/image_info.js
```diff
@@ -298,8 +305,13 @@ export default class ImageInfo {
         this.can_annotate = response.perms.canAnnotate;
         if (typeof response.meta.imageAuthor === 'string')
             this.author = response.meta.imageAuthor;
-        if (typeof response.meta.imageName === 'string')
+        if (typeof response.meta.imageName === 'string') {
```
jburel: Generating a short name here makes sense
but we should in that case remove the code in ``header.js``
This is now duplicated

waxenegger: that has happened ;-)

jburel: sorry I was looking at a local branch when I was making the comment 
thanks :-)

jburel: I think we should use the short name in the info too and use the full name in the tooltip
This can be done in another PR

jburel:
The unit is not correctly displayed in the header of the cvs file
[emd_1015 emd_1015.map_roi_measurements.c.pdf](https://github.com/ome/omero-iviewer/files/1234012/emd_1015.emd_1015.map_roi_measurements.c.pdf)

It had to save it as pdf so I can upload it to the PR

waxenegger:
@jburel what do you open that csv with? 

I double checked for that file and it displays the units fine on my side. I use the very same member variable for the units as the column header and the info. 

I suspect that the unicode value for angstrom is not well received by the the application you used to open the download. Check if you can adjust that.

I also double checked the validity of the csv with https://csvlint.io and that did not complain either.

jburel:
I open the file in Excel which is what most scientists will do

waxenegger:
ok unicode opening is pretty much a must (excel can do that if told explicitly), the greek letter μ for instance is only part of extended ascii (> 126) and as such does not offer direct compatibility with utf (as I understand from some quick research).

what I'll try is something else I encountered, namely to set a byte of order mark which is rumored to help excel to interpret the csv as utf-8 (without having to specify that option explicitly).
sorry, you'll have to test it, I do only have open source office solutions for either my win8 vm and my linux and either opened the csv correctly by default.

waxenegger:
@jburel I'm going to deploy this latest commit and you can try it out.

should it work I'd like to refine this to not be done for everybody. not sure how other I can do that reliably other than an "intrusive" dialog but it seems weird to ignore the merits of utf-8 just because excel is funny about it. also it's a pain to have the encoding tables being dragged in (for encodings other than the required windows one) which blows up the size of the bundle. I "hope" to achieve that by including it in the code somehow but the imports and namespacing could prove to a spoilsport

anyhow first we'll have to see if it works...

jburel:
* The unit is now correctly displayed in excel, but not in text edit so to avoid confusion we could use 
``Export measurements to Excel"
 * RFE:  the columns Z and T should be exported
 * Issue: for an image with pixels size e.g. 20985 user-1, the area/length is in micron in the table. This should be pixels but it is in pixels in the exported file (correct)

jburel:
Not included in the build

waxenegger:
hmm. thought https://ci.openmicroscopy.org/job/iviewer-merge/ ran before you merged https://github.com/ome/omero-iviewer/pull/102 (hence before the conflict). I'm a bit puzzled why it wasn't included, anyhow, it is in now.

sbesson:
@jburel @waxenegger is the format of the exported CSV file the same as the one currently used in the Measurement Tool? If we start consuming and exposing this format across our clients, we probably need to invest time in properly specifying and versioning it?

jburel:
The measurement tool export the results as an Excel file that what scientists are familiar with

waxenegger:
@sbesson do you mean the columns (essentially the information per row) that's in it or the character set encoding or the file format as such?

I use csv sticking to its conventions/standards as best. the problem that we had was that excel is a bit picky with a seemingly uncomplicated text format such as a csv because it wants to have it with a "windows-1252" encoding. 

if it's about the actual information (columns) I can always put more information in it, provided I have it.

jburel:
measurement tool has the following Columns: ``Z, T, Type, id`` where id is the shape id
iviewer: ``z, t, type, roi_id, shape_id``

sbesson:
Briefly discussed with @waxenegger and @jburel, I can see certainly two contentious points with regard to the export format specification: the Excel-specific CSV encoding and the choice of the column names. I feel like I need to do some more reading on the former.

In the context of CMSO, one thing we have looked into is the [tabular data package](http://frictionlessdata.io/guides/tabular-data-package/) specification which stored the data CSV file(s) and the metadata in a companion JSON file. Answering https://github.com/ome/omero-iviewer/pull/115#issuecomment-326256085, as we have been done for the JSON API, one idea would be to do something similar to add references to our schema for the defined columns like `Z`.

jburel:
I forgot to mention that insight uses ``poi`` to export as excel

waxenegger:
@sbesson generally speaking, some meta-info has its merits and would definitely be a very useful ideal. what I find absolutely remarkable, however, is that the nr 1 issue that can scramble the file output is not addressed in any of the resources I followed. I wonder, genuinely, why nobody bothers to address it, I could not find anything a while ago either when I was looking at csv specs. It seems they bother with line breaks and delimitors (for good reason of course) but there is a void in respect to defining character sets out there. I guess that means  everything is possible but why is that not one thing that's worth adding to the meta-data I wonder. They designate fields to other things after all. Strange, it's not that there aren't stackoverflow and other  posts out there with people who also battle this kind of problem...

Ok, as for this PR @jburel: What do we want to change?

PS: we can envision an export "facility" in the future with a dialog to customize the output (giving reasonable defaults)

jburel:
@bramalingam being a possible consumer of the generated cvs file 2 columns that could be useful
 * image_id
 * image_name. At the moment this is already in the name of the file, but if we start exporting data for more than one image then having the name will be useful

Otherwise tested the export file:
 * open with the correct header in Sublime
 * when opening in textmate, question is asked to choose the correct encoding and the unit in header is correct displayed
Overall the output can be used by 3rd party other than excel. not the case with insight

jburel:
Merging so @waxenegger can move forward, we can always review the encoding in a follow up PR if required.
The new fields are correctly included

joshmoore:
Do we have a place to record items for the first `1.0` release?

jburel:
https://trello.com/b/d9ihqOFS/webviewer-ng-priorities-list has been used to collect ideas/suggestion

jburel:
new board for must have for 1.0 https://trello.com/b/n0uYFFOB/iviewer-10

PR #116, opened 2017-08-17
Intensity Point Query

waxenegger:
This PR displays the intensity next to the mouse cursor after checking the checkbox in the settings panel on the right hand side.

Update: Instead of querying the point by itself a square of data is requested (with the actual point being the center)


waxenegger:
--exclude

jburel:
I know it is work in progress but I am not convinced by the way the intensity is displayed

waxenegger:
I don't know of any better way to be honest. The idea was that the values are displayed as a list below the field that enables/disables the querying (showing coordinates) whenever the querying is enabled.

There are only that many places I can think of...

Say the intensity where to be put into the right hand panel it would be kind of out of sight and where the click happens.
If it's in the viewport, the question remains besides turning it on/off (more in the right-click context menu?) where it should be shown. If it's hanging around next to  the cursor as a tooltip it could get annoying quickly I think. if it triggers some sort of even small popup it needs to be clicked away each and every time otherwise it has to disappear by itself after a while which pegs the question when (to give the user enough time to read)?

So, yes, by all means, once I have a concrete suggestion as to where to put it an how to enable/disable it, I can do it that way, so far I have no better idea really. The tooltip method is the only other one that seems plausible at the moment.

Maybe I implement both so that they can be seen side-to-side...

jburel:
We need to check what other tool are doing. No good suggestion yet
I noticed that the value, for example, was ``120,5`` for the map file I was looking
this should have been ``120.5``

waxenegger:
I think there is a bit of a misunderstanding there. what you see in the ui displayed are the pixel coordinates. these will be the x/y that are sent with the request so I displayed them for now. the idea was that underneath it (if querying is enabled) there will be a box that shows the intensities per channel.

if you open the dev console you see the intensity logged there. I tested the actual querying by setting points in insight and checking their min/max and then using these pixel coordinates for my view method to check what values it gave me. so far I haven't found a discrepancy. as is my method will give too big an accuracy for floating types and that has to be rounded. also you will see really small numbers in e-xx (exponential notation).

jburel:
Thanks for the clarification.
I will play a bit more with it and we can chat on Monday

waxenegger:
--exclude

jburel:
Discussed today with @will-moore, few suggestions:

 * Replace the "outside" button by a label solely displaying the settings
 * Instead of "outside",  do not display any values if the mouse is not on the canvas
 * Add to the right-hand panel in Settings a check box: Pixels Intensity next to Show histogram. we also discussed re-arranging the right-hand controls a bit: Move Interpolate above the "Save" buttons.  Replace Show histogram by Histogram (Will that be clear enough?)
 * Load the pixels intensity only for the active channels (that's we were thinking of having the control on Settings tab)
* Display "Loading message" where the mouse is and replace the text when the data are loaded
* If possible load a "box" around the mouse point. We could start by a 10x10 and check the performance. Maybe do some caching later on

waxenegger:
@jburel we have to discuss that again. from the textual description I cannot see how some things should look.

At the moment I'm only kind of sure of the following and I'm going to start with that:
- no outside display for the button
- right panel option renames and shuffle
- turning intensity on/off in the settings section

I'm not sure about the following:
- the query is already including only the active channels.
- the mouse movement per se does not really constitute a performance issue since there is no expensive layout change taking place. the request is not sent unless the mouse movement stalls for 1/2 to not fire off backend requests all the time. but the value display or mouse movement is not going to be a problem as such.
- a 'box' is already displayed next to the cursor or do you mean you want the cursor icon to change?
- should the original button go or are the values redundantly displayed now (next to cursor) or should the coordinates be displayed in the button and the the values in the box next to the cursor?

jburel:
@waxenegger I have listed all the points we discussed to be sure we were all on the same page
it was not a PR review as such

jburel:
Sorry if I was not clear
By box, I meant loading not only the values for a given point but retrieve the intensities for a rectangle around the point so that when the user mouse the cursor the values are already available
We will need to test the performance of the retrieval for such rectangle

The coordinates should stay where they are at the moment. But this should just be a text field not a button to turn on/off the retrieval of the intensities

waxenegger:
So the latest changes which are also deployed on web-dev-merge are: 
- turning on/off via the new checkbox in the settings (incl. rearrangement)
- Querying message next to cursor
- fixes (under the hood) such as mouse over controls don't trigger queries.

Retrieving a "buffer"/rectangle of data may make sense if we keep it small and querying is done within that vicinity. If the user queries points apart it'd probably not be of much help

jburel:
User usually selects a point and checks "around" it
It won't work in all cases but will cover a good number of workflow
This will obviously depend on the size of the rectangle we can retrieve

ol3-viewer/src/ome/ol3/controls/IntensityDisplay.js
```diff
@@ -288,17 +315,44 @@ ome.ol3.controls.IntensityDisplay.prototype.handlePointerMove_ = function(e) {
...
+        // we have to request the intensities
+        // for the z,t,c,x and y given
         var reqParams = {
             "server" : this.image_.server_,
-            "uri" : this.prefix_ + "/get_intensity/?image=" + this.image_.id_ +
-                    "&z=" + this.image_.getPlane() +
-                    "&t=" + this.image_.getTime() + "&x=" +
-                    parseInt(x) + "&y=" + parseInt(y) + "&c=" +
-                    activeChannels.join(','),
+            "uri" : /*this.prefix_ + */ "get_intensity/?image=" + this.image_.id_ +
```
will-moore: this.prefix_ needed?

waxenegger: thanks, that wasn't supposed to go in.

will-moore:
This is working pretty well, although some issues with performance with slow internet at home ( > 1 minute to open the viewer with 1.8 MB bundle.js)
As far as I can see from the functionality, it looks like you are loading a 5px strip of pixel data across the image each time. I can see that this is more straight-forward than loading a 'square' of data, but it means that I only have to move a small distance before I need to load more. But it's not too bad. I need to play a bit more with it...

Minor issues. 
 - I wonder if you could use a '+' cross hairs cursor when the pixel intensity is enabled.
 - Also, offsetting the tooltip a few pixels e.g. 20 - 30 downwards & right would mean that more of the image is visible around the cursor.
 - The tooltip flickers a lot when moving around (even on pixels where we have the data in hand), hiding and showing on each pixel mouse-out and mouse-in, which is hard on the eyes when trying to follow changes in pixel intensity. Is it possible to put a tiny delay on hiding it, or some other solution?

will-moore:
I think the "Pixel Intensity" checkbox and "Interpolate" checkbox need to be moved out of the "Settings" tab since it's likely you'll want to toggle these when looking at other tabs, e.g .ROIs.
It would be nice to have them accessible on the image plane itself, but there's already getting too many buttons there and we don't have nice clear icons for these buttons. Needs a bit more thought...

waxenegger:
A box is queried. As is the distance horizontally/vertically is set to 25 pixels with the point in the center. So no line querying across any image apart from the data in the original files being a sequential stream of bytes but that's the way it is and no different to anybody else retrieving tiles/data for that matter.

The tooltip can only flicker if you don't stay on the point you'd like to see and the data is cached for that patch already as there is no delay in that case. The delay is for backend requests only so that I don't request while the user is not at the point they want to query (interims requests). If that's desired then I'll have to change the delays because I'd not wait the same amount for both cases. Why put in an artificially long wait when you have something ready for display. In the same vein I don't think it's beneficial to leave the tooltip for longer because what if you want to move into the direction where the tooltip is, that's not useful. I need to be able to determine that the user has reached the point they desire to query. I have to rely on them stopping mouse movement or else I fire off requests all the time.

I'm going to change the cursor tip and the the offset to see the image. The latter I actually had in there before but removed it (accidentally) in another commit.

waxenegger:
one comment on the supposed slowness of things. I use vpn also with the latency of being in Australia. The bundle loads in 3 seconds for me, the resource requests (css and images) are in the static folders hence separate requests. Maybe there was something impacting on the connection that moment for you. I have never had to wait even close to a minute.

I will look into splitting up the bundle more but don't expect great things since webpack is not too shabby with dependency handling ordinarily. The crucial issue is that it's going to need a lot straight away including the compiled ol3 viewer and jquery/ui. There are 2/3 candidates that wepback might single out for on demand loading but I'll have to test if that has much of an impact.

jburel:
I will not move the "Pixel Intensity" and "Interpolate" checkboxes to the image
This will be far too much. Discussed with @will-moore last week, and we could not find a good solution at the time
Thinking required...

will-moore:
The flicker I'm talking about is not due to loading. It's much faster than that, like it's hiding and showing as I move from one pixel to an adjacent one.

waxenegger:
As I said, it can only go show/hide in a rapid way when the values are cached in that area. Both the data has to be there and the mouse has to have been moved for that too occur. the offset will help once I put it back but if you don't want that rapid flicker I have to put in a tiny delay for even the area with the cached values. that's more in line with what goes on when the user leaves the patch of cached values.

will-moore:
Apologies for the confusion.
I was suggesting that there could be a tiny delay to the *hiding* of the tooltip to stop the flicker, so as you move from one pixel to the adjacent pixel, the tooltip is never hidden but simply updates the values.

will-moore:
The offset is a definite improvement. Maybe increase it a bit more? 20-30 pixels?
If the pixel data is all integers, it would be nice not to show 3 decimal places. Can these be formatted to integers?

will-moore:
If I have two channels active: 1 & 2 and I show pixel data for a region, then I turn off channel 2 and turn on channel 3 and return my pointer to the same region, now I only see the cached pixel data for channel 1 (no new pixel data is loaded for channel 3).

will-moore:
A suggestion for the "pixel intensity" button. Perhaps we should go back to having the button in the viewer itself, using the cross-hair, now that we have the association with the cross-hair cursor and pixel intensity. With a tooltip I think most users would find and remember what this does.
The coordinates can be shown in a separate label (not a button) and this should be hidden when we have no coordinates to show. This means we can move the "Interpolate" checkbox back down.

This avoids adding of a new menu, which we may have to do soon if we add more features, but I think it should work well for the current functionality.

screen shot 2017-09-28 at 10 52 28

will-moore:
Here's the crosshair.png if it's useful...
crosshair

waxenegger:
@jburel is this the quasi final decision on the UI look, well let's rather say position. I'd hate to go back and forth between control and settings panel. The code change is not so negligible, in the former case I can actually have all code in the ol3 part whereas the latter means events and exposure of toggle on/off methods on the viewer itself.

also the spatial division of the control bits (toggle button and coordinate display) means some bending of the self-contained control layout but ok, that's not such a big deal, as long as I don't have to toss it out of the window after the review.

jburel:
Sorry but @will-moore and I did not discuss since last week. I think we need some mockups before thinking out loud on this PR otherwise we are wasting valuable time

jburel:
We will still need to review the location of the components.
Let's prepare some mockup (including the multiview part) before changing anything
Merging so we can move forward

PR #117, opened 2017-08-25
Shape Stats

waxenegger:
Uses https://github.com/openmicroscopy/openmicroscopy/pull/5465 to get shape stats.

These stats are included into the csv download.
2 remarks:
- if there are are unsaved shapes the stats cannot be included (backend has no knowledge yet)
- since csv is a flat layout there will be repeated values for shapes and their corresponding channel stats


waxenegger:
--exclude

jburel:
the stats are correctly included in the csv file
It will nice to have them display in the table
One idea will be to:
 *  replace the "show area/length" by "show stats"
 * display the stats next to area, length.
 * add a "load data" button to reload the stats to avoid numerous calls

This could be done in a follow up PR

waxenegger:
I know that table seems like the place to put the stats that we can now query but there are a few points that are a bit problematic with that:

If I put aside the "minor" detail of space I have the problem that the table like csv is a flat layout on the shape level. The stats, however, are per channel so that would add same shape rows as is. The second concern (with one load button for all) is the potential high number we'd have to calculate for. 
The third comment is in regards to not yet queried stats for rows which would remain empty.

While it's not the preferred solution, what if we do a pop-up that contains the per channel listing for the selected stats. A slight variation could be a sort of tooltip when over the shapes in the table or a 'button'/icon next to length/area that triggers the display/request per shape.
The major difference to the "row based" table solution is that it does not clash with its flat nature, shows no void when not available and the querying is more selective (as with load all).

Regardless of what is decided eventually I'm going to make some efficiency changes similar to the intensity querying.

jburel:
Exporting the stats with unsaved shapes work but having negative id (internal) could be very confusing ``shape_id -208``
maybe "-" will be better

waxenegger:
makes sense. so I'd do that for both ids (roi and shape), right?

jburel:
yes for both

jburel:
* No negative ID for roi/shape not saved. This works
* I noticed an issue when exporting stats. Below are the steps I follow
 * Create a shape on a 3 channels image. All channels are active
 * Save the shape
 * Export the stats
 * Turn off channels
 * Export the stats. 
 * The stats for all channels are returned and not only for the active channels

waxenegger:
there is at the moment a bit of an inconsistency there, that's true. The active channel logic is, in fact, applied to requests but I disrespect it once I have cached values. I'll correct that.

jburel:
Active channels are not correctly picked
Channel column should be after ``z/t`` and not after ``area/length``

waxenegger:
--exclude

jburel:
Everything works as expected now
Thanks @waxenegger

jburel:
Large amount of rois handling will be covered in the next phases of work
cc @joshmoore 
Merging

PR #119, opened 2017-08-28
support well/dataset opening (no initial image)

waxenegger:
So far opening an iviewer without an image was not possible, in part because the image id was part of the path. Likewise opening wells was not supported and datasets without choosing an image.

This PR seeks to correct that.

TEST: http://web-dev-merge.openmicroscopy.org
Best way to test is probably to use the open with button in the right hand panel, at least for the well I'm not sure if there is another way, for datasets you can also use the dataset node in the left hand tree and open with via the context menu. If you wish you can try via the browser url directly.

open_with


waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#468](https://ci.openmicroscopy.org/job/iviewer-merge/468/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/468/consoleText) for more details.
Possible conflicts:
  - PR #102 waxenegger 'roi count'
  - PR #115 waxenegger 'save area/length measurements as csv'
    - src/viewers/viewer-context-menu.js
  - PR #116 waxenegger 'Intensity Point Query'
    - plugin/omero_iviewer/urls.py

waxenegger:
--exclude

plugin/omero_iviewer/README.rst
```diff
@@ -44,7 +44,8 @@ To enable the "open with" feature:
 ::
 
     $ bin/omero config append omero.web.open_with '["omero_iviewer", "omero_iviewer_index",
-      {"supported_objects":["image"], "script_url": "omero_iviewer/openwith.js", "label": "OMERO.iviewer"}]'
+      {"supported_objects":["image", "dataset", "well"],
```
jburel: it should be "images"
The readme was not updated when the code was changed

jburel:
Updated on web-dev
https://github.com/openmicroscopy/management_tools/pull/536

waxenegger:
@jburel not sure why travis is acting up, the last doc change can hardly be the reason and the last commit before that worked as well. There is nothing wrong with the unit test suite run where it fails. Actually it fails before its execution. Is this just some streak of strangeness or was there a config change?

node and npm versions seem to be ok, that's where I might have suspected a side-effect.

waxenegger:
@jburel ok, I have tried to get some more information from the build in regards to phantomjs, the test libarary used via mocha-phantomjs. So I know from previous issues (locally) that the latter it is very version sensitve and I suspect that to be the problem here too.

So it wants a 1.9.7 version of phantomjs. It seems that there is a 2.0.0 installed as an ubuntu package (not npm globally), not sure when and whether directly or through some other dependency (as an update) and I'm thinking that because there is system default it interferes with it.

I'm not a travis expert so perhaps you know more how we can solve this. Not sure how far privileges go (the travis file seems to indicate that sudo=false and changing it is probably not good - not sure) and what we can do to avoid that package/version of the packages being in the system global scope.

I suspect that way it is used by mocha-phantomjs it uses the global version rather than 1.9.7 which is local to the build directory.

here is some output I gathered...

```
$ npm list -g --depth=0
/home/travis/.nvm/versions/node/v6.11.3/lib
└── npm@3.10.10



The command "npm list -g --depth=0" exited with 0.
$ ls -la /usr/local/phantomjs
lrwxrwxrwx 1 root root 26 Aug 29 02:49 /usr/local/phantomjs -> /usr/local/phantomjs-2.0.0


The command "ls -la /usr/local/phantomjs" exited with 0.
$ phantomjs -v
2.0.0


The command "phantomjs -v" exited with 0.
$ npm uninstall -g phantomjs


The command "npm uninstall -g phantomjs" exited with 0.
$ phantomjs -v
2.0.0


The command "phantomjs -v" exited with 0.
$ ant unit-tests-only
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
Buildfile: /home/travis/build/ome/omero-iviewer/build.xml

unit-tests-only:
     [echo] Fetching test libraries ....
     [exec] npm WARN deprecated npmconf@0.0.24: this package has been reintegrated into npm and is now out of date with respect to npm
     [exec] 
     [exec] > phantomjs@1.9.7-15 install /home/travis/build/ome/omero-iviewer/node_modules/phantomjs
     [exec] > node install.js
     [exec] 
     [exec] PhantomJS detected, but wrong version 2.0.0 @ /usr/local/phantomjs/phantomjs.
```

jburel:
probably due to the recent changes in travis
I will need to investigate it

jburel:
* Select a dataset or well in web
* right-click and select to open in ivewer
* The dataset or well is opened i.e. thumbnails are loaded on the left-hand side but no image displayed in the centre panel
* In that case, the first image in the dataset or well should be selected an display to avoid an extra first click
This will be more inline with the opening of images

jburel:
Clicking on the "well ID" when viewing an image brings the user correctly to the web in the webclient

jburel:
Tested with dataset and wells.
The first image is now correctly selected.
I have also tested with an empty dataset.
Everything works fine

Merging

PR #118, opened 2017-08-26
reverse (intensity) -> invert

waxenegger:
following renaming efforts: https://github.com/openmicroscopy/openmicroscopy/pull/5470

TEST: dev-merge.openmicroscopy.org/webclient/ 
verify that the invert works by toggling the checkbox

PR #120, opened 2017-09-04
Multiple Viewer Mode

waxenegger:
This PR aims at enabling the viewing of multiple images within one browser window. 

Eventually this should be useful for locking/syncing some of the images' "settings" (compatibility permitting). Doing so will need to be analyzed for each "setting" in detail and with care since they span different view/templates, data objects and observers and event handling code (incl. history logic). The prevention of syncing cycles and side-effects is paramount.

As is one image can "follow"/be linked to another, i.e. zoom,pans  and z/t changes (slider) will happen in both. If the channel checkbox is checked the channels changes are propagated too (incl. gray). Compatibility is a must. In specific: no bounds for zoom/x/y/z/t/ and c are exceeded. They will be clamped to their natural limits.


waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#484](https://ci.openmicroscopy.org/job/iviewer-merge/484/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/484/consoleText) for more details.
Possible conflicts:
  - PR #119 waxenegger 'support well/dataset opening (no initial image)'
    - src/app/context.js

waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#496](https://ci.openmicroscopy.org/job/iviewer-merge/496/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/496/consoleText) for more details.
Possible conflicts:
  - PR #116 waxenegger 'Intensity Point Query'
    - src/viewers/ol3-viewer.js
    - ol3-viewer/src/ome/ol3/utils/Misc.js
  - PR #118 waxenegger 'reverse (intensity) -> invert'
    - ol3-viewer/src/ome/ol3/utils/Misc.js

waxenegger:
--exclude

jburel:
@waxenegger the branch has conflicts
when it is fixed you can list it for review

will-moore:
I noticed you removed the resize handles from the bottom-right corner of each frame, probably because I didn't include them in my mock-ups. Sorry - that was just my omission because I was focussing on other issues and didn't get around to adding them in, but I think they are useful and should stay.
I'll try to emphasise in the text the changes that the mock-ups are supposed to illustrate - there will often be other details that are incorrect but are too time-consuming to fix, but I'll try to keep them to a minimum.

jburel:
I had not play with the feature for a bit, it certainly looks nicer.
* handle is back
* ability to show/hide the controls on the image itself. Maybe they should be hidden by default.

Is the number of groups corresponds to the number of images on left-hand side
This list could become fairly long

Let's review the mockup next week and take it from there

will-moore:
Looking good now.
As discussed with @jburel this morning, additional multi-viewer functionality will come in follow-up PRs, but we really need to be sure there are no blockers to the feature as a whole before we commit to it by merging this PR...

There is the potential to get ROIs out of sync when I open the same image in 2 viewers. I can edit ROIs differently for the 2 viewers then hit save for each in turn. Now, both viewers think that they are displaying the "saved" ROIs but in fact the ROIs from one have been overwritten by the other, so that by closing the viewer I will lose those ROIs.
I don't know the best way to solve this issue. Should we try to sync the ROIs in the 2 viewers so that any changes in one show up in the other while editing without saving? That would probably be the nicest solution but not sure if it's possible. Or we try to sync viewers on Save of ROIs, but this still has the possibility of overwriting unsaved ROI changes in one viewer.
I guess some of the same arguments apply to channels too, since you can save 2 different settings from 2 viewers. But at least you can sync channels and the Saved channels are indicated by updating thumbnail. Also we don't ask users elsewhere to always Save settings so they're considered less critical.
Needs discussion.


Minor fixes for this PR:
 - When each viewer opens, initial size of window could be square instead of rect. Simply make it's height equal to current width would be fine
 - Hide zoom etc controls by default when viewer is opened
 - For sync select, use "No sync" instead of "None". Can consider switching to icons etc in follow-up.
 - Combine the Z and T sync checkboxes into a single Z/T checkbox.

Follow-up PRs (see https://github.com/openmicroscopy/design/issues/79)
- Use double-click to open multiple viewers and remove "multi-viewer mode" (see 
- Layout helper "grid" button
- Lock icons for the "sync" select.
- CSS tweaks to centre the Z/T sliders etc in their whitespace and try to remove the whole row of whitespace for the resize handle?

waxenegger:
The encapsulation in terms of code and data is the very basis for enabling multiple windows. Syncing is triggered by an action and nothing but the action 9 + necessary parameters is communicated. This is why I feel very comfortable that there are going to be little side-effects. Also windows for same images are not kept track off for that same reason, for all intents and purposes they are treated as their own instance of an "image_config" underneath.

So for persistence the envisioned solution for mdi should be update/refresh after save of settings/rois if there happens to be more than one "image_config" referring to the same image id . That does not mean that I couldn't look into syncing rois actions (though I cannot give a guarantee about what is possible at this point) but I'll not make the mistake and connect the two notions of linking actions (which is what happens really) and sharing of state (a big no in that setup).

waxenegger:
--exclude

waxenegger:
@jburel I have a commit prepared to disable the switching to multiple viewer mode.

Of course we can keep this long-running PR open longer and I can exclude this or others depending on the needs/deploy and merge changes either way once a conflicting one gets merged but I'd rather like to propose a strategy that gets this PR merged with mdi disabled. Despite code changes, continuing to use the old single image view should not be problematic and it makes life easier to start new PRS and get in others while having a new multiple image view PR on the side which can deal with improvements and other changes that are still to be made. 

Does that sound reasonable?

jburel:
The PR is getting long already so I am more in favour of small PR
My point is that we need to make sure that we reach a "stable" implementation for multi-view, it does not have to major surprises down the line.

The ROI case needs to be discussed so we do not confuse users.

jburel:
Discussed today with @waxenegger and @will-moore:
 * prepare mockup for roi synchronization between viewers. This is not for this PR and probably not for version 1.0.0
 * in this PR, prevent the lost of ROIs by saving the shapes when the user selects a new image in the multiview
* RFEs  indicated on this PR have been moved to https://trello.com/c/n560sMVS/8-multiview

joshmoore:
NB: discussed wth @rgozim et al today the ROI design issues that are still in https://github.com/openmicroscopy/design.  Might be a good time to talk through them, and as @dominikl pointed out clarify the final decisions on them.

jburel:
@will-moore will update the card https://trello.com/c/wOEjcK3Y/17-roi-sync-between-multiple-viewers with some "synchronization" scenarios and uses cases
I think a new issue if needed should be created for ROI so we can have the issue above linked to the first version (so info might need to be extracted)

will-moore:
All my "Minor fixes for this PR" requests above are working fine.
ROIs save dialog when moving between images is kinda lengthy but I guess it does the job.

With one image with unsaved ROIs is open (not currently selected) and another viewer of the same image is opened, we don't the the Save warning immediately, so it's still possible to end up with out-of-sync ROIs in the 2 viewers, but this is a kinda edge case so I don't think it's a blocker for this PR.

jburel:
last comment added to https://trello.com/c/n560sMVS/8-multiview 
Merging so @waxenegger can move forward with other PRs

PR #122, opened 2017-09-14
use chrome headless with karma instead of phantomjs aka travis is a pain

waxenegger:
Travis decided to stumble at the beginning of unit-tests from one moment to the next. The first commit that triggered it was a README file 1 character addition so it is safe to say that no code change triggered this.

What followed was way-too-long an investigation that led to nothing else but the conclusion that the mocha default compiler had stopped being happy with the js input and threw a syntax error while parsing. This did not happen in the browser setup nor locally with the versions of node, npm, mocha, chai, phantomjs and mocha-phantomjs being synchronized. My best guess is that some other dependency that mocha used was different since it's more difficult to conceive that the same closure compiler (for month) would all of the sudden have started producing different output (without change to the code...).  

A solution was found by forcing mocha to use babel for compilation (which saves the closure compiler step) which necessitated the change of frameworks from mocha-phantomjs -> karma/chrome-headless since the former won't let you adjust the  compiler options for mocha. Probably not a disadvantage since phantomjs is not really a very active project any more it seems.

TEST: Feel free to execute the unit tests locally by checking out the branch and running ```ant unit-tests``` or be satisfied by the fact they run without error via travis.


snoopycrimecop:
Empty PR description. Please add a short summary of the PR scope and some testing instructions.

waxenegger:
@snoopycrimecop I prefer cats

jburel:
That was an effort!.. Travis is green :-)

will-moore:
Looks good...

```
$ ant unit-tests
...
     [exec] HeadlessChrome 0.0.0 (Mac OS X 10.10.5): Executed 30 of 30 SUCCESS (0.04 secs / 0.021 secs)
     [exec] TOTAL: 30 SUCCESS
     [exec] 

unit-tests:

BUILD SUCCESSFUL
Total time: 1 minute 5 seconds
```
And travis is green 👍 
Good to merge.

jburel:
merging so @waxenegger can be the other PR green again

PR #123, opened 2017-09-18
open wells/dataset follow up

waxenegger:
In https://github.com/ome/omero-iviewer/pull/119, when opening the first image by default, I missed that refreshing the thumbnails (click on the refresh icon) will also execute the code path of the first image to be opened. This is not desired behavior as the user just wants to refresh the thumbnails and not change image.

TEST: Open a well/dataset and check that the first image is opened by default. Open another image by clicking on it in in the dataset/well thumbnail list. Once displayed hit the refresh icon at the top of the thumbnail panel. The thumbnails should reload but the viewed image should not change.


jburel:
Good catch I missed that too

jburel:
Thumbnails are reloaded and selection not changed
Difficult to say if this the side effect
* log in as user-1
* open plage 2503 in webclient
* Open first well in iviewer
* all thumbnails are blacked. but this is not the case in webclient

will-moore:
The fix in this PR is working fine.
But I found a couple of issues with "openwith..."

If I've selected an image in a Well, e.g. http://web-dev-merge.openmicroscopy.org/webclient/?show=image-51717 (user-3)
Using the Open-with -> iviewer in right panel takes me to http://web-dev-merge.openmicroscopy.org/iviewer/ with an error message.

From a Dataset image, e.g. http://web-dev-merge.openmicroscopy.org/webclient/?show=image-76330 Open-with takes me to http://web-dev-merge.openmicroscopy.org/iviewer/?dataset=17803 and the image I selected is ignored and is not even available for me to see.

will-moore:
Testing locally the ```openwith`` is now giving me ```/iviewer/?images=76320&dataset=17803``` which looks good. But having built the app locally (```ant``` then ```npm run debug```) doesn't work in the browser so I can't test the rest. html and statics seems to load OK.
I see no errors in Chrome, but this in Firefox:
```
Unhandled rejection No PLATFORM.Loader is defined and there is neither a System API (ES6) or a Require API (AMD) globally available to load your app.
```
It's been a while since I last built it. Any ideas?

waxenegger:
Wiping the node_modules directory in the project root is highly recommended if you haven't built in a while, especially with the aurelia/ol3 updates that took place. 

Also double check that you have, at a minimum, versions 6.x.x or greater for node (and 3.x.x or greater for npm).

will-moore:
OK, that worked, thanks.
So, testing just that branch locally is working pretty well.
Only found 1 minor bug.
If I select a large number of images (more than 10) and open iviewer ```/iviewer/?images=4417,4418,4419,4420..... etc ``` then I browse to the second 'page' of thumbnails and select an image to view. Now, when I refresh the thumbnails it returns to only showing the first 10 thumbnails, not the 'page' that I'm currently on, so the currently viewed image is not shown in the thumbnails.
I get the same behaviour if I open just a Dataset in iviewer ```/?dataset=1301```.

waxenegger:
The multiple images case is such that the refresh is quite different from the other cases. I did never keep any page number in memory, only the present image viewed. Compared to the other cases of open with there is also no backing backend request (for dataset/well children) since the list is assembled based on user input (maybe address bar input even) and could be inhomogenous. So I'll need to put the code in place to remember the page if that's something that is desired, along with the image not changing.

As for the dataset I have to check what happens there. I would have thought the page nr should be considered after requering the list.

will-moore:
Related to refresh, I wonder if you could use the Bootstrap 'refresh' or 'repeat' icons from http://bootstrapdocs.com/v3.3.6/docs/components/ since they look much nicer and you don't need to include a ```refresh.png```

screen shot 2017-09-20 at 09 19 10

waxenegger:
don't mind glyphicons are easier, and yes, in that case it looks nicer.

will-moore:
I'm not familiar with the code and I'm sure there's a logic to what you say, but if you have ```?images=1,2,3,4,5,6,7,8,9,10,11,12``` and user is viewing image 12, then when you refresh you just need to show thumbnails ```11 and 12```? You don't see that you need to store a ```page``` variable.

waxenegger:
whether you want to call it page or start or something totally different doesn't really matter. whether you achieve it by an array/list scan in one case or a backend request in the other is a technical detail. what's for sure is that the start is needed. what's also for sure is that the thumbnail panel is handling many more cases than it did initially and so my main main goal is to converge as much as I can code-wise and not create special code paths if I don't have to.

jburel:
I think we should not try to tackle the minor refresh issue in this PR
What might work for images, might not work for datasets or wells
so side-effects could be introduced 
I have created a card to capture the comment https://trello.com/c/QZCP3Rp6/9-refresh-thumbnails

jburel:
Icon looks better now
Thanks merging

PR #126, opened 2017-09-27
webpack loading - revisited

waxenegger:
Concerns have been raised about the bundled size of the js in https://github.com/ome/omero-iviewer/pull/116. While it's difficult to conceive that the long wait in the case described is due to the bundle size, there is some room for improvement as far as loading for resources goes.

A few points to clarify size and what/how much can be done:
- the dependencies that are used are not always included in their entirety, i.e. d3 and jquery-ui where only the relevant bits are pulled in via the import statement.
- the dependencies that are big are almost all needed instantly. in such a case there is no gain in separating them out. this is especially true for the compiled ol3 viewer.
- the one exception from the above is "text-encoding" which is used to be able to produce an excel compatible csv file (win-1252 charset)

This PR splits up the bundle into 3 components: a webpack loader init.js, a main.js containing the immediately needed code and dependencies (aurelia, ol3-viewer & jquery - as well as their child deps) and a deps.js which contains d3, file-saver and the big text-encoding lib. The latter is included in the html with a defer attribute which should affect a later evaluation by the browser (not download!). Given that script includes are sequentially dealt with by the browser the main.js should be evaluated before which means that aurelia can meanwhile do its bootstrapping.

To cut back on the css requests -there were 7 (3 jquery-ui, 1 bootstrap, 1 spectrum colorpicker, 1 ol3 and 1 aurelia app related - as well as their size, they have been concatenated into 1 (and processed with cleancss to reduce them ever so slightly) as part of the build process.

The webpack version was also raised to 3.6.0. The 2.7 version was rather old and 3.6.0 is said to include some improvements. It's definitely preferable for the debug version. While in 2.7 the source map option was unbearably slow it can be used again in 3.6.

TEST: check that everything (js, css, glyphicons, icons) loads propery. the dev-tools will show a summary of the requests and their size and duration.


will-moore:
Thanks for looking into this.
Everything is working fine and seems to be a bit quicker (testing from home again today) so I think it's good to merge. For the record, this is what I'm seeing...

screen shot 2017-09-28 at 11 33 50

A couple of thoughts if we wanted to optimise further (don't know how feasible these are and I'm not suggesting that fast loading over low bandwidth is a priory feature)...
 - If we wanted to show "something" on screen ASAP then the UI framework could be rendered without needing to wait for the ol3 viewer?
 - Is it possible to kick-off the ```deps.js``` download "after" the ```main.js``` has arrived so that they're not competing for bandwidth?

Anyway, good to merge.

waxenegger:
Thanks for the screenshot. Ok, I have misinterpreted the defer, from what I see and read here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

I was under the impression that it'd not stall the process but it does because its evaluation by the browser has to be awaited before DOMContentLoaded. Doesn't matter that parsing and other things happen meanwhile. Clearly aurelia does not bootstrap sooner.

Maybe the async attribute can help though I'd also suspect the download to happen straight away. It only wouldn't if I inserted the script into the page at the point that aurelia has bootstrapped. That's, however, dangerous I think. If it runs through the nested template/view code and encounters an import statement, it'll want to load the code and if it's not there at that point there's going to be an error.

Let's try and see what async does and I'll also locally experiment with script injection.
Webpack by itself does, supposedly, offer dynamic loading via require callbacks but I don't think that fares well with the aurelia framework. I would not know how to merry the 2 given that aurelia uses its own 'webpack plugin' and mainly ES2015 import style which is inherently static.

PR #127, opened 2017-09-28
tiled image tests

waxenegger:
this PR is for testing, to illustrate the effect that some measures have.

for a start the 8 in between resolution levels have been halved.
also, as opposed to choosing the next closest resolution level, the lower one is chosen always unless we hit a natural resolution level. This means that there is no such case when tiles get smaller (as in the case we snap to the higher resolution). Unfortunately this also means that quality is lost when the naturally lower resolution tiles need to be made bigger.


waxenegger:
This is not the way to go. 

https://github.com/ome/omero-iviewer/pull/130 is more promising, in conjunction with a tile size change for in between resolution level zooms ...

PR #133, opened 2017-11-01
Info - redundant unbind

waxenegger:
I noticed that the info page code had the unbind() method defined twice (perhaps an accident while merging?). Due to  unbind()'s role this did not have a noticeable impact on the app but should be corrected nonetheless.

Test: Check that the info tab displays properly (web-dev-merge.openmicroscopy.org)


waxenegger:
--exclude

jburel:
Thanks for tracking that one down
Info tab works as expected

PR #134, opened 2017-11-14
Header Menu

waxenegger:
Adds a menu to the header.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -1215,10 +1215,9 @@ thumbnail-slider img {
 }
 
 .mdi-icon {
-    float: left;
     margin-left: 10px;
     cursor: pointer;
-    top: 8px;
+    top: 5px;
     font-size: 22px;
 }
 
--- a/css/app.css
+++ b/css/app.css
@@ -175,7 +175,9 @@ a.ui-button:active,
     background-color: #1F6AAA;
     border-width: 0;
 }
-
+.fixed-header .dropdown-menu {
+    min-width: 100px;
+}
 .fixed-header div {
     margin-left: 20px;
     margin-right: 20px;
@@ -1173,7 +1175,7 @@ thumbnail-slider img {
     top: 45%;
     left: 40%;
 }
-.modal-about div, .fixed-header li {
+.modal-about div {
     text-align: center;
 }
 
--- a/css/app.css
+++ b/css/app.css
@@ -1122,3 +1122,15 @@ thumbnail-slider img {
     display: inline-block;
     padding-left: 0px;
 }
+
+.modal-about .modal-content {
+    width: 550px;
+    max-height: 220px;
+}
+.modal-about .modal-dialog {
+    top: 45%;
+    left: 40%;
+}
+.modal-about div, .fixed-header li {
+    text-align: center;
+}
--- a/css/app.css
+++ b/css/app.css
@@ -169,25 +169,30 @@ a.ui-button:active,
 .fixed-header {
     position: fixed;
     margin: 0px;
-    overflow: hidden;
+    white-space: nowrap;
     min-height: 40px;
     max-height: 40px;
     background-color: #1F6AAA;
     border-width: 0;
 }
 
-/** Text displayed in the header */
+.fixed-header div {
+    margin-left: 20px;
+    margin-right: 20px;
+    display: inline;
+    line-height: 40px;
+}
 .fixed-header-text {
     text-align:center;
     font-size: 22px;
     color: white;
     margin-top: 5px;
 }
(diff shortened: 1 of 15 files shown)


waxenegger:
--exclude

jburel:
it looks better 
the multiview button will need to be reviewed, a toggle button "Multi View" will be more explicit. Its location should be after the menus I think

waxenegger:
@jburel Should we move the button here or get rid of it altogether (see trello link in https://github.com/ome/omero-iviewer/pull/147) ?

jburel:
My preference will go for a toggle button with Text rather an icon. ``Multi-View`` might be a bit long but we have some space (especially if we remove the name later on)
In any case, I think it should be after the menus

waxenegger:
ok, I'll do that for the next review round.

pwalczysko:
screen shot 2017-12-14 at 14 23 43
This should read ROIs (capital O, capital I)

pwalczysko:
screen shot 2017-12-14 at 14 24 39
Can you wrap the link behind a header (maybe "OMERO.iviewer on Github") ?

pwalczysko:
screen shot 2017-12-14 at 14 27 05
Rather align the menu items to left, than center

pwalczysko:
screen shot 2017-12-14 at 14 27 05
Unfortunately, this very menu is displayed on Windows 7 Firefox. I suppose the idea is to display ``Ctrl+C`` etc. there

waxenegger:
@pwalczysko strictly speaking, operating system detection in browser's should not be relied on (see: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform) but for that purpose there is not much of an issue I suppose...

pwalczysko:
screen shot 2017-12-15 at 10 30 10
This improved, but could that link please not be blue and underlined, instead, the menu item should look like the first one "About OMERO.iviewer" item.

pwalczysko:
screen shot 2017-12-15 at 10 33 55
Not sure if this change was done in this PR. I had a problem to understand at first whether the button notifies the state I am in or the state I could go to. 
Maybe if you just put "Go to Multi view" and analogously for Single view ?

pwalczysko:
The change of the formatting of the github link is working as expected. On second thought, because there is a tooltip on the "Mulit view" / "Single view" button, I think we can actually go with that.

Ready to merge FMPOV.

jburel:
I do no see any change when clicking on the multi-view/single-view button
The text changes but not the UI.

Now that we have the menu, the name in the header will be too much.
We can think about it and change it in another PR

pwalczysko:
>The text changes but not the UI.

@jburel This works for me. Maybe: When clicking on "Multi view" first time after you open an image in iviewer (which means you are most probably in Single view), you will not see any change immediately. You have to click on another image in order to get the Multi view. After that, two panels will open (because you are in Multi view now). When selecting back "Single view", you will see that the two panels vanish and just one is displayed.
Firefox on Mac for me.

pwalczysko:
@will-moore is also getting confused by the fact that "Multi view" is displayed when he is in the Single view. We probably should revisit this in another PR

jburel:
I should have clarified what I meant by "don't work"
if the multi-view button is clicked on, the centre panel should show change the ui to the one display when 2 images for example are selected. This will avoid confusion

waxenegger:
@jburel After talks we have had and the your coment in this PR I thought the envisioned action was to convert the icon into a button because you said it would make it more explicit,

Doing so there was no change whatsoever in multiple view and with how windows are opened. Are you saying that this should change now? There is also still this other way of opening out there (by means of click), so what are we settling for or are we changing it now to then change it again...

I'm not sure, maybe you are suggesting that I should show a small image window the very moment one switches to multi view as opposed to it filling the entire space in the middle. Is that so?

waxenegger:
@jburel I have pushed a commit and deployed it on web-dev-merge. Let me know if that's what you are after or if you had something else in mind.

The approach that is presently in the design docs is best reserved for this [PR](https://github.com/ome/omero-iviewer/pull/147)

Already too many PRs open that require careful manual merging from the looks of it.

jburel:
Turning the icon into button is more explicit
While looking at the functionality from a user point of view not knowing the feature. I realised it could be confusing not to change the display
I will test the feature when I have a chance.

pwalczysko:
@waxenegger Note https://github.com/openmicroscopy/design/issues/81 - with @will-moore we agreed on this strategy before christmas.

jburel:
The suggested improvement should be done in another PR
Merging this PR so we can include other PRs and review them

PR #135, opened 2017-11-17
Intensity Query - On/Off Toggle in Control

waxenegger:
see: https://trello.com/c/imsCSShL/10-intensity-point-control-location

At the moment the intensity query can be turned on/off by clicking the cross icon on the control. When the mouse cursor is within the image extent the control will display the coordinates, otherwise nothing.

The intensity values are displayed next to the cursor once queried. Only intensities for active channels are queried and displayed.


waxenegger:
--exclude

waxenegger:
--exclude

pwalczysko:
Is the checkbox in this PR still on the Settings pane ? During the workflow, we found out that the place to have it is the ROI pane, because that is where you would use it from

jburel:
It is no longer on the pane @pwalczysko

waxenegger:
well, it's not going to be in the roi pane since the last time this was discussed, it was suggested it go into the controls again.
the thing to note though is that intensities are in no way depending or related to rois per se and rois will be loaded if one clicks the roi pane so that if somebody just wanted to get to the intensities under the cursor they'd always have the rois there as well.

pwalczysko:
understood

pwalczysko:
@waxenegger : Please deliver at least a basic testing instructions. I have read the trello card, but this is rather a free-flowing discussion than a statement of what we should see now in the UI.

pwalczysko:
screen shot 2017-12-14 at 10 30 14
1. I find the numbers in this box far too small, and very light grey on white, so hard to read.
2. before the ``+`` sign is clicked, I was wondering what does the empty box to the left of it mean, and why does it take so much space. After I clicked the plus, it became clear that the box accomodates the coordinates. Could the box be displayed only after I have clicked the plus ?

pwalczysko:
screen shot 2017-12-14 at 10 41 32

When the numbers come out with the second line being 2-digit after the first line being 3-digit, the formatting is a bit odd (the second line is strangely indented, see above)

waxenegger:
@pwalczysko thank you for the review

I'm going to increase the font for the coordinates in the box. It means that I have to make the box bigger which means in multi view mode the individual window will need to be bigger which is noticeable if your screen is smaller -as is mine- but it shouldn't hurt too much I suppose.

I can look into hiding the box if the cursor is not within the image extent and therefore not showing the image coordinates but it would not give more space since I have to have that space there when I need it unless we skip the display of coordinates altogether, or put it in the popup (intensity display).

I'll also look into the alignment issue.

waxenegger:
I have increased the font for the coordinates display. To not need to make the box bigger I omitted the X and Y in front of the coords but if the general consensus is that that's a bad idea then I'll add them back in and make the box bigger.

pwalczysko:
The box for the channel intensity (on the image itself) is now very nicely formatted and gives a good experience.

The X, Y position box: 
  - much easier to read
  - X, Y should be explicitly shown, taking the ImageJ as example, also, it is not so terribly obvious which one is which
  - if the others think that X,Y is not necessary to show, then please make a space at least after the comma after the first number

pwalczysko:
Great improvement, easily readable now.

Ready to merge.

jburel:
we could try to use a glyphicon in another PR
merging

PR #136, opened 2017-11-21
Acquisition & Import Date Display

waxenegger:
see: https://trello.com/c/Dse1xHso/5-acquisition-date-in-info

The info tab displayed what was the import date at times (plus it omitted the time).
With the changes in this PR the handling is analog to the web client down to the image methods used to get the respective dates:

The import date is displayed in full always. The acquisition date (incl. time) if exists.

TEST: http://web-dev-merge.openmicroscopy.org


jburel:
Tested with images with and without acquisition date
everything works as expected.
Including this PR in the coming 0.4.0 release

PR #138, opened 2017-11-22
Release 0.4.0

jburel:
Prep for release 
cc @sbesson @waxenegger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,13 @@
 
 This version includes:
 
-- displayed acquisition date when available
+- displayed both import date and acquisition date when available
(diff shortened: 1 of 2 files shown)


jburel:
where are the questions?

sbesson:
Apologies @jburel, I managed to save my review without the associated inline comments. The two questions were:

- for the first item, should we mention import and acquisition dates as both are supported
- for the last item and generally, would it be useful to specify which version of Openlayers is used?

jburel:
@sbesson adjusted

jburel:
https://github.com/openmicroscopy/www.openmicroscopy.org/pull/159 has been merged

Merging to prep release

PR #142, opened 2017-11-30
Webpack Dev Server Setup - Reincarnated

waxenegger:
With increased public interest in setting up iviewer for development and the first choice being webpack dev server 'mode' I revisited ```npm run dev```.

Turns out there is handy webpack devserver proxy functionality so that the running of scripts/altering of webserver configurations is not necessary.

With this PR one can just run the above command and then navigate to http://localhost:8080.
Java Script and Css is served by the dev server, so hot deploy will work -give or take an occasional browser refresh that might be necessary.

The devtool settings is set to _eval-source-map_. This makes for slightly longer initial/incremental builds but gives a full source map. I'll leave it up to the developer to change that as they wish: https://webpack.js.org/configuration/devtool/

TEST: Has to be done locally. So, anybody who has checked out the project and built and deployed it at least once on their local machine is in good shape to use this branch and test (see instructions above)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,7 +64,7 @@ Either development setup expects the following:
 
 **Notes**:
 
-The webpack dev-server config expects a local OMERO server at http://localhost.
+The webpack dev-server config expects a local OMERO server at http://localhost (default port 80).
 Should the server instance use a different port you will need to modify all
 proxy target entries in `webpack.dev.config.js <webpack.dev.config.js>`_:
 
@@ -89,6 +89,29 @@ you will need to change its port property in `webpack.dev.config.js <webpack.dev
     devServer: {
         port: your_port
     }
+
+
+The initial data type (e.g. image, dataset, well) and its respective ID can be set/changed
+in `index-dev.html <src/index-dev.html>`_:
+
+.. code-block:: html
+
+    <html>
+        <head>
+            <link rel="stylesheet" type="text/css" href="build/css/all.min.css" />
+
+            <script type="text/javascript">
+                // modify according to your needs
+                // in particular: choose an existing id !
+                window.INITIAL_REQUEST_PARAMS = {
+                        'VERSION': "DEV_SERVER",
+                        'WEB_API_BASE': 'api/v0/',
+                        //'IMAGES': "1",
+                        'DATASET': "1",
+                        //'WELL': "1"
+                };
+            </script>
+    ...
 
 Documentation
 =============
--- a/README.rst
+++ b/README.rst
@@ -43,13 +43,27 @@ please read the section *ol3-viewer* below!
 Development
 ===========
 
-Run:
+Building the uncompressed version of iviewer as described above the application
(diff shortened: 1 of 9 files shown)


waxenegger:
--exclude

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#586](https://ci.openmicroscopy.org/job/iviewer-merge/586/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/586/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - README.rst

--conflicts

jburel:
* Both server and web needs to run, this should be indicated in the readme
* in ``index-dev``, web might not be running on port 8080.
* Indicate to change modify the id in ``index-dev``, it is assumes
if web is running on a different port. It will not work despite changing the port in index-dev

jburel:
in ``web-dev.config`` is no longer needed
```
 devServer: {
    host: 'localhost',
    port: 3000
  },
```

waxenegger:
@jburel I'm going to add some lines in the README for clarification. 

Generally speaking people shouldn't be adjusting the config which is why I didn't want to encourage it via the documentation. In the past that lead to constant back and forth of questions. The setup is already so non-standard (as in pure webpack development server) and change of ports are probably not the best idea for the purpose of a development setup. That way people can use the standard omero server install or from github source build with an install via the admin guide and the will web accessible at default http 80. On some systems changing ports will also mean that the person who does it ensures that the port is not taken, within the allowed range, selinux allows it as well, etc. The choice for 80 (web server default) and 8080 (webpack dev server) was the most conservative one, I didn't want to cause more confusion than necessary.

jburel:
Could you indicate to change ID in the index-dev?

After adjusting the port as mentioned earlier it works very well

README.rst
```diff
@@ -58,12 +58,37 @@ Either development setup expects the following:
...
 ::
+
     $ export PYTHONPATH=$PYTHONPATH:/path/to/iviewer-project-root/plugin
     $ bin/omero config append omero.web.apps '"omero_iviewer"'
 
-**Note**:
+**Notes**:
+
+The webpack dev-server config expects a local OMERO server at http://localhost.
+Should the server instance use a different port you will need to modify all
```
jburel: mention the default?

README.rst
```diff
@@ -35,21 +35,83 @@ which can then be used like any Django plugin.
...
 ::
- 
+
     $ npm run dev
 
-To be able to run the ``webpack-dev`` server locally at: ``localhost:3000``
+will build the bundle and start the webpack dev-server (localhost:8080)
+
+Either development setup expects the following:
+- an OMERO server installation
```
jburel: the bullet list is not correctly rendered on gh.
I think you need to add a line before the first item

jburel:
Thanks for the update

jburel:
Tested  the option to run webpack on a different host
Everything works fine.
Merging

PR #130, opened 2017-10-11
Tile size increase

waxenegger:
As part of some more investigation into the problem described in https://trello.com/c/lWWQS0Vg/4-server-side-load this PR can help to reduce problems. It relies on: https://github.com/openmicroscopy/openmicroscopy/pull/5542

This in only one step in the right direction, there are other reasons/causes for a larger number of tile requests and therefore higher server load. Problem is that in order to address them an iviewer alone solution does not suffice.

PR #144, opened 2017-12-06
Quantization Support

waxenegger:
see: https://trello.com/c/aSMC1nBY/9-expose-advanced-rnd-settings-control

relies on: https://github.com/openmicroscopy/openmicroscopy/pull/5592

includes the UI bits now. the advanced rendering settings can be turned on per channel:
image


waxenegger:
--exclude

joshmoore:
What would it take to not have this PR tie the iviewer version to a specific OMERO.web version?

waxenegger:
@josh there is already work done and merged (the PR referenced), there are other dependencies and there is a tile size PR that also needs the web. I guess what I'm trying to say is that iviewer does have dependencies and always will. 

Of course I realize it's desirable to keep them minimal and I have already included the web api for rois and eventually we can think about the other big image_data request (the one that is kind of a "bootstrap"). In general I work towards an as much decoupled version as possible but there are limits and especially for things like stats and such. For the tiling I'd have to do the entire rendering stuff in the view.

To answer your question, it's a major pain because that workflow with the corresponding requests for rendering settings would be diverging and, frankly, I'd have to ditch the copy and paste and saving workflow as is and replace everything altogether. I'd have to do so in the python view and there would yet again be a dependency on the blitz gateway.

joshmoore:
Understood that it's a pain and it might be that we aren't yet there yet, but I'd rather see options in iviewer being disabled if, e.g., talking to an OMERO 5.4.1 rather than having it impossible for it to do so.

waxenegger:
Well the PR doesn't have go in, in fact I don't see how it can go in without me making major changes, sadly a disabling is not really a solution in this case. I'd have to disable copy and paste generally or people will wonder why the additional quantization settings won't get copied. I can not do imported if it happened to have these settings on them because the request I used for that purpose does not include it either. That's basically my dilemma.

waxenegger:
@joshmoore I come to think of it I might have actually misunderstood you entirely. 

So are concerned with backward compatibility or the code changes in https://github.com/openmicroscopy/openmicroscopy/pull/5592 as a means to add the missing bits to get that functionality into iviewer?

Because for the former I'm pretty sure I can find an easy way of disabling it even without an explicit version check (which I'd rather avoid because it could start bad practice of having ifs scattered all over). 

But to enable the quantization options in iviewer altogether without the changes in the other PR (regardless of what server release it would be in) would be a bit of a pain.

joshmoore:
I'm thinking in terms of "capabilities":

| iviewer | web | behavior                            |
|---------|-----|-------------------------------------|
| old     | new | should ignore the new json elements |
| new     | old | should gracefully disable features  |
| new     | new | obviously optimal                   |

waxenegger:
@joshmoore thank you for organizing the cases into a table, makes it easy for me.

case 1 is not a problem. the additional information won't be used.
case 2 I just had a look and I'm going to handle it like inverted, so no problem either.

Which brings me to case 3 and my initial response/worries and the existence of this PR to begin with.
I wanted to make sure that I have all the bits I rely on to add this functionality to iviewer and I figured the best way to ensure that is to actually implement what's needed.

This PR will probably remain open for longer at any rate, nothing's been decided on the location and looks of the UI to tweak the the settings but being roughly aware of the omero server releases I wanted to deal with this dependency (https://github.com/openmicroscopy/openmicroscopy/pull/5592) as soon as possible so that this feature can find its way into iviewer as soon as possible.

In the same spirit https://github.com/ome/omero-iviewer/pull/130 requires https://github.com/openmicroscopy/openmicroscopy/pull/5542 as one measure to address this: https://trello.com/c/lWWQS0Vg/38-server-side-load. Again, iviewer alone won't be able to do the job.

waxenegger:
--exclude

src/utils/converters.js
```diff
@@ -208,4 +208,29 @@ export class Converters {
 
       return shape;
     }
+
+    /**
+     * Takes a css color string in hex notation and returns the RGB int  values
```
mtbc: double space

jburel:
Discussed today:

 * Disable Gamma for Linear and Logarithmic. The value is not taken into account by rendering engine
 * Show the "map" widget by default when non linear is saved
 * @will-moore to test with image with large number of channels to test the URL length

will-moore:
Just looking at the length of query strings, each channel in the  ```map``` JSON list is now about 80 characters:
```
{"inverted":{"enabled":false},"quantization":{"family":"logarithmic","coefficient":1}},
```
Each channel is also about 20 characters in the form:
```
1|80:203$00FF00,-2|83:1277$FF0000
```
So we are at about 100 characters per channel.
Previously (without ```quantization``` in the ```maps``` query string) we were at about 50 characters per channel.
So, this PR will approximately halve the number of channels we can support before we need to change nginx's settings to allow longer query strings.
Currently we know the limit is below 70.
Could test with real images to get a more accurate numbers, but that should give a good approximation.

One possibility (as discussed) is to only add the 'quantization' to the maps query string IF it's not linear. 
The same is possible with 'inverted' setting, so this could be done in another PR, or this PR if not too much work for quantization.

jburel:
I have noticed a problem with Save to All. Not sure if it is related to this PR or another backend problem
This requires further investigation

jburel:
Tested: 
In both cases view an image with map=linear and map=polynomial with gamma !=1
* server 5.4.0, iviewer with this PR: controls not accessible: as expected
* server 5.4.5, iviewer without this PR: no controls. everything works as expected

waxenegger:
@jburel yes, "Save to All" just posts to `webgateway/copyImgRDef`. I have no control beyond that coming from the web.

src/settings/channel-range-advanced.js
```diff
@@ -0,0 +1,260 @@
...
+     */
+    GAMMA_MIN_MAX = [0,4];
+
+    /**
+     * the families allowed
+     * @memberof ChannelRangeAdvanced
+     * @type {Array.<string>}
+     */
+    families = [
+        'linear',
```
jburel: ideally that should be retrieved from the server (enum)
This could be improved later one
add a TODO for now

jburel:
@waxenegger I have added a card for the save to all. Thanks.
Minor comments (typo and add a todo). Everything is working. I tested last night various combinations and no problem noticed

jburel:
* The families are correctly loaded. Thanks

Small problem found:
* Select an image, change the family to polynomial with a gamma value != 1.
* A new image is rendered
* Switch to linear, the image is not rendered

waxenegger:
@jburel I tried very hard to reproduce that but couldn't. 

I had the dev tools open on the network tab, filtering for image requests. The only case where I didn't see new requests was when the very same url params had been requested previously.

jburel:
I think I understood what happened. I was toggling between various mapping options expo/poly etc
and change the gamma and got confused by a change between poly and exponential (with gamma< 1)
Everything works as expected

PR #146, opened 2017-12-11
Ol3 Upgrade

waxenegger:
PR raises version to 4.6.3 and adds/changes code necessary to react to ol3 changes.

As usual it's a good idea to have this PR around for a while to see if there are any side-effects.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -7,7 +7,7 @@
 
     <property name="version.major" value="1" />
     <property name="version.minor" value="0" />
-    <property name="ol3-version" value="4.4.2" />
+    <property name="ol3-version" value="4.6.4" />
 
     <property name="closure-compiler.jar" value="${basedir}/closure/compiler.jar" />
     <property name="closure-calcdeps.py" value="${basedir}/closure/calcdeps.py" />
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -514,9 +514,9 @@ ome.ol3.Viewer.prototype.bootstrapOpenLayers = function(postSuccessHook, initHoo
        projection: proj,
        center: imgCenter,
        extent: [0, -dims['height'], dims['width'], 0],
-       resolutions :
-           ome.ol3.utils.Misc.prepareResolutions(zoomLevelScaling),
-       resolution : actualZoom
+       resolutions : possibleResolutions,
+       resolution : actualZoom,
+       maxZoom: possibleResolutions.length-1
     });
 
     // we have a need to keep a list & reference of the controls


will-moore:
I'm seeing some strange behaviour in Chrome (62.0.3202.94) and Firefox (58.0b4) on my Mac.
For non-tiled images, I can't zoom in beyond 45%. Images initially open at 100% but as soon as I try to zoom in or out with the mouse-wheel, they reduce down to 45% and won't go above that. I can zoom higher using the + button or setting the value, e.g. to 100%, but as soon as I zoom with mouse it falls to 45% again.
When I checked out snoopy daily and built locally, I didn't initially see this bug. It was only when I built OL viewer with ```ant``` that I saw the same behaviour as on web-dev-merge.
Strangely @pwalczysko is not seeing this on his machine.
Big (tiled) images are also working fine.

waxenegger:
@will-moore @pwalczysko 
This PR should, in fact, contain code that fixes that very behavior. Are you sure you have no stale versions (cached). Open the dev tools, refresh and make sure "disable cache" is checked.
As for local builds you have to wipe node_modules (especially with library upgrades).

I just tested with chrome, opera, vivaldi, firefox on both linux and win8 in a VM (also IE there) and I don't see that issue any more.

will-moore:
This is what I just tried....
```
rm -rf node_modules/
git fetch snoopy 
git checkout snoopy/master/merge/daily
ant
npm run debug
```
Restarted web, cleared browser cache etc
I see the issue described above.
```
rm -rf node_modules/
git checkout origin/master
ant
npm run debug
git graph
*   2b22cf9 (HEAD, origin/master, origin/HEAD) Merge pull request #145 from hflynn/patch-4
```
Restarted web, cleared browser cache etc
I see the issue described above.

So, it seems the problem is in a currently-opened PR.
NB: We only see this issue when using the track-pad on a Mac to scroll. Using a physical mouse-wheel to scroll doesn't show this problem.

will-moore:
I repeated the above using this branch:
````
* 24f2f9a (HEAD, harald/ol3_upgrade) could as well bump to 4.6.4
```
and I see the issue here too. So it seems to be from this PR.
NB: fixed my previous comment - (I *don't* see this issue in origin/master).

waxenegger:
ok then, I'm going to push a commit and we'll see what that does.

will-moore:
I'm afraid that didn't fix it for me.
I tried checking-out your first commit of this PR da7ebde, built that and found that the problem exists there. The previous commit 4f22c95 is working fine (if that's any help)?

waxenegger:
You are pulling this PR, right, not taking a different branch and cherry-picking individual commits?
The one that I thought might help was https://github.com/ome/omero-iviewer/pull/146/commits/b6b5652a3feb7e9a9d09da935383ac358747d979

The previous commit you mentioned is from a different project. In a sense there is no previous to the state of not working for you. Previous would be before the ol3 upgrade (without this PR), right?

will-moore:
Right, without this PR works OK. Happy to try any fixes you have...

waxenegger:
@will-moore ok I force pushed the new code. it's less of a (unneeded) sledgehammer approach to be honest.

will-moore:
@waxenegger Thanks - that seems to be working fine for me now.

PR #148, opened 2017-12-18
add delete key handler

waxenegger:
see: https://trello.com/c/SaPar49q/19-delete-key-to-delete-rois

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -153,7 +153,7 @@ export default class Context {
       * application wide keyhandlers.
       * see addKeyListener/removeKeyListener
       * entries in the map are of the following format
-      * e.g.: key: 65, value: {func: this.selectAllShapes, args: [true]}
+      * e.g.: key: 'A', value: {func: this.selectAllShapes, args: [true]}
       *
       * @memberof Context
       * @type {Map}
@@ -455,7 +455,8 @@ export default class Context {
         if (window.onkeydown === null)
             window.onkeydown = (event) => {
                 let command = Misc.isApple() ? 'metaKey' : 'ctrlKey';
-                let keyHandlers = this.key_listeners.get(event.keyCode);
+                let keyHandlers =
+                    this.key_listeners.get(event.key.toUpperCase());
                 if (typeof keyHandlers === 'undefined' ||
                     event.target.nodeName.toUpperCase() === 'INPUT') return;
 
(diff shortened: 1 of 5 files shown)


mtbc:
Heh, I'm reminded of https://github.com/openmicroscopy/openmicroscopy/pull/3771#issuecomment-97716437!

jburel:
@will-moore did you have a change to look at it?
I re-list on Monday

will-moore:
Simply using the Delete key on Mac didn't work for me. I had to google to find this solution: http://osxdaily.com/2013/04/26/delete-key-mac-forward-del/
Holding down the ```fn``` key while hitting Delete worked OK.
But it should be possible to make this easier for Mac users?

Otherwise all works fine.

waxenegger:
Can you do me a favor and tell me the keycode if you hit just the DEL key: http://keycode.info/

I'd rather solve it by additional keycode than rewrite the code to support some fn combination for the one mac exception.

will-moore:
Yep, ```8``` 
http://keycode.info/ in nice!

waxenegger:
Thanks. 

I'll push some commit later trying another, supposedly better approach.

will-moore:
I used https://craig.is/killing/mice for OMERO.figure - Does everything I needed, handles various OSs, key combinations etc.

waxenegger:
I have had a cursory glance and it uses `KeyboardEvent.which` with fallback onto `KeyboardEvent.keyCode` whereas I use the latter. From what I could gather which works better with Macs. 

But he "supposed problem" with that library as well as my solution is that both properties are deprecated. The absurdity of it is that the envisioned future solution is KeyboardEvent.code which in turn is not yet supported by all browsers: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code. So this leaves KeyboardEvent.key.

joshmoore:
https://github.com/ome/omero-iviewer/pull/148#issuecomment-361774198

@will-moore : do any other repositories use Mousetrap? Are there any other strategies currently in use in OME repositories?

will-moore:
@waxenegger Working fine for me now, thanks.

@joshmoore We don't use Mousetrap anywhere else. Just use ```event.which``` and ```event.keyCode``` in one or two places in webclient/webgateway.

pwalczysko:
I will test it on Windows laptop

pwalczysko:
Win7 laptop 

   - Firefox
   - IE
   - Chrome

The Delete key works as expected. 

On my Mac with attached Windows keyboard - both Delete and Backspace keys work.

PR #149, opened 2017-12-19
csv export - omit zero point stats

waxenegger:
one of the comments in https://trello.com/c/xSkJ2mju/22-drawing-rois-outside-images was that zero point stats i.e. shapes that are fully outside should not be included.

This PR leaves a row for the shape in question with all columns populated (incl. area and length) except the ones that concern the stats.

PR #150, opened 2018-01-04
csv export - omit zero point stats

jburel:
one of the comments in https://trello.com/c/xSkJ2mju/22-drawing-rois-outside-images was that zero point stats i.e. shapes that are fully outside should not be included.

This PR leaves a row for the shape in question with all columns populated (incl. area and length) except the ones that concern the stats.

Replaces https://github.com/ome/omero-iviewer/pull/149
now that the method has moved
cc @waxenegger

### DIFF
--- a/src/app/header.js
+++ b/src/app/header.js
@@ -111,8 +111,8 @@ export class Header {
      *
      * @memberof Header
      */
-     onImageConfigChange() {
-         this.image_config = this.context.getSelectedImageConfig();
+    onImageConfigChange() {
+        this.image_config = this.context.getSelectedImageConfig();
 
         // clean up old observers
         this.unregisterObservers(true);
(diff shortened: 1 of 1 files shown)


pwalczysko:
I suppose this is not included, as there are failed checks here, and web-dev-merge shows a different behavior than described here (there are the zeroes in the stats of the ROIs outside the image)

jburel:
@pwalczysko sorry I did not realise travis failed

pwalczysko:
This works as described.

I think this is a great improvement of the existing behaviour. But I would actually go further,   

1. removing also the length and area. 
2. dealing with ROIs which are partially in and partially out of the image, as these are not treated now at all. I mean by this that the summaries of intensities which are omitted for the ROIs completely outside of the image are listed/included in a ROI which is partially out of the image. But because the area is definitely the area of the whole ROI, I am really not sure what happened with the intensity values outside the image. I suppose they were omitted too and this is how it should be, but as a user which does not know this discussion, I would not be able to come to such conclusion, being confused by the area parameter.


Happy for this to be merged as an improvement and work on points 1. and 2. above in subsequent PRs.

jburel:
@pwalczysko I will deal with 1 in this PR and 2. should be done in a second PR

jburel:
@pwalczysko fixed point 1
Deployed and ready for review

pwalczysko:
The [point 1.](https://github.com/ome/omero-iviewer/pull/150#issuecomment-355541636) works as epected now as well. Ready to merge.

jburel:
Second point is captured on the card mentioned in the description.
Merging

PR #153, opened 2018-01-06
export channel name

jburel:
Use channel label instead of index

to test:
 * open an image
 * Select rois
 * export
 * check that the channel's label is displayed instead of the index

See https://trello.com/c/86NGl6mO/26-iviewer-export-cvs-channels-name

### DIFF
--- a/src/app/header.js
+++ b/src/app/header.js
@@ -309,7 +309,7 @@ export class Header {
                                 csv += csvCommonInfo + "," + emptyRow;
                                 break;
                             }
-                            let value = channels[stat.index].label;
+                            let value = format_value(channels[stat.index].label);
                             csv += csvCommonInfo + value + csvMeasure +
                                     stat.points + "," + stat.min + "," +
                                     stat.max + "," + stat.sum + "," +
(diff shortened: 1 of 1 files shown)


mtbc:
This works okay even if there is a comma in the channel name?

jburel:
I have not tried it. I do not think that will be the "real" case unless the user decides to edit the channel name and introduce comma which does not make much sense.

mtbc:
Okay. Wasn't sure if there's some generic utility method already handy for escaping strings for CSVs.

jburel:
@mtbc I will have a look

jburel:
We could replace "comma" by "dot" since we could have value`` 617,0`` 
this could also be the case for the stats (this was requested for insight)

pwalczysko:
I am afraid, the Export from the iviewer ROIs is not working on web-dev-merge today at all. It just does not produce any excel sheet to download, instead, after some time, the image gets refreshed and the Settings tab gets selected. 

Will probably need some investigation.

jburel:
@pwalczysko sorry about that it should be fixed now. I will trigger a build when travis is gree

will-moore:
This is working fine for me. Good to merge.

PR #154, opened 2018-01-10
travis

jburel:
Attempt to fix travis failure


jburel:
Merging so we can review the other PR
This should be reverted when possible

PR #156, opened 2018-01-28
update selected shapes array once new shape has been saved (id sync)

waxenegger:
Newly created shapes will receive a proper id from the backend after persistence. While these shape ids get synced/updated accordingly they get overlooked in the selected_shapes array (noticed in https://trello.com/c/Hi0GAFjH/80-bug-iviewer-roimeasurements-roi-selection-and-export). This PR fixes that.

TEST:
1. Draw 1...n new shapes.
2. End drawing.
3. Select 1..ln of these new shapes
4. Hit "Delete"
Expected Result: They should all appear deleted (gone and marked red in list)


jburel:
Tested both workflows described in https://trello.com/c/Hi0GAFjH/80-bug-iviewer-roimeasurements-roi-selection-and-export
It now works
Thanks @waxenegger

PR #157, opened 2018-01-30
Use image name for tab title

waxenegger:
The image name is now displayed in the browser's tab.

see: https://trello.com/c/E49kKiWR/16-rfe-image-name-in-tab-title

Test: Check that the tab title shows the image name for both multi and single image mode. In multi image mode the tab title should reflect the active/selected image.


jburel:
The short name is displayed, the full name should be displayed in the tool tip

Looking at the UI of the info tab we will later need to improve the display
this is outside the scope of this PR

PR #158, opened 2018-01-30
Use full support URL for PyPI

hflynn:
As per https://github.com/ome/omero-figure/pull/267 this updates the README to use the full URL for the support file so that the link works from PyPI


jburel:
Thanks
Looks good. 
Merging

PR #160, opened 2018-02-01
csv export - unattached shapes

waxenegger:
see: https://trello.com/c/yaN31Wzy/22-do-not-default-to-a-plane

The following changes have been made in this PR:
- Only shapes that are attached to specific t/z planes will be sent to be queried.
- These unattached shapes will be exported with their stats columns empty
- Commas in channel label or file name will be quoted (in the csv spirit)

Note: We could go all quotes, even for column values that are numeric or don't contain spaces, separators or quotes. I have one older software that insists on that but, strictly speaking, quoting values is optional except for when there are "special" characters such as the ones mentioned.

@jburel I'm going to add an option to the menu for UTF export, such that both excel and "less proprietary" csv are available. If that's too much, I can rollback that commit...

TEST: 
- Export selected shapes where at least one has an unattached z/t (or both). Check that stats cells are empty.
- Try out the non excel option for export and open with a non-microsoft office (e.g. libre/open office), google docs, or an open online csv viewer, e.g. http://www.convertcsv.com/csv-viewer-editor.htm, http://www.becsv.com/csv-viewer.php

### DIFF
--- a/src/app/header.html
+++ b/src/app/header.html
@@ -81,8 +81,14 @@
                     <li class="${image_config.regions_info.ready &&
                                  image_config.regions_info.selected_shapes.length > 0 ?
                                     '' : 'disabled-color'}">
-                        <a click.delegate="saveRoiMeasurements()"
-                           href="#">Export</a>
+                        <a click.delegate="saveRoiMeasurements(false)"
+                           href="#">Export (Excel)</a>
+                    </li>
+                    <li class="${image_config.regions_info.ready &&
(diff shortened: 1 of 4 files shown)


will-moore:
Shapes without Z/T don't have stats included in csv. 👍 
Export of CSV format with commas etc opened fine in Google docs.

Good to merge.

jburel:
We will probably need to review the name and display of the various menu options before release.
Not in this PR. 
merging

PR #161, opened 2018-02-02
Roi Table - multiple select

waxenegger:
see: https://trello.com/c/U2PLznR0/21-multi-selection-of-rois-in-table

TEST: select multiple entries in the roi listing (in the right hand panel) using shift for range selection and ctrl/command for adding/removing single entries.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -777,6 +777,10 @@ thumbnail-slider img {
 }
 .regions-table-row * {
     height: 100%;
+    user-select: none;
+    -ms-user-select: none;
+    -moz-user-select: none;
+    -webkit-user-select: none;
 }
 
 .regions-table-col {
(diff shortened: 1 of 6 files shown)


waxenegger:
--exclude

pwalczysko:
Works as expected on 

Win7
    
   - Firefox
   - Chrome
   - IE11

Mac

   - Firefox

I am sure @will-moore will check Chrome on Mac.

will-moore:
Works great for me.
One minor suggestion - In the case when I have multiple shapes selected, then I de-select some until I only have a single shape selected, it would be nice if theZ, theT (and theC) updated to show that shape (I notice the colour choosers are already doing this). 
Changing theZ and theT is probably not needed when multiple shapes are selected. I think the current behaviour is OK (although mimicking the behaviour of the colour choosers to show the most recently clicked shape is an option)?

will-moore:
This is working really nicely now, thanks. Good to merge.

PR #162, opened 2018-02-02
Open with

will-moore:
See https://trello.com/c/o2mb45PD/17-iviewer-open-with-other-apps

To test:

 - open-with options should appear as links in Info Panel.
 - Check that they are rendered correctly and that the links open the correct URLs in a new Tab.

Open_with scripts expect global methods ```OME.setOpenWithEnabledHandler``` and ```OME.setOpenWithUrlProvider```

cc @waxenegger @jburel 

screen shot 2018-02-02 at 22 23 16

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -226,6 +226,9 @@ export default class Context {
         // set up luts
         this.setUpLuts();
 
+        // set up Open_with
+        this.setUpOpenWith();
+
         // open what we received as inital parameter
         this.openWithInitialParams();
 
@@ -295,6 +298,33 @@ export default class Context {
                         if (isInList) i++;
                     });
                 for (let [id, conf] of this.image_configs) conf.changed();
+            }
+        });
+    }
+
+    /**
+     * Sets up 'Open With' by via AJAX call then loading scripts
+     *
+     * @memberof Context
+     */
+    setUpOpenWith() {
+
+        // now query the luts list
+        let server = this.server;
+        let uri_prefix =  this.getPrefixedURI(WEBGATEWAY);
+        $.ajax(
+            {url : server + uri_prefix + "/open_with/",
+            success : (response) => {
+                if (typeof response !== 'object' || response === null ||
+                    !Misc.isArray(response.open_with_options)) return;
+
+                window.OME.OPEN_WITH = response.open_with_options;
+                // Try to load scripts if specified:
+                window.OME.OPEN_WITH.forEach(ow => {
+                    if (ow.script_url) {
+                        $.getScript(ow.script_url);
+                    }
+                });
             }
         });
     }
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -18,6 +18,7 @@
 import {noView} from 'aurelia-framework';
 import {EventAggregator} from 'aurelia-event-aggregator';
 import Misc from '../utils/misc';
+import OpenWith from '../utils/openwith';
 import ImageConfig from '../model/image_config';
 import ImageInfo from '../model/image_info';
 import RegionsInfo from '../model/regions_info';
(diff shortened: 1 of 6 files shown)


src/app/context.js
```diff
@@ -299,6 +302,33 @@ export default class Context {
         });
     }
 
+    /**
```
waxenegger: @will-moore I know there is some setup code in context.js, the big difference to the open with is that this code needs an absolute guarantee to run before anything else, specifically the aurelia initialization which I have pretty much no control over in terms of order.

I guess what I'm saying is that I'd rather have the bulk of the code out of context and closer to where its happening. I try to encapsulate the functionality right in the files where it's needed. If there is a needed sequence of actions I think it should be achievable via callbacks in your case.

As a matter of fact, there is enough code and most of it being self contained/global in nature that I'd put it into its own file and then just call with the few parameters needed in the respective places.

I think the utils directory might be a good place and you could do it as in https://github.com/ome/omero-iviewer/blob/master/src/utils/misc.js with "static" methods so that you just need to call something like ``` OpenWith.init ```

waxenegger: forgot to mention: the noView decorator would be needed for that additional file (like Misc class), anything else should be more or less analogous (export and import statements).

will-moore: It's kinda analogous to the ```setUpLuts()``` in that we need to load the open-with options ASAP. In fact, we also need to load the ```openwith.js``` file for each open-with and these need to be loaded before the right-panel renders.
When testing with local server (single request processed at a time) I sometimes found that the right panel Info tab would render before the ```openwith.js``` loaded e.g. for webgateway viewer. This would then give an incorrect URL to that viewer.
I didn't see this problem when using nginx - since the simultaneous processing of request meant that the ```openwith.js``` was generally returned faster - before the ```imgData``` json and before the Info tab rendered. But this is still not guaranteed.

I'll move as much of the code as I can to it's own file.

waxenegger: Yes. Moving as much as you can is a good idea.

On an aside, setUpLuts is instantly needed because the settings options display the png instantly, the actual height depending on the number of luts is needed and image data uses the luts table for lookup table internally. Settings are also the default to show. T

Open with would still render correctly (and is waiting on image_info ready already) if the script pulling wasn't done right away, right? Another idea is a loaded counter in a global/context variable and the open with links being enabled once everything is there. 

As a matter of fact, perhaps what I said in the above paragraph is not even necessary because the open_with_links array that you use should synchronize between ui and collection that is to say once you add it to the array, the ui will update. So before you add an entry, the link will not be there even.

waxenegger:
sorry, I was perhaps a bit vague. maybe it's easier to convey the idea via code. here's a possible alternative approach, combining open with code and fetching scripts a bit later (at least after image data requests...which does feel a bit faster). There is more potential to optimize/delay the latter if needs be.

https://github.com/waxenegger/omero-iviewer/commit/a30ddd45bb1d32dd515c35ea9fa2a08ebe5571c3

will-moore:
@waxenegger Thanks for tidying the code. I removed some of the setting of global variables, including ```window.OME = {};``` during the ```fetchOpenWithScripts()``` which was wiping out the ```window.OME.setOpenWithUrlProvider``` method.
These methods are the only variables that need to be global. The actual ```OPEN_WITH``` array is not global now, which is a bit tidier.

waxenegger:
yes, that was an oversight. forgot to delete these 2 variable definitions after I created initOpenWith.

src/utils/openwith.js
```diff
@@ -102,8 +100,7 @@ export default class OpenWith {
     static getOpenWithLinkParams(image_id, image_name, iviewer_url) {
         let ret = [];
```
waxenegger: The ret variable can go now since you are returning the result of map and filter instead of the iteration I had

waxenegger:
Code-wise everything looks fine. Testing the open with links works as intended.
Not that this is relevant to this PR at all but strangely enough I happened to choose a big images for player and the 3D viewer which leads to me waiting for Godot.

jburel:
Could you exclude the "old" viewer?

will-moore:
@jburel Excluded webgateway viewer from open-with.

jburel:
Thanks
Merging

PR #163, opened 2018-02-06
set tooltip/opacity for restricted permissions

waxenegger:
see: https://trello.com/c/4Z3O6Z6F/18-show-roi-shape-ownership

Setting the opacity for the row is just an idea to indicate that the user does not have full permissions.
Not sure if the tooltip text should be changed or if 'restricted permissions' is good enough.

### DIFF
--- a/src/regions/regions-list.html
+++ b/src/regions/regions-list.html
@@ -101,12 +101,23 @@
                     </div>
                     <div if.bind="!(shape.deleted && shape.is_new) &&
                                    !(roi.shapes.size > 1 && !roi.show)"
-                        title="ROI: ${roi_id} - Shape: ${shape_id}"
+                        title="${'ROI: ' + roi_id + ' - Shape: ' + shape_id +
+                                 (!shape.permissions ||
+                                    (shape.permissions &&
+                                     shape.permissions.canAnnotate &&
+                                     shape.permissions.canEdit &&
+                                     shape.permissions.canDelete) ? '' :
+                                        '\n(restricted permissions)')}"
                         id="${'roi-' + shape.shape_id}"
                         class="regions-table-row"
                         css="${shape.selected ? 'background-color: #d0d0ff;' : ''}
                              ${shape.deleted ? 'color: red;' :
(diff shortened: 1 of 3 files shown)


pwalczysko:
I think this feature works as expected. For the tooltip though, follow what there is in other places, basically ``Owner: $NAME`` rather than ``restricted permissions``

screen shot 2018-02-06 at 13 12 02

pwalczysko:
Thank you. Works as expected. Ready to merge FMPOV.

src/regions/regions-list.html
```diff
@@ -101,12 +101,23 @@
                     </div>
                     <div if.bind="!(shape.deleted && shape.is_new) &&
                                    !(roi.shapes.size > 1 && !roi.show)"
-                        title="ROI: ${roi_id} - Shape: ${shape_id}"
+                        title="${'ROI: ' + roi_id + ' - Shape: ' + shape_id +
+                                 (!shape.permissions ||
+                                    (shape.permissions &&
+                                     shape.permissions.canAnnotate &&
```
jburel: to simplify code maybe introduce a "can_do(shape)" method taking a shape as an argument

waxenegger: I know it doesn't look nice but in templates I usually refrain from using methods. For one I'd have to define them in the corresponding js to be able to use them and then I cannot reuse them outside but the actual reason I avoid them is that js object properties are smartly reevaluated on change and updated by aurelia by means of using an observer underneath (when it parses the template expressions). Aurelia cannot decorate function so change can be missed. In this particular example there wouldn't be a side-effect I suspect because the permission wouldn't change but if it did the corresponding UI change would not happen because the function cannot be watched for change, only properties can be "decorated" with an observer that notes the change. do you want me to make a local function for the check?

jburel: a local function will work to avoid duplication..

jburel:
Moving the code to the template did not impact the feature.
It works as expected
We should probably  always indicate the owner to the inline with the web client
This is an RFE
Merging this PR

PR #164, opened 2018-02-06
enable scroll for settings tab (if needed)

waxenegger:
see: https://trello.com/c/DDHoaxvr/25-test-with-large-number-of-channels

TEST: user 6, http://web-dev-merge.openmicroscopy.org/iviewer/?images=95503&dataset=22101


will-moore:
Works fine. Good to merge.

PR #165, opened 2018-02-07
Shape linked to specific Channel - Rendering

waxenegger:
see: https://trello.com/c/P8ikdorl/26-iviewer-channel-specific-display-of-rois

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1822,6 +1822,9 @@ ome.ol3.Viewer.prototype.changeChannelRange = function(ranges) {
 
     omeroImage.changeChannelRange(ranges);
     this.affectImageRender();
+
+    // update regions (if necessary)
+    if (this.getRegionsLayer()) this.getRegions().changed();
 }
 
 /**


will-moore:
This is working fine.
However, I think we need to show theC in the shapes table (even if we leave the editing of theC to another PR) because without it you have no way of knowing why your shapes have disappeared or are not showing up, which is confusing.

will-moore:
@waxenegger Having played with this a bit, it seems to me that we should apply the same logic to shapes with theC as we do with theT and theZ - i.e. if you select a shape with theC not null, we should show that shape on the viewer AND turn the channel ON if it isn't on. This is what we do for theZ and theT (we go to the appropriate Z/T plane in order to show the shape there). So, with theC we should turn the channel ON if it's not on, instead of not showing the shape at-all which is kinda confusing.
Hope it's not too much work to make the ROI panel talk to the settings panel?

will-moore:
NB: for future testing, I set theC on the larger ellipse in this image: http://web-dev-merge.openmicroscopy.org/webclient/img_detail/71351/?dataset=17803 (user-3)

waxenegger:
@will-moore I have made the change that will turn on the channel in https://github.com/ome/omero-iviewer/pull/161 where I had to make related changes. If I hadn't there would be yet another conflicting PR.

We can either leave this one open, or I can integrate the one commit/change into the other PR and then close this one altogether.

waxenegger:
@jburel sorry I forgot to copy you in on this one initially since I did the requested change in the other PR (comment above) which is now merged. It's not a problem really, this one can still be dealt with in a "isolated" way in terms of review and merging but that one commit is still needed for the channel toggle to reflect in the UI right away.

will-moore:
Turning on the Channel when shape with theC is selected is working fine.
Still would be nice to see theC column in the shapes table.

jburel:
Let's add theC in a follow PR so we can unblock other PR
Card created https://trello.com/c/scTDGiHY/30-display-thec-in-roi-table

PR #168, opened 2018-02-12
delete tmp data after initRegions

waxenegger:
The data (json) from the regions request is better deleted after the initialization of regions in the "ready" observer has fired. 

Testing with IE I happened to notice that the regions layer was not added in multi viewer mode when a new image was opened (very likely due to a different execution order for async handling and/or memory deletion policy).

TEST (especially in IE10+ since this does problem does not manifest itself in chrome and FF)
Open an image, switch to multi viewer mode, view ROIS (clicking tab), then open more images (with rois). They should display properly.

### DIFF
--- a/src/viewers/ol3-viewer.js
+++ b/src/viewers/ol3-viewer.js
@@ -251,7 +251,10 @@ export default class Ol3Viewer extends EventSubscriber {
                         (newValue, oldValue) => {
                             if (this.viewer === null) return;
                             this.viewer.removeRegions();
-                            if (newValue) this.initRegions();
+                            if (newValue) {
+                                this.initRegions();
+                                delete this.image_config.regions_info.tmp_data;
+                            }
                     });
             // mdi needs to switch image config when using controls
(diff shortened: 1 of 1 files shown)


waxenegger:
@jburel I'm a bit puzzled by the error from travis. does the new test framework not run `ant unit-tests-only` in the checked out git project root. if so it should have all rights to do its work in that directory and any of its sub directories, right?

jburel:
https://github.com/ome/omero-iviewer/pull/170 fixing travis has been merged
you may want to merge develop to have travis green again

dominikl:
👍  Tested with IE 11, seems to work alright, couldn't see any difference between IE and Chrome. ROI layer displayed correctly in all multi views. I just noticed that the multi view viewer is slightly too small to show the whole image. Don't know if that's an issue which is already noted somewhere. Certainly not related to this PR.

Single view:
screen shot 2018-02-19 at 10 21 39

Multi view(s):
screen shot 2018-02-19 at 10 31 36

PR #169, opened 2018-02-13
Rois Persistence Revisited

waxenegger:
see: https://trello.com/c/Kn7HBtKx/13-bug-100-rois-deletion-fails
and somewhat related: https://trello.com/c/GiDmiT8N/4-permission-testing-new-role

The new approach breaks persistence into the following sections:
- new and deleted (eliminated on the js side and not being sent - mentioned for completeness sake)
- new
- modified
- empty rois, i.e. deleted shapes resulting in an empty roi which has to also be deleted
- deleted shapes

The size of json sent for the ajax request should be smaller now and the amount of marshalling work reduced. For deletion only the ids are sent. No objects are marshalled for return but instead only ids to sync in the front-end and receive the new ids for new objects.

The preferred method of deletion is `deleteObjects` on the roi level for empty rois.

The only exception is for the deletion of individual shapes in a rois with multiple shapes which would have to be created outside of iviewer, e.g. script or insight. With those the `removeShape() `method is used such as to not create index holes in the collection. I have opted for this solution for now since no server solution is in sight (see 2. trello discussion). Another approach is to not mind about creating index holes (using `deleteObject` as well) but that necessitates changes in omero-marshal which as is gets tripped by the None entries that are created in the shapes collection when it tries to marshal them. A "solution" I don't even want to put out there is the move back to marshalling in the iviewer view which was abandoned in favor of the web api.

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1639,7 +1639,7 @@ ome.ol3.Viewer.prototype.getSmallestViewExtent = function() {
  * Persists modified/added/deleted shapes
  * see: {@link ome.ol3.source.Regions.storeRegions}
  *
- * @param {Array.<number>} rois_to_be_deleted an array of roi ids for deletion
+ * @param {Object} rois_to_be_deleted an associative array of roi ids for deletion
  * @param {boolean} useSeparateRoiForEachNewShape if false all new shapes are combined within one roi
  * @param {boolean} omit_client_update an optional flag that's handed back to the client
  *                  to indicate that a client side update to the response is not needed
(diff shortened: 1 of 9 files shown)


will-moore:
This seems to be working fine. I haven't tried deleting large numbers to test performance yet.

Just looking at the JSON you're sending... I can understand the format OK, but it's not exactly what I would expect. For example, if we were to move this to the ```webgateway``` to be used by other clients, then the usage of strings to list Shapes under an ROI as ```"ROI_ID:Shape_ID"``` would probably be better as simply ```Shape ID``` (number) since this is already nested under ```ROI ID```.
Also, the ```deleted``` object under ```rois``` refers to deleted Shapes, not ROIs? which is a bit confusing.

screen shot 2018-02-19 at 10 25 53

I don't think this needs to be changed for this PR, since we can consider is "internal". But, as with ```webclient``` "internal" JSON methods, these have a habit of being needed and used elsewhere, so it's nice to get it right for general usage.

waxenegger:
It is very much internal. It's a practical approach and a response json not intended to be consumed for general api usage but instead tailored to what iviewer relies upon in order to work, optimized for lookup using its internal, composite keys, syncing rois states and avoiding loops for conversions and additional requests.

That said, most bits of the saving route (the python bits) are straight-forward and the essence of it can be easily extracted and basically reduces to saveArray and deleteObjects/removeShape (with a marshal if needed - add, modify).

In the zoom session we had it was said that path viewer does not sync the way iviewer does. Likewise any other client or api might want to make different decisions: could be you'd rather use saveAndReturnIds for new ones instead, could be you'd want to provide independent methods for the categories of persistence, just to name 2 examples.

will-moore:
Just tried creating 200 single-shape ROIs, then updating 200 ROIs and deleting 200 ROIs in the same operation and it only took ~2 secs, so performance looks pretty good 👍

PR #170, opened 2018-02-13
permission issue build

jburel:
Remove node_modules folder after run prod in setup.py
This should have no impact when running locally ``npm run prod``
I have created a card to review later on the build system
https://trello.com/b/clkD8YDR/iviewer-priority-list

Possibility: this could become the last step for ``prod``, ``debug`` etc. This means that each local build will take a bit longer
Check that travis is green

### DIFF
--- a/plugin/setup.py
+++ b/plugin/setup.py
@@ -50,6 +50,7 @@ class RunProd(Command):
     def run(self):
         if not os.path.isdir('omero_iviewer/static'):
             self.spawn(['npm', 'run', 'prod'])
+            self.spawn(['rm', '-rf', '../node_modules'])
 
 
 cmdclass['run_prod'] = RunProd


waxenegger:
@jburel it's great that travis is green but I would absolutely dislike the wiping of node_modules to happen as part of the regular build as you suggested. For me the refetching of all dependencies makes up a really large wait in development.

The CI tests for PRs for wev apps has also increased substantially with the test infra in place but a local slowdown for development would be a true disaster.

jburel:
The way it is now in place it will not affect a local build when you run any command like
``npm run dev/debug/prod``
it will only delete ``node_modules`` if you run ``python setup.py sdist``

waxenegger:
Thank you. Feel free to merge this whenever you see fit. I wouldn't mind my PRs to go green again :wink:

PR #166, opened 2018-02-08
opening multiple images with double click

waxenegger:
see also: https://trello.com/c/BFCxWDcW/24-to-discuss-multiview-design-issue-81

Double clicking on the thumbnail will open up a new image in multi view mode. From then on one can continue double clicking to add new image windows, or single click which will replace the presently active/selective window/image with the chosen one. After closing all but one image iviewer reverts to single view mode.

PR #172, opened 2018-02-19
draggable viewer limits

will-moore:
Small fix for a minor bug I noticed:
Currently it's possible to drag a viewer (in multi-viewer mode) so that the header
disappears off the top of the page, so that it's not possible to drag it again or to close it.

This fixes that bug so that it's not possible to drag the viewer above the top of the draggable area.
It also puts a limit on how far left you can drag a viewer (although it does allow the left of the viewer to go outside the drag area.


waxenegger:
The code makes sense and works. 

Unfortunately I came across another problem which is I suspect related to the other UI glitches that we talked about:

image

I just pushed past the right hand panel, likewise one can push further down than should be. 

I tried https://github.com/ome/omero-iviewer/pull/144 (quite old - no multiple windows) and https://github.com/ome/omero-iviewer/pull/169 (rather recent) locally and neither one has the css changes. Neither have the collapse/expand problem and the latter doesn't have the pushing past right hand panel in multi viewer mode.

will-moore:
https://github.com/ome/omero-iviewer/pull/174 should fix side panel resizing.

jburel:
difficult to say if #174 will fix the issue (not included in build)
but to today it is like
(https://user-images.githubusercontent.com/1022396/36469637-cc5b56a6-16df-11e8-8c56-3fa03e7f4f89.png)

waxenegger:
I'm not at all happy with the fact that we are making up for something that should not be possible to begin with. Space should not just be extendable beyond the limits of the screen, especially not the right. This is not a mobile setup where things are in flux but a fixed setup as far as basic layout goes. We are addressing the symptoms instead of the cause.

In fact I have had another issue crop up occasionally which is indicative of the canvas element having expanded ever so slightly after page load at a point when it should not because nothing makes up for that (an updateSize in openlayers is needed). One cannot see it as such but it had the effect that the rois drawing was off. Imagine the mouse pointer being to the left or right of where the actual line is drawn. I know that effect I can say that with absolute certainty that it happens after the size of the canvas has changed. If people are prepared to put up with these random glitches in the future so be it...

waxenegger:
the start method throws an error when one tries to drag the window:
image

will-moore:
I never saw that error on this branch alone. Maybe things are behaving a bit differently with the double-click multi-viewer mode. I'll try it with more branches merged in...

will-moore:
I can look at reverting and removing the ```flex``` layout. The reason I used it originally was because it appeared to be a more robust way of handling the left and right panel resizing, since you didn't have to do so much calculating of frame margin and I couldn't get that code to work once I'd removed the gap between the centre and side panels.
The fact that it had other side effects has only come up once that was merged. I was going to look at reverting or using alternative layouts yesterday but got side tracked by the Polyline display issues.

Agreed, it should not be possible for the container to expand automatically, but it should also not be possible to drag a viewer out of the frame, which is what this PR is addressing and has nothing to do with the flex issues.

waxenegger:
@will-moore I have a feeling, the only difference that you need to try is to build with `npm run prod` and you'll see it. 

I haven't tried this out but I believe that the arrow function takes the context from the object that you have used to hand over to the draggable function. I believe they are very strict in the standard to enforce that and so are the transpilers. For that reason it will not capture the let declared variable 'frame' but instead interpret it as a new declaration in a new scope/context.

Try to define the arrow function outside the object and/or bind it explicitly. As I said I haven't tried this out and I could be wrong but it's worth a try. 

I doubt that the merging in causes the problem...

will-moore:
@waxenegger It turns out that it's the changes in https://github.com/ome/omero-iviewer/pull/166/files#diff-b445f40c6859f02edacb925f6ab66d49R182 that limit the scope of ```frame``` that cause this bug when merged with this PR. I've reversed my last commit so we can get a usable build. Can try to bring it back once #166 is merged and I can fix conflict.

waxenegger:
I don't understand how these 2 cannot conflict on merge: https://ci.openmicroscopy.org/job/iviewer-merge/673/console?
I have had snoopy report conflicts when 2 different lines at different ends of the css file were modified that could have been merged automatically, and here it merges where massive changes have been made to the same code region...

will-moore:
It seems the easy way of stopping the frame extending outside the ```.frame``` at-all is to use
```
container.draggable({
            containment: $('.frame'),
            ...
```
but this is quite restrictive (not allowed outside at-all) which may not be what we want, since there might be times you want a part of a viewer to be outside the frame.

waxenegger:
I agree, it's probably too restrictive. I'm wondering if we could allow to go a bit higher up even. On my laptop at least I have quite a bit of freedom up top.

src/viewers/ol3-viewer.js
```diff
@@ -195,6 +195,21 @@ export default class Ol3Viewer extends EventSubscriber {
...
         container.draggable({
             handle: '.viewer-handle',
+            start: () => {
+                minX = frame.position().left + 100 - parseInt(this.image_config.size.width, 10);
+                minY = 0;
+                // Width is smaller when thumbs expanded. Fix by adding pos.left
+                maxX = frame.width() + frame.position().left - 100;
+                maxY = frame.height() - 100;
+                console.log(frame.position().left, this.image_config.size.width, 'minX', minX, 'maxX', maxX);
+            },
```
waxenegger: log relic

waxenegger:
Working well now. Only that one console.log leftover.

jburel:
The feature works nicely now.
I will merge when the ``console.log`` is removed

will-moore:
Sorry - had removed it locally but not committed. Removed now.

will-moore:
NB: @waxenegger As you saw with the error above, even if this merges cleanly with your branch, a refactor is needed to expose the ```frame``` and potentially other variables since your PR has their scope restricted in an ```if``` block, if I remember correctly.

jburel:
travis is green merging

PR #173, opened 2018-02-20
add channel column

waxenegger:
see https://trello.com/c/scTDGiHY/30-display-thec-in-roi-table


jburel:
Column added merging so we can move forward

PR #175, opened 2018-02-23
Rois Json - Point String Whitespace

waxenegger:
see: https://trello.com/c/tBWyuJV5/31-roi-iviewer-not-drawn


pwalczysko:
@harald116 : In today's build, probably thanks to your ome/omero-iviewer: Pull Request 175 ? the whitespaces are handled and this has the pleasant effect that the tracks are now fully visible and selectable for most of the ROIs in e.g. image 88101
There are still some ROIs which behave wrongly on that image, e.g. the ROI 890358 (all shapes in that ROI are not selectable and not visible on the image, including the shape 921140)
But, as you actually wrote below, and @will_moore confirmed studying the json api 
there are ``NaN``s detectable in this shape (see screenshot below)

Thus it seems that all is good here, as we acknowledged that you cannot deal with NaN s and the ball is on @michaelporter10 's court 

screen shot 2018-02-23 at 10 50 34
screen shot 2018-02-23 at 13 16 18

PR #176, opened 2018-02-26
minimal support for masks (not editable)

waxenegger:
Masks were only investigated at the very early stage of iviewer, in fact the implementation happened in the ol3 viewer only and was based on a separate layer approach. Consequently, as is masks are not working.

This PR introduces minimal support for masks. They can be deleted but cannot be edited/modified otherwise.

Essentially the old implementation is thrown out and replaced by an ol3 style approach which makes it possible to treat a mask as a "shape" along with the workflow/logic that is in place for other types of rois, and to eventually modify and persist it as part of the same workflow.

Test Image: http://web-dev-merge.openmicroscopy.org/webclient/img_detail/45726/?x=300&y=600&zm=150 (user-1)


waxenegger:
--exclude

jburel:
* Display of masks works correctly.
* I can select masks in table or on image but I cannot move them (expected)
* I can drop a shape on top of the mask, but I cannot select it by clicking on it. Selection only works from table
* I am not able to move a shape drawn on top of a mask.

waxenegger:
The masks are not editable for now. They can be deleted. Editing is a challenge for both the interactions since masks are not geometries for openlayers, that is especially true for the modification interaction (where you can move vertices). The rectangle is an visual trick and a second style rather than the original shape/mask. Translations I probably could get to work (ol3 wise) but the persistence I'm afraid will require me to load the original mask object before I can updateAndSave or else I loose things. So they are more expensive and cannot be treated 1:1 as the other modified shapes.

The problem with shapes being obscured by a mask should not occur any more. I thought I had addressed that in my latest commit as well.

waxenegger:
Ah I see what you mean with the shape selection. Well, that's general behavior with any shape. You cannot select a shape that's overlapping any other shape (regardless of masks) unless it's fully covered by that other shape. There is no way of knowing for an area that is a set of part shapes overlapping which one you meant to select. You have to select the wanted shape where it doesn't overlap.

waxenegger:
Although, on second thought, as is, with the mask not being editable I could give preference to other non mask shapes. In general, however, the behavior is that way, and has always been as I cannot determine intentions in an area of overlap.

waxenegger:
Note: the last commit is only of a technical nature. basically the selection logic is not called again if in fact  shape(s) have been selected which was the case before and meant already selected shapes couldn't be "grabbed" everywhere.

will-moore:
The selection of masks seems to be working quite nicely. If another shape is completely within the mask then it is selectable over the mask, but if it is only partially overlapping then the mask is selected over the shape.
Good to merge from that point of view.

NB: I just noticed a minor issue that I remembered seeing at the Cambridge workshop: On the image above, the Polyline is shown in the Table with the same icon as a Line, which is not expected since these 2 shapes have different icons in the toolbar. It makes it hard to tell which shapes in the table are Polylines.
The same could also be said for Arrows: It would be nice to be able to tell the difference between Lines and Arrows in the table, although this is less of an issue (and probably harder to fix since they are both Line types) than the Polyline issue.
The Polyline display is maybe a hangover from Insight where you couldn't draw Polylines directly?

will-moore:
I opened https://github.com/ome/omero-iviewer/pull/177 to fix Polyline icons in table.

PR #177, opened 2018-03-07
Show polyline icon in table

will-moore:
From comment at https://github.com/ome/omero-iviewer/pull/176

This adds a polygon icon to the shapes table, so you can tell Polylines from other Lines.
Also adds Arrow icons (single headed and double headed) to indicate Arrows.

To test, create Arrows and PolyLines and they should look different from lines in the table.
Add or remove arrow heads (none, single, double) from Lines and see that table shows them correctly.

screen shot 2018-03-07 at 13 25 29


will-moore:
Ah, I found another work-around (see above)

will-moore:
OK, maybe your approach is a bit cleaner. I'll improve mine to also handle double-arrows...

will-moore:
Hmm - maybe I'll leave it as it is, unless you have any objection? Or, to distinguish between single and double-headed arrows I could do
```
${shape.type === 'line' &&
                                (shape.MarkerStart === 'Arrow' ||
                                 shape.MarkerEnd === 'Arrow') ?
                                 ${shape.MarkerStart.toLowerCase()}${shape.MarkerEnd.toLowerCase() : shape.type}-icon">
```
similar to my commit above, but this gets a bit unreadable?

waxenegger:
I know I had the toLowerCase in there but it's not needed. After the data was requested I reduce the `@type` to just the type (with a toLowerCase). It's never touched again. We can omit it.

will-moore:
The ```type``` is lowercase but the ```MarkerStart``` and ```MarkerEnd``` aren't.
I guess we don't need to use lowercase ```arrow``` for class names, but that's just what I have now...

waxenegger:
yes, that concerns just the type. 

as for the approach, both are fine, don't think browsers treat the order of the classes differently, right?

PR #179, opened 2018-03-08
light admin - controls disabled

waxenegger:
see https://trello.com/c/GiDmiT8N/4-permission-testing-new-role

The permission check for shape deletion/edit/save included image.can_annotate as well.

### DIFF
--- a/src/model/regions_info.js
+++ b/src/model/regions_info.js
@@ -539,8 +539,7 @@ export default class RegionsInfo  {
                 permission !== 'canEdit' &&
                 permission !== 'canDelete')) return false;
 
-        return this.image_info.can_annotate &&
-                !(typeof shape['permissions'] === 'object' &&
+        return !(typeof shape['permissions'] === 'object' &&
                 shape['permissions'] !== null &&
                 typeof shape['permissions'][permission] === 'boolean' &&
                 !shape['permissions'][permission]);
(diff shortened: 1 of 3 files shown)


mtbc:
Probably worth testing combinations of write data and delete data privileges.

pwalczysko:
In the can-only-delete situation, this PR fixes the old-ish problem of not being able to delete others ROIs on others images. This means that the cell J3-J6 of https://docs.google.com/spreadsheets/d/1GP-Kk0lpo_r7uhsZ-Til_XhOZn2HCnpfnSSiZE7jiVs/edit#gid=1765618607 are now ``passing`` with this PR. This has also the nice beneficial effect of not having the problem highlighted in cells K3-K6 of the same sheet -> because the Delete key now ``should`` remove the shape, it does, and all is good.

pwalczysko:
In the can-both-Delete-and-Write situation. the improvement is here as well:

   - the problem in cell I29 is fixed, because the light admin now can delete the ROI in private group, which is as expected
   - the light admin can also edit others' ROIs in private group, which is as expected

pwalczysko:
All is good here. Ready to merge FMPOV

pwalczysko:
Oh, yes, and as @mtbc pointed out, I added some tests for one who cannot do anything (neither delete nor write). In such sitaution, in all 4 types of groups, you cannot edit, delete, draw and Save ROIs and Rnd setting. So that is as expected too.

Ready to merge.

PR #180, opened 2018-03-10
don't prompt for rois saving twice

waxenegger:
see: https://trello.com/c/3NmcHtvD/35-duplicate-saverois-dialog

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -110,7 +110,7 @@ export class Index  {
                 }
             }
             if (hasBeenModified)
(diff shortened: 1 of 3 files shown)


jburel:
Only one dialog pops up now
The text in the dialog should be ``ROIs and ROI(s)`` instead of ``ROIS and ROI(S)``

jburel:
@waxenegger I think now all instances of ROIS have been fixed

waxenegger:
thank you, forgot about the other cases.

PR #181, opened 2018-03-13
alternative way of marking intensity querying as (in)active

waxenegger:
see: https://trello.com/c/fZcjaS7f/33-crosshairs-button-toggle

maybe another way of doing it, not changing things too much, especially space/icon wise...


jburel:
I think it looks better with the new approach

jburel:
as indicated I prefer the new approach, @pwalczysko @will-moore could you have a look?

will-moore:
Looks fine.
I wanted to make the ol3 buttons look the same as the right-panel buttons for hover and 'selected', and also to centre the crosshairs icon on the button, so I pushed those changes to https://github.com/will-moore/omero-iviewer/commit/ff72468f96877a68e352c3105119f0f5d1e916a7

I seem to remember we originally had the mouse cursor switch to a crosshair when the intensity querying was on (that's why I wanted to use a crosshair icon for the button) but that doesn't seem to be working now? (even before this PR). Would be nice to have that back, unless there was a reason it got changed?

waxenegger:
@will-moore I checked in IE, FF and Chrome and the crosshair cursor is still there.

no changes were made to: https://github.com/ome/omero-iviewer/blob/master/ol3-viewer/src/ome/ol3/controls/IntensityDisplay.js#L351

will-moore:
For me, on web-dev-merge and locally, I always see a regular pointer on the canvas (Chrome and FF), even when I use devtools to set a different css value for cursor.
Maybe it's an OS issue rather than browser issue. @pwalczysko You see crosshair cursor when intensity querying is on and mouseover the image?

waxenegger:
Could be OS related, not sure, I see it on windows and linux. You said you saw the crosshair in the past though, was that on your 'regular' machine/OS?

pwalczysko:
>Maybe it's an OS issue rather than browser issue. @pwalczysko You see crosshair cursor when intensity querying is on and mouseover the image?

Tested on Mac and Windows

    - Mac (Chrome, FF) : no crosshair
    - Win7 (IE11, FF) : no crosshair

I think @will-moore @waxenegger you two might be talking about 2 different things, the crosshair on the image itself (which is not there atm) and the crosshair on the icon in the top right of central pane of iviewer.

waxenegger:
ok this is insane. I tested that on the server and with win8 and IE11 and fedora 26. how can it be there and then dissapear. there is something fishy here. I haven't found an issue on that css setting either.

no I mean the crosshair cursor! I tried making a screenshot but it doesn't capture the darn cursor...

will-moore:
The crosshair cursor issues I was having were solved by quitting Photoshop and Illustrator as described https://stackoverflow.com/questions/18434626/why-is-cursorpointer-effect-in-css-not-working

This is working fine for me now - Good to merge.

PR #182, opened 2018-03-13
csv - export: use all rois as a default (if nothing selected)

waxenegger:
see https://trello.com/c/db3huf65/34-roi-intensity-export-disabled

Note: given the level of uncertainty in the card this PR aims at only addressing the case where nothing was selected.


waxenegger:
closing this one for now after discussion with @jburel (potential work load could be detrimental).

PR #183, opened 2018-03-13
csv export - don't try requesting for big images

waxenegger:
while working on https://github.com/ome/omero-iviewer/pull/182 I noticed that the front end tries to send a request for rois intensities in the case of tiled images. this is not supported by the server and it sends back a json response saying so. consequently, however, no roi listing is exported.

this PR redirects for tiled images so that one gets a csv with front end measures only as is the case with new but unsaved shapes.

TEST: find a big image with rois (or draw new ones and save them), select a bunch, then hit export. you should get a csv that contains them.

### DIFF
--- a/src/app/header.js
+++ b/src/app/header.js
@@ -262,7 +262,7 @@ export class Header {
             regInf.selected_shapes.filter((s) => s.indexOf("-") == -1);
         // if we have no saved shapes or no active channels ...
         // forget about it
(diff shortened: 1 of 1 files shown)


jburel:
* Opened a big image and export few rois: The area and length are exported
* Opened a "small" image and export few rois: area, length, stats are exported

Work as expected

PR #185, opened 2018-03-16
Rois Pasting

waxenegger:
see also: https://trello.com/c/Yr5NmQ0B/30-copy-paste-rois

If shapes are copied between 'compatible' images, i.e. the target image is equal or bigger than the source image their coordinates remain the same. If not they are put in a random location.
NOTE: This does not apply to the context menu (right-click) pasting option "Paste ROI here" which will still use the mouse location to allow the user to place the shape!

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -1412,7 +1412,7 @@ ome.ol3.Viewer.prototype.setRegionsModes = function(modes) {
  * - position => the position to place the shape(s) or null (random placement)
  * - extent => a bounding box for generation, default: the entire viewport
  * - theDims => a list of dims to attach to
- * - scale_factor => a scale factor to apply because of image dimension differences
+ * - is_compatible => the image size of the originating image was smaller or equal
  * - add_history => determines whether we should add the generation to the history
  * - hist_id => the history id to pass through
  */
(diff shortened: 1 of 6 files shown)


pwalczysko:
All ROI types work except:

   - polyline
   - polygon

The polyline and polygon seem to be offset by varying amount when pasted via Ctrl+C Ctrl+V in the ROI table. This seems to be something like 1-2 pixel shift. Interestingly, the shift happens only when you copy and paste the polygon(line) just after drawn. Every subsequent Copy of the first Copy is placed exactly on the same spot as expected.
screen shot 2018-03-16 at 10 22 35
screen shot 2018-03-16 at 10 22 56

will-moore:
```
main.js?_iviewer-0.4.1:80 Uncaught TypeError: Cannot read property 'width' of null
    at new Hh (main.js?_iviewer-0.4.1:80)
    at gn (main.js?_iviewer-0.4.1:80)
    at dn.H (main.js?_iviewer-0.4.1:80)
    at new dn (main.js?_iviewer-0.4.1:80)
    at i.initViewer (main.js?_iviewer-0.4.1:170)
    at e (main.js?_iviewer-0.4.1:170)
    at main.js?_iviewer-0.4.1:170
    at t.m [as callSubscribers] (main.js?_iviewer-0.4.1:96)
    at t.call (main.js?_iviewer-0.4.1:96)
    at t._flushQueue (main.js?_iviewer-0.4.1:9)
    at t.flushMicroTaskQueue (main.js?_iviewer-0.4.1:9)
    at MutationObserver.<anonymous> (main.js?_iviewer-0.4.1:9)
```

I was going to report that everything was working fine for me - but noticed @pwalczysko  comment.
I didn't notice this before - it seems a tiny effect from rounding coordinates DOWN to integers.
```
"Points": "75.07135522888797,15.85939644730555 13.836484724443721,74.2068629394135 111.5225774041125,102.92875392068377 114.30971142657464,47.30078918750597",
```
becomes
```
"Points": "75,15 13,74 111,102 114,47",
```

waxenegger:
@will-moore, @pwalczysko thank you for pointing that out. while the canvas won't give sub pixel placement the parseInt did does incorrectly truncate the float which produces different rounding/aliasing results in the canvas.

will-moore:
Rounding of floats for polygon/polyline on copy/paste is fixed.

PR #186, opened 2018-03-19
don't reset initial version parameter

waxenegger:
see: https://trello.com/c/AdRIcaBJ/37-viewing-large-non-tiled-image-fails

when the image is initially opened in iviewer the omero server version information is handed in by the python layer. since initial url values get wiped after the initial load so was the version information which caused the issue described in the trello card.

also, the iviewer version display was fixed, and in between small and pyramid size images (see trello card) are now backwards compatible to omero server versions below 5.4.4.

TEST: use the images in this dataset:  http://web-dev-merge.openmicroscopy.org/webclient/?show=dataset-1682 (user-3).

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -250,7 +250,7 @@ ome.ol3.Viewer = function(id, options) {
         actual_version = parseInt(actual_version.replace(/[.]/g, ""));
         if (isNaN(actual_version)) return false;
 
-        return version >= actual_version;
+        return actual_version >= version;
     }
 
     /**
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -455,7 +455,10 @@ ome.ol3.Viewer.prototype.bootstrapOpenLayers = function(postSuccessHook, initHoo
        channels.push(newC);
     }
 
-    // create an OmeroImage source (tiled)
+    var isTiled =
+        typeof this.image_info_['tiles'] === 'boolean' &&
+            this.image_info_['tiles'];
+    // create an OmeroImage source
     var source = new ome.ol3.source.Image({
        server : this.getServer(),
        uri : this.getPrefixedURI(ome.ol3.WEBGATEWAY),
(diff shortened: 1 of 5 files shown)


will-moore:
Testing on web-dev-merge with https://github.com/openmicroscopy/openmicroscopy/pull/5692 works fine.
However, testing this locally against develop I still get 404s with ```?tile=0,3,4,512,512``` so I don't think this will work with anything less than 5.4.5 (assuming we get that PR in for 5.4.5).
Otherwise I guess we have to revert to using ```?region=512,512,512,512``` for 'large' images?

jburel:
We need to be compatible with older versions.

waxenegger:
then I have to revert to the state prior to the previous PR and tile sizes won't be able to be set flexibly.

waxenegger:
@will-moore just to be absolutely sure: when you say you tested locally against develop, you don't happen to have a discrepancy in that the omero server branch you used will give already 5.4.4 for a version but https://github.com/openmicroscopy/openmicroscopy/pull/5542 is not in that branch yet. if that were so then iviewer would believe it can make the tile requests when in fact the server side code is not there. the reason I'm asking is that the tile request should not fail with a 404 in the small and pyramid backed image case even without the PR mentioned above (i.e. < 5.4.4). The special case are the bigger images that are not backed by a pyramid.

will-moore:
I currently have
```
>>> omero_version.omero_version
'5.4.2-ice36-SNAPSHOT'
```
This is only failing for large non-tiled images where iviewer is asking for e.g. ```?tile=0,2,2,512,512```.
Big and small images are working OK.

waxenegger:
yes, that's the part that I'm trying to figure out. Testing this PR locally with the server images that I downloaded from that dataset on the server it switches to using ?regions=

jburel:
Could it be the version handling since it is 5.4.x-something?

will-moore:
You need to create the images in OMERO via the Python API not via import. E.g. export OMERO.figure as New OMERO image.
Or run script Channel_Offsets on the image you imported to create a new image via the API.
See https://trello.com/c/AdRIcaBJ/37-viewing-large-non-tiled-image-fails

will-moore:
Those images are still using ```?tile=``` for me but that is working for imported images, not for API-generated images.

waxenegger:
But I got the same errors on import of these images prior to this PR. They have tiles: false and a missing tile_size. How come?

waxenegger:
what's committed at the moment represents a compromise as opposed to rolling everything back: 

If large images are backed by pyramids iviewer uses 512^2 tile size (provided the server version supports it), otherwise it will fall back onto `'&region'`.

will-moore:
This is working fine for me locally (but not "512^2 tile size" as in your comment)...
I'm seeing tiles of various sizes (depending on the image) ```?tile=3,1,0,256,256``` and regions of 512 x 512 e.g. ```?region=1024,1536,512,512```.

waxenegger:
I assume you are in the < 5.4.4 case (unsupported). In that case the default tile size is used since I cannot change it anyhow for pyramid cases. The idea of the previous PR (and this one) is to request bigger tiles. This could not be done prior to the webgateway changes.

There is no obstacle for the regions request to use the bigger tile size, it always supported it. Moreover I'm using it since the image data request does not give me the tile size back in the json since its not tiled.

jburel:
Working fine on devmerge too.
I tried locally without the PR and hit the problem on image from ROIs as expected

PR #187, opened 2018-03-19
About dialog url

will-moore:
Since we now have https://www.openmicroscopy.org/omero/iviewer/ available, this adds the link to the iviewer about dialog.

Also removed some css rules to allow dialog to expand to fit additional text and have larger footer (which also affects other dialogs).

Before:
screen shot 2018-03-19 at 11 28 10

After:
screen shot 2018-03-19 at 11 28 26

About dialog:
screen shot 2018-03-19 at 11 28 40

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -1155,13 +1155,9 @@ thumbnail-slider img {
 }
 .modal-content {
     width: 300px;
-    max-height: 250px;
 }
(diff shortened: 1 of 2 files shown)


hflynn:
👍 for link to the website page

jburel:
Everything looks good.
Thanks

PR #188, opened 2018-03-22
bug: don't count new but instantly deleted

waxenegger:
was part of the remarks on this card: https://trello.com/c/z3AruQMZ/38-feedback-from-mporter

https://github.com/ome/omero-iviewer/pull/72 was later abandoned in favor of general persistence on save with a warning about the number of deleted shapes on save.

this delete count includes newly drawn shapes that got deleted which is confusing and unnecessary since these shapes never existed on the server to begin with.

TEST: draw or paste a new shape(s), delete it/them, then save. There should not be a warning as is in the case when you delete an already existing shape and do the same

### DIFF
--- a/src/model/regions_history.js
+++ b/src/model/regions_history.js
@@ -65,6 +65,13 @@ export default class RegionsHistory {
     historyPointer = -1;
 
     /**
+     * @see checkIfHistoryHasOnlyNewlyDeleted
+     * @memberof History
+     * @type {boolean}
+     */
+    hasOnlyNewlyDeleted = true;
+
+    /**
      * a reference to RegionsInfo
      * @memberof History
      * @type {RegionsInfo}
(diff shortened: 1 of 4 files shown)


jburel:
There is no warning when clicking saved
I draw a shape, delete it then technically I have nothing to save.

waxenegger:
@jburel this last commit should disable the save button for the case when the only changes happened to newly created (drawn/pasted) shapes, without the need to loop over all shapes, just based on the change history.

jburel:
The button becomes greyed out when I click on it and shows the "forbidden" symbol
I was expecting it to be greyed out after clicking the delete button
This could be too many changes for now

waxenegger:
@jburel did you test locally because the changes aren't deployed on the server yet?

jburel:
tested on web-dev
I understood the changes were deployed.
Rebuilding now in that case

waxenegger:
yes, it's a bit misleading, I did a merge of iviewer code via ci but I decided against a web-dev-merge-deploy because it hit me that there was a time change and it was too close to stand-up already.

jburel:
Tested on web-dev-merge after rebuilding.
I have tried various combinations and it works nicely now.
Merging

PR #189, opened 2018-03-22
Bug: drawing mode and context (right) click

waxenegger:
was part of the remarks on this card: https://trello.com/c/z3AruQMZ/38-feedback-from-mporter

While in drawing mode a right click for context triggers a mouse down event as well to the effect that this is regarded the start of a shape drawing. 

What this PR does:
In the openlayers interaction a condition checks that we don't have a right-click for a context menu so that no new shape drawing will be initiated.

What this PR doesn't do:
It doesn't abort the drawing interaction as such, e.g. in the case of a started drawing of a shape a right-click will not abort the drawing, likewise clicking anywhere but the context menu will start a new shape.

TEST: While in drawing mode right-click to bring up context menu. This should not trigger the start of a new shape.


jburel:
Compare web-latest and web-merge, this is now fixed

PR #191, opened 2018-03-22
calculate length for line geometries only

waxenegger:
see: https://trello.com/c/z3AruQMZ/38-feedback-from-mporter

to be consistent with insight and not violate the definition of length for non line type of geometries length is  only calculated for (poly)lines. doing it this way is actually more compact than changing display for both css export and rois table and has the positive side-effect that less calculation has to happen.

TEST: choose an image with rois and/or draw some. Check that the length column in the regions listing in the right hand panel (after selecting "Show Area/Length" option) shows only length in the case of line geometries. Also select rois and then export them to csv/excel via the header menu. Only rows for (poly)lines should show a value in the length column.


jburel:
* length for lines/polylines and arrows, no area
 * area displayed of other shapes, no length
 * text: no area no length

It works as expected

PR #192, opened 2018-03-28
Openlayers Upgrade

waxenegger:
use latest version: 4.6.5


jburel:
We probably need to that one opened for a while

PR #194, opened 2018-04-02
Projection Sync Multi View

waxenegger:
https://trello.com/c/zmaYZqM3/47-projection-multi-view

Note for testing: as with any other settings, the syncing of changes will be turned on only at locking the viewer windows to the same group. Only if the checkboxes for the actions (e.g. z/t, view, channels) are checked will a sync for these actions be performed.
image

### DIFF
--- a/src/app/index.html
+++ b/src/app/index.html
@@ -69,7 +69,7 @@
         <div class="frame col-xs-12">
             <div class="${context.useMDI && context.image_configs.size > 1 ?
                             'viewer-mdi' : 'viewer'}
-                        ${context.useMDI && context.selected_config === id && 
+                        ${context.useMDI && context.selected_config === id &&
                             context.image_configs.size > 1 ? 'mdi-selected' : ''}"
                  repeat.for="[id, conf] of context.image_configs"
                  css="${context.useMDI && context.image_configs.size > 1 ?
@@ -121,7 +121,7 @@
                                              (lock.CHAR === 'zt' &&
                                               conf.image_info.dimensions.max_z > 1 ||
                                               conf.image_info.dimensions.max_t > 1))"
-                                title="Lock ${lock.LABEL}}">
+                                title="Lock ${lock.LABEL}">
                              <input type="checkbox"
                                  checked.bind="conf.sync_locks[lock.CHAR]"
                                  change.delegate="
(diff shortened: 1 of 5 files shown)


jburel:
Rebuilt on web-dev-merge with this PR included
Test 1:
* Open an image
* Double-click on the same image
* in multi-view: Lock the images
* Click on projection in one of them, the projection are synchronized as expected

Test 2:
* Open an image
* Click on projection
* Double-click on the same image
* in multi-view: Lock the images
* The second image "switches" to the projection mode when I move one of the slider's knob. If it will preferable to switch to the projection mode when the user locks the image

Also noticed (not related to the changes) that the tooltip of the synchronize options ((``Z/T``, View`` and ``Channels``) contains an extra ``}`` at the end

jburel:
Projection is now synchronized with ``Z/T`` as expected
One improvement that could included:

 * Open an image, double click on thumbnail
 * Lock the images
 * Make sure that Z/T is not selected
 * Click the projection icon in one the image
 * Click Z/T box in the same image. The Z/T box is correctly selected in the second image.
 * The projection is not turned on in the second image. I was expected to have the projection on in that case. It is activated when slider's knob is moved on the first image

will-moore:
@jburel I think that's the way it works for all "sync" actions. They are only sync'd when you change the settings in the window, not when you turn on/off various checkboxes.
E.g. when I turn on Channels checkbox, I don't want to immediately sync all channel settings. I may just want to sync the ON/OFF state when I toggle channels, but don't want to change rendering levels etc.
Same for Z/T - I may want to sync T but not sync Z (e.g. I have 2 windows that I want to scroll through T while looking at different Z sections). 
If we tried to sync everything when Z/T checkbox is clicked then this wouldn't be possible.

will-moore:
Working fine for me too. Looks good to merge.

waxenegger:
yes, in general the new additions to the sync group do not get updated with the present state of the others. 

To force a sync on checking of the individual categories can be done, provided that's what's wanted. The question in regards to it would be who gets whose state but I suppose it would be the window that the checkbox belongs to.

jburel:
Discussed with @waxenegger @will-moore, this can be adjusted depending on feedback
Merging

PR #196, opened 2018-04-04
Multi View - Saving Rendering Settings for All Open Images

waxenegger:
see: https://trello.com/c/ftiTGD9T/49-multi-view-save-actions

1. this PR aims at providing a means/shortcut via the header's File menu to save open images in multi view mode, without having to select them individually and do so.

2. it also offers a means to capture their canvases/viewports collectively (File menu option) - so far the png "screenshot" could only be done individually. the images will be saved in a zip.

Note for 1: If there were changes in an image that was opened and edited in more than one window this is treated as a "conflict" they won't get saved and a dialog shows the respective file(s).

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -1299,3 +1299,10 @@ thumbnail-slider img {
     min-width: 45px;
     max-width: 45px;
 }
+.skipped-list-files {
+    overflow-x: auto;
+    overflow-y: auto;
+    max-height: 150px;
+    font-size: 15px;
+    margin-top: 15px;
+}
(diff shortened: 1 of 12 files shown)


waxenegger:
--exclude

will-moore:
Both these features are working for me.
 - Save All Image Settings - saved and updated thumbnails as expected.
 - Save All Viewports as PNGs (zipped) also downloaded a zip with correct images.

screen shot 2018-05-09 at 10 50 48

NB: This PR is conflicting with #201 so it would be good to get this in ASAP so I can resolve the conflict.

PR #199, opened 2018-04-06
multiple_viewers_layout

will-moore:
This attempts to improve on the random layout of viewers when opened by double-clicking.

To test:

 - The first 4 viewers should be laid out in a 2 x 2 grid. 
 - Additional viewers have random location as before.
 - If any of the first 4 viewers are closed or moved out of their original position, then subsequent viewers will be opened in that spot.

Possible improvement: Could make the viewers bigger if screen space is larger?

screen shot 2018-04-06 at 17 17 39

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -757,8 +757,8 @@ export default class Context {
      * based on the position and size of each image viewer
      *
      * @memberof Context
-     * @param {x} id the ImageConfig id
-     * @param {y} forceRequest if true an ajax request is forced to update the data
+     * @param {x} x coordinate
+     * @param {y} y coordinate
      * @return {list} of ImageConfig objects
      */
     getImageConfigsAtPosition(x, y) {
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -753,6 +753,31 @@ export default class Context {
     }
 
     /**
+     * Retrieves any image configs that occupy a given x, y position,
+     * based on the position and size of each image viewer
+     *
+     * @memberof Context
+     * @param {x} id the ImageConfig id
+     * @param {y} forceRequest if true an ajax request is forced to update the data
+     * @return {list} of ImageConfig objects
+     */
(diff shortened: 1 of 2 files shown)


will-moore:
Now, we test each position in a 2 x 2 grid to see if any viewers are in that space, and if not then the new viewer opens there.

will-moore:
This seems to be merged in https://github.com/snoopycrimecop/omero-iviewer/commit/d5e2de8902ba04321a052f6375d0e9365bfb12bf but I don't see my changes in the app or in the http://web-dev-merge.openmicroscopy.org/static/omero_iviewer/main.js?_iviewer-1.0.0 file.
Any idea what's up @jburel (not urgent)?

jburel:
This only works if the screen is big enough
If I resize a bit the window after the first time, it does not lay them out nicely again.
An improvement but a "grid button" (cf. figure) will probably be better

will-moore:
There's always going to be ways you can make it behave strangely, or different ways to optimise depending on your priorities and screen size.
But I think it's an improvement, even with a small screen.
A grid button could make a nice addition in the same way as in figure, but (as with figure) the default should be to try position the viewers in a nice way without you having to click a "grid" button every time.
Is this "good to merge" or are there improvements to be made to the layout strategy?
I'd like to see what you're seeing with "If I resize a bit the window after the first time, it does not lay them out nicely again".

src/app/context.js
```diff
@@ -752,6 +752,31 @@ export default class Context {
         return this.getImageConfig(this.selected_config);
     }
 
+    /**
+     * Retrieves any image configs that occupy a given x, y position,
+     * based on the position and size of each image viewer
+     *
+     * @memberof Context
+     * @param {x} id the ImageConfig id
```
jburel: Could you update the description of the parameters?

jburel:
This is certainly an improvement
The feature will need to be reviewed later on
Merging

PR #201, opened 2018-04-24
Quantization family only if needed

will-moore:
This attempts to deal with the issue of long query strings for ```render_image``` if we add quantization settings to every channel, as described at https://github.com/ome/omero-iviewer/pull/144#issuecomment-378880392

Initial (simple) commit simply omitted the ```quantization``` from the ```maps``` if the family was "linear".
E.g. for a single channel image with "linear" family:
```
?maps=[{"inverted":{"enabled":false}]
```
However, in the case that this image has *saved* settings with a non-linear quantization, this would render the saved settings (instead of "linear"). 

So, we need to know when the current settings are different from the saved settings, so we can add the 'linear' family when the saved quantization is not linear.
```
maps=[{"inverted":{"enabled":false},"quantization":{"family":"linear"}}
```
We can simply make a copy of the ```channels_info_``` when the ```ol3 Image``` is created, BUT we also need to know when the user hits Save.
This required a new Event, triggered when the user hits Save, to update the ```saved_channel_info``` in the ol3 Image.

NB: we omit the "coefficient" when family is "linear" to avoid bug in webgateway that may not be fixed by the time that iviewer is released https://github.com/openmicroscopy/openmicroscopy/pull/5740 . However, this doesn't fix the coefficient range limited to 0-1 which is also fixed by that PR. NB: I'll close that PR while this one is tested, so we can check the behaviour of iviewer with current OMERO.

To test:
 - Open iviewer showing an image that has small number of channels (easier to see what's going on).
 - Open browser devtools, click Network tab and choose Img to see what image URLs are being requested.
 - When rendering setting changes are made, check the URL of rendered images. Note that the quantization is missing from the ```?map``` (see above).
 - Choose to set quantization family of Polynomial and adjust gamma. Image should be rendered correctly and URL should include "polynomial" setting and "coefficient".
 - Revert back to "linear" and the URL should revert to not include quantization. Also check the rendered image reverts to linear appearance.
 - Now switch to "polynomial" again and Save settings.
 - Then revert to "linear again". This time, the URL *will* include quantization "linear", but this shouldn't include "coefficient". Rendered image should appear correct.
 - Check that this works with multiple viewers (with *same* Image) updating correctly on Save
    - Open an Image with "linear" quantization in 2 viewers
    - In one viewer, choose "polynomial" and Save. Both viewers update.
    - In the other viewer, revert to "linear" and check that this is displayed correctly.

NB: I noticed that ```Save To All``` is ignoring quantization, so this will need fixing in another PR (webgateway). Can workaround by Save first, then Save to All.

### DIFF
--- a/ol3-viewer/src/ome/ol3/source/Image.js
+++ b/ol3-viewer/src/ome/ol3/source/Image.js
@@ -271,28 +271,32 @@ ome.ol3.source.Image = function(options) {
                 url += (!channelInfo['active'] ? "-" : "") + (c + 1);
                 url += "|" + channelInfo['start'] + ":" + channelInfo['end'];
                 url += "$" + channelInfo['color']; // color info
-                var m = {
-                    "inverted" : { "enabled" :
+
+                var m = {};
+                if (channelInfo['active']) {
+                    m["inverted"] = { "enabled" :
                         typeof channelInfo['inverted'] === 'boolean' &&
-                            channelInfo['inverted']}
-                };
-                // Only need to include family if different from default
-                var family = channelInfo['family'];
-                var family_not_default = (family !== "linear" ||
-                                          (family === "linear" && this.saved_channels_info_[c]["family"] !== "linear"));
-                if (typeof family === 'string' &&
-                    family !== "" &&
-                    family_not_default &&
-                    typeof channelInfo['coefficient'] === 'number' &&
-                    !isNaN(channelInfo['coefficient'])) {
-                        m["quantization"] = {
-                            "family": family,
-                        };
-                        // Only need coefficient if family is not 'linear' or 'logarithmic'
-                        if (family !== 'linear' && family !== 'logarithmic') {
-                            m["quantization"]["coefficient"] = channelInfo['coefficient'];
-                        }
-                };
+                            channelInfo['inverted']
+                    }
+
+                    // Only need to include family if different from default
+                    var family = channelInfo['family'];
+                    var family_not_default = (family !== "linear" ||
+                                              (family === "linear" && this.saved_channels_info_[c]["family"] !== "linear"));
+                    if (typeof family === 'string' &&
+                        family !== "" &&
+                        family_not_default &&
+                        typeof channelInfo['coefficient'] === 'number' &&
+                        !isNaN(channelInfo['coefficient'])) {
+                            m["quantization"] = {
+                                "family": family,
+                            };
+                            // Only need coefficient if family is not 'linear' or 'logarithmic'
+                            if (family !== 'linear' && family !== 'logarithmic') {
+                                m["quantization"]["coefficient"] = channelInfo['coefficient'];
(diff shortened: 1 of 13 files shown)


ol3-viewer/src/ome/ol3/source/Image.js
```diff
@@ -271,14 +276,21 @@ ome.ol3.source.Image = function(options) {
...
                 if (typeof channelInfo['family'] === 'string' &&
                     channelInfo['family'] !== "" &&
+                    family_not_default &&
                     typeof channelInfo['coefficient'] === 'number' &&
                     !isNaN(channelInfo['coefficient'])) {
                         m["quantization"] = {
                             "family": channelInfo['family'],
-                            "coefficient": channelInfo['coefficient']
                         };
+                        // Only need coefficient if family is not 'linear'
```
jburel: coefficient is not needed for logarithmic too

src/viewers/ol3-viewer.js
```diff
@@ -264,6 +266,8 @@ export default class Ol3Viewer extends EventSubscriber {
                     updates.push(chanUpdate);
                 };
                 this.viewer.changeChannelRange(updates);
+                // Viewer must know it's current settings are the saved settings
```
jburel: typo its

will-moore:
As discussed with @pwalczysko and @jburel, we can save more query length by only adding ```map``` for active channels.

pwalczysko:
All works as described in the header of this PR. 

@will-moore plans to stop showing the map quantization info for disabled channels. Happy for this to be done in this or follow-up PRs.

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#749](https://ci.openmicroscopy.org/job/iviewer-merge/749/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/749/consoleText) for more details.
Possible conflicts:
  - PR #196 waxenegger 'Multi View - Saving Rendering Settings for All Open Images'
    - src/viewers/ol3-viewer.js
    - src/settings/settings.js
  - PR #199 will-moore 'multiple_viewers_layout'
    - src/viewers/ol3-viewer.js

--conflicts

will-moore:
To test the last commits, need to retest all that's in description and check that nothing gets added to the ```?maps``` list if a channel is off (except an empty ```{}```). No 'inverted' or 'quantization' dictionaries.

jburel:
@pwalczysko will you have time to retest today?

will-moore:
Sorry - forgot to list this morning

pwalczysko:
Works as expected as per previous workflow. Additionally, also with the channels switched off works fine.

Ready to merge FMPOV

jburel:
Tested with an image with 76 channels
http://web-dev-merge.openmicroscopy.org/iviewer/?images=96303&dataset=21301
all channels active!

PR #202, opened 2018-05-15
Z-projection disabled big images

will-moore:
This disables the Z-projection button if the Image is loaded with tiles.
Currently iviewer loads images with tiles if ```sizeX * sizeY > 4000000```.

To test:
 - Find image where ```sizeX * sizeY > 4000000``` with Z-stack and check if Z-projection button disabled.
 - Clicking button should have no effect.
 - For images below this limit, Z-projection button should behave as before (disabled while Z movie is playing).


jburel:
tested using http://web-dev-merge.openmicroscopy.org/iviewer/?images=118090&dataset=23405

jburel:
works as expected

PR #206, opened 2018-08-08
Fix to lower case bug from for in loop

will-moore:
Something in the CI build of iviewer is adding an extra key ```__au_patched__``` to lists, so that using ```(for var x in list)``` is iterating over extra keys that are not in the list.
This is causing a bug, but will probably cause others once this is fixed.
E.g. in console of iviewer on web-dev-merge
```
test = ["a", "b"]
for (var t in test) { console.log(t, test[t]) }
    0 a
    1 b
    __au_patched__ 1
```

This leads to the following error
```
Uncaught TypeError: r.toLowerCase is not a function
    at wn (main.js?_iviewer-0.5.0:formatted:7720)
    at new pn (main.js?_iviewer-0.5.0:formatted:7466)
    at i.initViewer (main.js?_iviewer-0.5.0:formatted:57827)
```

The ```__au_patched__``` is probably coming from Aurelia ```(au)``` but can't find any info by googling for this.

### DIFF
--- a/ol3-viewer/src/ome/ol3/Viewer.js
+++ b/ol3-viewer/src/ome/ol3/Viewer.js
@@ -353,6 +353,7 @@ ome.ol3.Viewer.prototype.bootstrapOpenLayers = function(postSuccessHook, initHoo
     var dims = this.image_info_['size'];
     if (this.image_info_['zoomLevelScaling']) {
        var tmp = [];
+       console.log('zoomLevelScaling', this.image_info_['zoomLevelScaling'])
        for (var r in this.image_info_['zoomLevelScaling'])
            tmp.push(1 / this.image_info_['zoomLevelScaling'][r]);
        zoomLevelScaling = tmp.reverse();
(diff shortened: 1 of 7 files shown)


will-moore:
With that fix above, we are now seeing the next error as expected:
```
Uncaught TypeError: Cannot read property 'min' of undefined
```
There are going to be a LOT of places that are going to fail with the extra ```__au_patched__``` attribute of Array. Probably need to go through and fix them all....
Which is a good idea regardless: https://stackoverflow.com/questions/2265167/why-is-forvar-item-in-list-with-arrays-considered-bad-practice-in-javascript

jburel:
Back to normal with this PR on web-dev

will-moore:
@jburel The build went back to normal (no ```__au_patched__``` in Array). Not due to this PR, which only fixed 1 of the many errors that this causes.
I am in the process of fixing as many of these as I can find, since we don't understand the build error we don't know when it might happen again.

jburel:
@will-moore ok I did not have time to check build

will-moore:
Seems the build is OK and not had time to complete all fixes yet...

--exclude

PR #207, opened 2018-08-29
Thumbnail scrolling

will-moore:
See https://trello.com/c/JLpfTbmW/4-improve-thumbnails-loading-display

Instead of clicking up/down to load more thumbnails, the thumbnail panel is created at the full size (with only the visible thumbnails loaded) and more thumbnails are loaded when we scroll.
There are lots of different cases to handle and a lot has changed so everything needs testing:

 - Expected behaviour: any thumbnails that appear in the panel should be loaded (no empty grey squares).
 - While actually scrolling we don't load thumbnails, since that would load the whole dataset if you scroll fast from one end to the other. Instead, wait until scrolling slows / stops, then load thumbs for visible thumbnails.
 - You can see when loading is happening in the devtools Network tab.
 - In all scenarios below, check thumbs on load, then scroll about, then refresh thumbs (when scrollbar is in various locations).
 - Check that IE works.

**Scenarios to test:**

 - Single image in 1 or more Datasets, double-clicked in webclient. url like ```/webclient/img_detail/4413/?dataset=1301```
        - Check that thumbs loaded from *current* parent Dataset (not a different Dataset that the image is in)
        - Try double-clicking images in webclient near the start of the Dataset and at the end. Thumb panel in iviewer should scroll to the show the selected image.
        - Try with a small Dataset (< 5 images) and as large as possible.
        - Remove the ```?datasets=ID``` from the URL and reload. If image is in single Dataset, thumbs should be loaded. If more than 1 Dataset, thumbnail panel hidden.
 - Multiple selected images > Open with iviewer. Url like ```/iviewer/?images=4402,4403,4404,4405 ```
 - Orphaned image, double-clicked in webclient. e.g. ```webclient/img_detail/57012/``` should hide thumbnail panel.
 - Well - openwith > iviewer.  Url ```/iviewer/?well=148``` First image should be selected.
 - Datasets - openwith > iviewer. ```/iviewer/?dataset=10480``` First image selected.

### DIFF
--- a/src/app/thumbnail-slider.js
+++ b/src/app/thumbnail-slider.js
@@ -290,7 +290,6 @@ export default class ThumbnailSlider extends EventSubscriber {
      */
     initializeThumbnails(refresh = false) {
         // standard case: we are an image
-        console.log("initializeThumbnails", refresh, this.context.initial_type);
         if (this.context.initial_type === INITIAL_TYPES.IMAGES) {
             if (this.context.initial_ids.length > 1) {
                 // we are a list of images
@@ -316,7 +315,6 @@ export default class ThumbnailSlider extends EventSubscriber {
      * @memberof ThumbnailSlider
      */
     gatherThumbnailMetaInfo(image_id) {
-        console.log("gatherThumbnailMetaInfo", image_id);
         let url =
             this.context.server + this.webclient_prefix + "/api/paths_to_object/?" +
             "image=" + image_id + "&page_size=" + this.thumbnails_request_size;
@@ -358,28 +356,16 @@ export default class ThumbnailSlider extends EventSubscriber {
                     }
                 }
 
-                console.log("imgInf.parent_type, imgInf.parent_id", imgInf.parent_type, imgInf.parent_id);
-
                 if (imgInf.parent_id === null) {
                     this.hideMe();
                     return;
                 }
 
-                // let initialize = (parent === null);
-                // TODO: why would initialize not be true here?
-                let initialize = true;
-                console.log('parent', parent);
-
                 let start_index = 0;
                 if (parent) {
-                    // set count, start and end indices (if count > limit)
-                    // this.setThumbnailsCount(parent.childCount);  --
-                    // if (this.thumbnails.length > this.thumbnails_request_size) {
-                        // start_index = (parent.childPage - 1) * this.thumbnails_request_size;
-                        start_index = parent.childIndex;
-                        console.log('parent.childIndex', parent.childIndex);
-                    // }
+                    start_index = parent.childIndex;
                 }
+                let initialize = true;
                 this.requestMoreThumbnails(initialize, false, start_index);
             },
             error : (response) => this.requestMoreThumbnails(true)
@@ -414,12 +400,9 @@ export default class ThumbnailSlider extends EventSubscriber {
             if (panel)
             scrollTop = panel.scrollTop;
         }
-        console.log('loadVisibleThumbnails scrollTop:', scrollTop);
 
         let thumb_start_index = parseInt(scrollTop / this.thumbnail_size);
         let thumb_end_index = parseInt((scrollTop + this.slider_height) / this.thumbnail_size);
-        console.log('this.thumbnail_size', this.slider_height);
-        console.log('scrolled to thumb_index...', thumb_start_index, thumb_end_index);
 
         let unloaded = [];
         for (var idx=thumb_start_index; idx<=thumb_end_index; idx++){
@@ -430,7 +413,6 @@ export default class ThumbnailSlider extends EventSubscriber {
         if (unloaded.length > 0) {
             thumb_start_index = unloaded[0];
             thumb_end_index = unloaded[unloaded.length-1];
-            console.log('unloaded thumbs: thumb_start_index, thumb_end_index', thumb_start_index, thumb_end_index);
             this.requestMoreThumbnails(init, refresh, thumb_start_index, thumb_end_index);
         }
     }
@@ -464,7 +446,6 @@ export default class ThumbnailSlider extends EventSubscriber {
             limit = thumb_end_index - offset;
         }
 
-        console.log("requestMoreThumbnails... offset, limit", offset, limit);
 
         let url = this.context.server;
         if (this.context.initial_type === INITIAL_TYPES.DATASET ||
@@ -484,7 +465,6 @@ export default class ThumbnailSlider extends EventSubscriber {
             {url : url,
             success : (response) => {
                 this.requesting_thumbnail_data = false;
-                console.log("...requestMoreThumbnails init", init, this.thumbnails.length);
                 if (init) {
                     // we want the total count
                     if (typeof response !== 'object' || response === null ||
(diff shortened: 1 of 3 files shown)


mtbc:
I suspect that dropping IE11 (or not) should be a clear, conscious decision given that we often choose to support old stuff.

jburel:
Basic testing:
* Open dataset with < 10 Images by double-clicking on an image. The correct thumbnails were loaded
* Open a larger dataset (~80 images) visible thumbnails loaded. Scroll down to the bottom: more visible thumbnails are then loaded. Moving backward forward to load the rest of the thumbnails

jburel:
``
- Remove the ?datasets=ID from the URL and reload. If image is in single Dataset, thumbs should be loaded. If more than 1 Dataset, thumbnail panel hidden.
- Datasets - openwith > iviewer. /iviewer/?dataset=10480 First image selected.
- Well - openwith > iviewer. Url /iviewer/?well=148 First image should be selected.
- Orphaned image, double-clicked in webclient. e.g. webclient/img_detail/57012/ should hide thumbnail panel.
This works fine
``
Multiple selected images > Open with iviewer. Url like /iviewer/?images=4402,4403,4404,4405
``
If I select all the images in the dataset All the thumbnails are loaded
but If I open the dataset or double click on a thumbnail to open the viewer. Only few thumbnails are loaded.

jburel:
I don't think we can drop IE

jburel:
Safari:
* Clicking on well with only one field does not hide the thumbnails panel. It is empty
pr-207_safari

jburel:
Noticeable loading difference between FF and chrome 
In FF I have time to see "Not Found"
even for small dataset

jburel:
FF 
Same problem as the one described in https://github.com/ome/omero-iviewer/pull/207#issuecomment-421030884

This time
it does not work for well with multiple fields.
pr-2017-ff

mtbc:
I too saw that "Not Found": https://openmicroscopy.slack.com/archives/C0K5EED3R/p1535619171000100

jburel:
kymograph is correctly displayed cc @pwalczysko 
pr-207-kymo

jburel:
While testing 5.4.8 
Looking at data belonging to another user when that user has not viewed the image
leads to "Not found" in the left-hand panel and empty bird eye view
we should handle such case a bit better

will-moore:
@jburel Re: "If I select all the images in the dataset All the thumbnails are loaded
but If I open the dataset or double click on a thumbnail to open the viewer. Only few thumbnails are loaded."
As long as all the thumbnails that are *visible* are loaded then I think this is OK?
I assume that the number of images listed with ?image=1,2,3 is not going to be larger than a safe number of thumbnails to load (e.g. 200, or even more).

jburel:
All the thumbnails were loaded even the non visible when I do a selection of images
I probably select ~100

will-moore:
Also noticed that e.g. http://web-dev-merge.openmicroscopy.org/iviewer/?well=339085 (user-1) doesn't work (no thumbnails or image loaded).

will-moore:
Re: https://github.com/ome/omero-iviewer/pull/207#issuecomment-421319766 The 'Not Found' in this case is the same as previously (not related to this PR). To fix that we could use /webclient/render_thumbnail/ID instead of /webgateway/render_thumbnail. Then we'd get the 'No Image'  icon same as webclient?

will-moore:
Fixed the placeholder "Not Found" while images loading in Firefox.
Also fixed bugs with SPW images in a Well (```/webclient/img_detail/ID/```) and opening Well directly (```/iviewer/?well=ID```). In both cases, the thumbnails within the Well should be shown in the thumbnail_slider, even if it's only 1 field (same behaviour as before).

src/app/thumbnail-slider.js
```diff
@@ -700,6 +785,15 @@ export default class ThumbnailSlider extends EventSubscriber {
             }, 2000);
     }
 
+    setThumbnailsCount(count) {
+        console.log('setThumbnailsCount:', count, 'was:', this.thumbnails.length);
+        if (typeof count !== 'number') {
+            count = 0;
+        }
+        this.thumbnails.length = count;
+        this.thumbnails.fill({'version': 0, 'title': 'unloaded'});
```
mtbc: Given https://github.com/ome/omero-iviewer/pull/207#issuecomment-421028262 is this okay?

jburel:
* Wells with 1 or multiple fields are now correctly displayed in FF,  Vivaldi and Safari
* I also select a large number of images, loading occurred when scrolling down the list

No cover yet: IE testing

src/app/thumbnail-slider.js
```diff
@@ -100,25 +101,18 @@ export default class ThumbnailSlider extends EventSubscriber {
...
+     * size of thumbnails in slider (height + margin)
      * @memberof ThumbnailSlider
      * @type {number}
      */
-    thumbnails_start_index = 0;
+    thumbnail_size = 90;
 
     /**
-     * the end index of the displayed thumbnails
+     * height of slider, so we which thumbs have scrolled into view
```
jburel: missing verb

src/app/thumbnail-slider.js
```diff
@@ -141,22 +135,25 @@ export default class ThumbnailSlider extends EventSubscriber {
...
+                [IMAGE_VIEWER_RESIZE,
+                    (params={}) => this.resizeViewer(params)]];
 
     /**
      * @constructor
      * @param {Context} context the application context (injected)
      * @param {BindingEngine} bindingEngine the BindingEngine (injected)
      */
-    constructor(context, element, bindingEngine) {
+    constructor(context, element, bindingEngine, taskQueue) {
```
jburel: Declare new parameter

src/app/thumbnail-slider.js
```diff
@@ -314,94 +329,105 @@ export default class ThumbnailSlider extends EventSubscriber {
...
+            '@id': id,
+            Name: 'image: ' + id
+        }));
+        this.addThumbnails(to_add, 0);
+    }
+
+    /**
+     * Loads unloaded thumbnails that scrolled into the thumbnail panel
+     *
+     * @param {number} scrollTop Current scroll position of the panel
```
jburel: declare parameter

src/app/thumbnail-slider.js
```diff
@@ -480,35 +510,59 @@ export default class ThumbnailSlider extends EventSubscriber {
     }
 
     /**
-     * Adds thumbnails to then be loaded.
+     * Adds thumbnails with Image IDs to this.thumbnails array.
+     * We create a new list from old, replacing the thumbnails (instead of
+     * modifying the old list) to make sure the changes is observed and
```
jburel: is/are

src/app/thumbnail-slider.js
```diff
@@ -480,35 +510,59 @@ export default class ThumbnailSlider extends EventSubscriber {
...
-                if (!skip_decrement) this.thumbnails_start_index--;
+                return thumb;
+            }
+        });
+    }
+
+    /**
+     * Adds thumbnails with Image IDs to this.thumbnails array.
+     * We create a new list from old, replacing the thumbnails (instead of
+     * modifying the old list) to make sure the changes is observed and
```
jburel: is/are

src/app/thumbnail-slider.js
```diff
@@ -480,35 +510,59 @@ export default class ThumbnailSlider extends EventSubscriber {
...
+     * UI updates.
+     *
+     * @param {Array.<Object>} thumbnails the thumbnails to be added
```
jburel: thumbnails not a parameter

jburel:
Testing in IE 11 
it works nicely too

PR #208, opened 2018-09-03
Cache image settings

will-moore:
See https://trello.com/c/22S56JnA/2-single-view-cache-settings

To test: 
 - View image, edit settings:
     - Channels on/off, pick channel colours / LUTS, adjust sliders
     - Channels advanced settings - gamma, family etc.
     - Greyscale checkbox
     - Pan, rotate and zoom image
     - Z and T indexes
     - Z-projection
 - Move to a different image by clicking on thumbnail
 - Return to original image by clicking on thumbnail
 - Viewer should appear same as when you left it (same rendering settings, zoom, pan)
 - Try with multiple viewers. Not sure of the precise behaviour we want here but does it behave unexpectedly?

Known issues: 
 - moving back and forwards several times between images without panning/zooming/rotating, pan/zoom/rotate gets reset.
 - projection doesn't get re-applied in some cases with multiple viewers.


jburel:
* Click on FullRange the image is updated.
* Move to another Image
* Go back to the first Image
* The Image displayed corresponds to Full Range. but the Min/Max button is selected

jburel:
We could probably cache Histogram  and Interpolate

will-moore:
@jburel The Min/Max, Full Range & Imported buttons probably shouldn't be 'radio' buttons where the last-clicked button is 'selected' since the settings don't continue to be 'Full Range' etc.
Currently the 'Min/Max' button is always selected, even if you've never set an image to Min/Max.

jburel:
discussed with @will-moore 
do not select any button e.g. min/max is always selected even if the window input != [min, max]

will-moore:
The Min/Max, Full Range & Imported buttons aren't 'selected' now (don't show state).
Also, the cached rendering settings, greyscale and Z/T are added to Undo queue so that the Save button is enabled. This also means you can use Undo to return to the unsaved state. NB: minor issue - to reuse existing methods for 'pasting' settings, the channel settings are added to the Undo queue separately from Z/T and greyscale. So, you may need 2 clicks on Undo to return to unsaved state.

To test:
 - Adjust rendering settings (including Invert, LUTS, family, coefficient etc)
 - Adjust Z/T and greyscale
 - Move to another image and back again.
 - See that all settings are applied and that Save and Undo are enabled.
 - Save should work as expected. Undo will undo the channels settings first, then Z/T & greyscale.

jburel:
After clicking on any of the range button e.g. Min/Max it stays "selected". I think this should not be the case

The rest works nicely.

src/model/image_info.js
```diff
@@ -385,6 +388,79 @@ export default class ImageInfo {
...
+                              old_val : this.dimensions.t,
+                              new_val: cached.t,
+                              type : "number"});
+                response.rdefs.defaultT = cached.t;
+                this.dimensions.t = cached.t;
+            }
+            if (cached.projection) {
+                // Don't need to update response since Z dimension-slider
+                // will update the ol3-viewer
+                // response.projection = cached.projection;
```
jburel: remove commented out code

jburel:
In multi-view it does not really work:
* Double-click on the same image to enable multi-view
* Do not synch the images
* Turn off one channel for example
* Select another image
* The new image will be correctly displayed in the window with focus
* Go back to the original settings.
* It does not display the last viewed settings

will-moore:
Last commit should fix the multi-view issues described above

rgozim:
Latest commit solved the bug described by @jburel

pwalczysko:
In multi-view, when 2 viewports of the same image are open, then

   - if there are two different rnd settings in each of the two viewports, then showing another image in the viewport and going back to the orginal one will have just one rnd setting "remembered" for both viewports with the same image (I think this is acceptable - I guess you can "remember" just one rnd per image, right ?)

pwalczysko:
Bug:

    - Projection seems to be not working too well, or the caching of it is not right



screen shot 2018-09-26 at 10 59 32


user-4, http://web-dev-merge.openmicroscopy.org/webclient/?show=image-79356

Workflow

    - navigate to the image
    - open it in iviewer
    - change Rnd, selecting a LUT
    - click on Projection button, drag the sliders to extreme max and min possible
    - navigate to another image
    - come back to the previous image (the one you did Projection on)
    - observe disrcrepancy in what you saw before, and what you see now when you came back
    - save the rnd -> observe discrepancy between thumb and full viewer (see screenshot)

pwalczysko:
I think the problem https://github.com/ome/omero-iviewer/pull/208#issuecomment-424658770 stems from the fact that Projection is not cached, it is not even saved after Save button is clicked. This might seem "logical" to us, but for a user: I do not see why Projection is so "special".

pwalczysko:
>moving back and forwards several times between images without panning/zooming/rotating, pan/zoom/rotate gets reset.

confirmed. this might be one to have a look again, but I don't think it is a blocker.

pwalczysko:
In summary, three issues, https://github.com/ome/omero-iviewer/pull/208#issuecomment-424657485 (just one Rnd remembered per image, maybe a valid approach), https://github.com/ome/omero-iviewer/pull/208#issuecomment-424658770 (projection - there is clearly a regression bug) and https://github.com/ome/omero-iviewer/pull/208#issuecomment-424668920 - maybe for improvement later in another PR - obviously a deficiency, but one with which it would be possible to live

will-moore:
Fixed the toggling of Z-projection having no effect on the image-viewer.

src/model/image_config.js
```diff
@@ -150,6 +151,129 @@ export default class ImageConfig extends History {
         this.revision++;
     }
 
+    /**
+     * Applies the rendering settings, keeping a history of the old settings
+     *
+     * @param {Object} rdef the rendering defintion
```
jburel: typo

src/model/image_config.js
```diff
@@ -150,6 +151,129 @@ export default class ImageConfig extends History {
...
+     * @param {boolean} for_pasting true if rdef supplied it for pasting of settings, false otherwise
```
jburel: it/is

jburel:
I have recorded the comment https://trello.com/c/c3BU4L1x/63-cache-settings-follow-up
@will-moore is working on an alternative solution. Recording in case @will-moore cannot finish the alternative option so it does not get forgotten

PR #209, opened 2018-09-21
Enable and workaround for doubleClickZoom functionality

will-moore:
See https://trello.com/c/WqYPu4Uj/7-zoom-double-click

To test:
 - Double-clicking on the image should zoom in.
 - Shift-double-click should zoom out.
 - Check that the amount we zoom in is reasonable on various images (big vv small) at various zoom levels.

NB: Some strange behaviour where the config passed to ```doubleClickZoom({'delta': 10})``` was being ignored required a workaround.
Possibly because of issues with compiling, minimising and the null check at https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/DoubleClickZoom.js#L36 which was always reverting to ```1``` default whereas the ```duration``` check below was respecting the a duration value passed in.


will-moore:
Ooops - opened against wrong branch. Closing...

PR #210, opened 2018-09-21
Enable and workaround for doubleClickZoom functionality

will-moore:
See https://trello.com/c/WqYPu4Uj/7-zoom-double-click

To test:
 - Double-clicking on the image should zoom in.
 - Shift-double-click should zoom out.
 - Check that the amount we zoom in is reasonable on various images (big vv small) at various zoom levels.

NB: Some strange behaviour where the config passed to ```doubleClickZoom({'delta': 10})``` was being ignored required a workaround.
Possibly because of issues with compiling, minimising and the null check at https://github.com/openlayers/openlayers/blob/master/src/ol/interaction/DoubleClickZoom.js#L36 which was always reverting to ```1``` default whereas the ```duration``` check below was respecting the a duration value passed in.

### DIFF
--- a/ol3-viewer/src/ome/ol3/globals.js
+++ b/ol3-viewer/src/ome/ol3/globals.js
@@ -324,7 +324,7 @@ ome.ol3.AVAILABLE_VIEWER_CONTROLS = {
  * <li>keyboardZoom [defaults:true, enabled: false]</li>
  * <li>shiftDragZoom [defaults:true, enabled: false]</li>
  * <li>shiftDragRotate [defaults:true, enabled: false]</li>
- * <li>doubleClickZoom [defaults:true, enabled: false]</li>
+ * <li>doubleClickZoom [defaults:true, enabled: true]</li>
  * <li>pinchRotate [defaults:true, enabled: false]</li>
  * <li>boxSelect [defaults:false, enabled: false]</li>
  *</ul>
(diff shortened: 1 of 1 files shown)


pwalczysko:
Works fine on Firefox Mac and IE Win7 and FF Win7

Ready to merge FMPOV

ol3-viewer/src/ome/ol3/globals.js
```diff
@@ -428,6 +430,18 @@ ome.ol3.defaultInteractions = function() {
                     "ref": new Constructor(V['options']),
                     "defaults" : V["defaults"]
                 };
+                // For some reason, calling new ol.interaction.DoubleClickZoom({"delta": 10})
+                // here, the 'delta' option is ignored. But if we create a defaults collection
+                // the same class is created and the {'delta': zoomDelta} is not ignored.
+                // We use altShiftDragRotate: false so that doubleClickZoom should be
+                // the first item in the ol.Collection.
+                if (K === "doubleClickZoom") {
```
waxenegger: @will-moore I wanted to just quickly point you towards what's a bit of deceptive behavior from the closure compiler but in this case you should be able to achieve the same without this special if but specifying the properties you want to set as a string, e.g `"options": {"duration" : 1000, "delta": 10}` 
The reason I say this is that I suspect very strongly that the closure compiler mangles the property which it does not do if supplied in "string notation". It never touches strings.

Now I don't think this is a major problem any more but I refrained from using ECMA2015 constructs in the ol3 code such as "let". As is the ol3 viewer gets built by the closure compiler and is then used as a import and as such does not go through the babel transpiler. Problems nowadays should be rare but I did encounter one problem in the past using such a feature and IE11 (not 'let' though but one of the ones that got implemented later on...), that's why I'm mentioning it -for completeness sake ;-)

will-moore: Hi @waxenegger, thanks for the input. I played a bit more and things seem to be working better now. When I tried it before with e.g. ```{"duration" : 1000, "delta": 10}``` the ```duration``` was getting recognised but the ```delta``` wasn't.
Anyway, I seem to have it working without the workaround now in #213.

PR #211, opened 2018-09-27
Always show ROI/Shape owner. Unsaved if no shape ID

will-moore:
See https://trello.com/c/gWqOfidt/61-show-better-tooltips-on-rois

This change displays the Owner of the Shape even if that is the current user or the current user is able to edit the ROI (e.g. admin). Previously we only showed owner if they couldn't edit the Shape.

Also, if the shape is unsaved, tooltip show "Unsaved" instead of showing negative ROI and Shape IDs.

To test:
 - check the tooltip on Shapes you own (and others)
 - check tooltip on unsaved vv saved shapes.
 - check tooltip on multiple shapes within ROI (created in Insight)

### DIFF
--- a/src/regions/regions-list.html
+++ b/src/regions/regions-list.html
@@ -79,7 +79,8 @@
                     <div if.bind="(roi.shapes.size - roi.deleted) > 1 &&
                                    roi.shapes.size > 1 && $first"
                         id="${'roi-' + roi_id}"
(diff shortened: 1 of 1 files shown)


will-moore:
This seems to be working on web-dev-merge now (although I noticed it doesn't show Owner of first save of ROI, but does when you load the image with existing ROIs):

screen shot 2018-10-08 at 06 37 18

rgozim:
Looks good to merge.

PR #217, opened 2019-01-07
Add textStatus to ajax error message.

will-moore:
This could help users reporting problems with image loading.

During investigation of https://trello.com/c/X5egtRjc/3-iviewer-fails-to-load-images-with-pixel-size-nan-or-inf it would have been more useful to get the extra info on why image loading failed, in this case the ```textStatus``` was "parsererror":

screen shot 2019-01-07 at 03 57 34

This can be tested by opening http://web-dev-merge.openmicroscopy.org/webclient/?show=image-24401 in iviewer (user-3)

### DIFF
--- a/src/model/image_info.js
+++ b/src/model/image_info.js
@@ -312,7 +312,7 @@ export default class ImageInfo {
                 if (this.context.isRoisTabActive())
                     conf.regions_info.requestData();
             },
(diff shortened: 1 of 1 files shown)


jburel:
This will help the reporting of issue. 
Thanks

PR #227, opened 2019-01-29
Vectortile investigation

will-moore:
This is the first steps towards support of large numbers of ROIs in iviewer.

PR opened in 2019 and again in 2025, and updated to use an OMERO.table of the Shapes in the Image to support loading shapes by tile/region.

It uses a OpenLayers ```VectorTile``` layer to load shapes by tile.

Setup...

 - I used https://github.com/will-moore/python-scripts/blob/018b3aee645fa38d0a7207a756b701356d5766e8/skimage_segmentation.py to create ~250000 polygons on a big image (see screenshot)
 - Then ran https://github.com/will-moore/python-scripts/blob/018b3aee645fa38d0a7207a756b701356d5766e8/roi_shape_boundingboxes_to_table.py on the Image to create an OMERO.table with columns `X1, X2, Y1, Y2` to define the bounding box of each Shape.  OMERO.table has namespace `omero.shape.boundingbox.coords`.
 - The `iviewer/shapes_by_region/IMAGEID/Z/T?tile=ZOOM,ROW,COL,WIDTH,HEIGHT` request opens the table, finds rows that have an intersecting bounding box and uses the `shape_id` to load the shapes from OMERO.

Screenshot 2025-03-27 at 15 51 32

Limitations / questions:
  - Currently we only allow loading of Shapes for the 100% zoom level. We don't have a "Pyramid" of objects as we do for tiles. If we want to be able to browse the whole image to see which regions have some ROIs in them, we'll need to think how to store a Pyramid of ROIs. Currently, clicking on a shape in the ROI Table browses to that shape at 100% zoom so you can see it.
 - In this PR, if the image has more ROIs than a single "page" then we use a ```VectorTile``` layer and ```TiledRegions``` source *instead* of the ```Vector``` layer and ```source/Regions``` class. This means that all the shape drawing and manipulation code in the ```source/Regions``` class is not available, so we can't drag, edit or draw shapes. If we want to support this in future we could possibly:
    - Use a ```VectorTile``` layer for showing existing shapes and add a ```Vector layer + Regions``` overlay to allow drawing new shapes.
    - OR investigate the drawing and editing of shapes in a ```TiledRegions``` source. This would probably involve splitting the existing ```source/Regions``` class to separate the drawing & editing code from the ```ol.source.Vector``` base class to allow reuse. Possibly a lot of refactoring!

Testing
=====

 - Open the image in iviewer.
 - Loading ROIs tab should load the first 500 ROIs as before (SLOW!)
 - Then the centre panel loads ROI tiles for current T index.
 - Moving through T should load ROI on the image for new Time-point.
 - Pagination controls above the ROI table should work as expected (previous ```<``` and next ```>``` buttons and a range slider) and reload ROIs for the chosen page.
 - Clicking Shapes in the ROI table should take you to that shape and show it selected.
 - Clicking Shapes on the Image should select them and highlight the corresponding shape in the table IF that shape is currently loaded in the table. If NOT, we need a call to work out which page of ROIs the shape belongs to, then we load that page of the table (SLOW)!!
 - See that show/hide of shapes is working
 - Known issue: Labels aren't scaled correctly for Tiled ROIs - we don't have access to the image's scale in the same was as for Vector layer.
 - No creation or editing of ROIs is currently possible if > 500 ROIs. Otherwise this should behave as normal.


will-moore:
As discussed with @jburel, the ROI table pagination alone is a big win (allows access to all ROIs) and could be released ahead of the ROI Tiling. Moved that to #231.

PR #231, opened 2019-02-06
Roi table pagination

will-moore:
This now paginates the ROI table by Z/T plane for multi-plane images.
This is a *much* nicer experience than simple pagination of all ROIs (testing with the Multi-T image below). It should allow you to work with ROIs "as normal" for images with less than 500 ROIs per plane. For Big Images (or any image planes with > 500 ROIs per plane) we paginate the ROIs within that plane.

To test, open an image with lots of ROIs (user-3):
 - Multi-Z/T/C image http://web-dev-merge.openmicroscopy.org/webclient/?show=image-161151
    - NB: Not all planes have ROIs on them. Some Shapes have C set, most don't.
 - Browsing, editing & deleting ROIs/Shapes on this image should work "as normal" *except* that you can't browse the ROI table for shapes on other planes. E.g:
    - If you create more ROIs (drawing or copy/paste) then save and go to another time-point and back to current time-point, you should see the ROIs you created.
    - Delete all the ROIs on a particular Z/T plane - save.
    - Edit existing ROIs and save.
    - Setting T or Z to 'unattached' for a Shape should make it show up in all T / Z planes.
    - Create, edit or delete a shape, then try to change Z/T or 'page' (see below). Should get "Please save first" dialog.
 - If there are more than 500 ROIs to show, then we do pagination. E.g:
    - Enable Z-projection for the image above. ROI table will show all ROIs within the Z-range.
    - Choose a plane, select all ROIs, copy and paste (then drag them and set colour so as to set them apart from others). Repeat until you have > 500 ROIs for that plane. Save. Then go to a different plane and back to this one and you should see it paginated.
    - Can change page by ```< or >``` buttons, slider or by putting number in input. No updates possible while the ROIs are loading and invalid numbers should be ignored. Also, no change possible with unsaved ROIs.
    - Big Image: http://web-dev-merge.openmicroscopy.org/webclient/?show=image-160953
 - Behaviour should be completely unchanged on images with < 500 ROIs.

screen shot 2019-02-12 at 03 05 10

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -206,6 +206,49 @@ def persist_rois(request, conn=None, **kwargs):
 
     return JsonResponse(ret)
 
+@login_required()
+def rois_by_plane(request, image_id, the_z, the_t, conn=None, **kwargs):
+    """
+    Get ROIs with Shapes where Shapes are on the specified Z and T plane.
+
+    Includes Shapes where Z or T are null.
+    """
+
+    params = omero.sys.ParametersI()
+    params.addId(image_id)
+    filter = omero.sys.Filter()
+    filter.offset = rint(request.GET.get("offset", 0))
+    filter.limit = rint(request.GET.get("limit", 1000))
+    params.theFilter = filter
+    query = """
+        select roi from Roi roi
+        join fetch roi.details.owner as owner
+        join fetch roi.details.creationEvent
+        left outer join fetch roi.shapes as shapes
+        where (shapes.theZ = %s or shapes.theZ is null)
+        and (shapes.theT = %s or shapes.theT is null)
+        and roi.image.id = :id order by roi.id""" % (the_z, the_t)
+
+    query_service = conn.getQueryService()
+    rois = query_service.findAllByQuery(query, params, conn.SERVICE_OPTS)
+
+    marshalled = []
+    for r in rois:
+        encoder = omero_marshal.get_encoder(r.__class__)
+        if encoder is not None:
+            marshalled.append(encoder.encode(r))
+
+    # Modify query to only select count() and NOT paginate
+    query = query.replace("select roi ", "select count(distinct roi) ")
+    query = query.replace("fetch", "")
+    query = query.split("order by")[0]
+    params = omero.sys.ParametersI()
+    params.addId(image_id)
+    result = query_service.projection(query, params, conn.SERVICE_OPTS)
+    meta = {"totalCount": result[0][0].val}
+
+    return JsonResponse({'data': marshalled, 'meta': meta});
+
 
 @login_required()
 def image_data(request, image_id, conn=None, **kwargs):
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -39,8 +39,13 @@ import omero.util.pixelstypetopython as pixelstypetopython
 from version import __version__
 from omero_version import omero_version
 
+from . import iviewer_settings
+
 WEB_API_VERSION = 0
 MAX_LIMIT = max(1, API_MAX_LIMIT)
+
+ROI_PAGE_SIZE = getattr(iviewer_settings, 'ROI_PAGE_SIZE')
+ROI_PAGE_SIZE = min(MAX_LIMIT, ROI_PAGE_SIZE)
 
 PROJECTIONS = {
     'normal': -1,
@@ -77,6 +82,7 @@ def index(request, iid=None, conn=None, **kwargs):
         'api_base', kwargs={'api_version': WEB_API_VERSION})
     if settings.FORCE_SCRIPT_NAME is not None:
         params['URI_PREFIX'] = settings.FORCE_SCRIPT_NAME
+    params['ROI_PAGE_SIZE'] = ROI_PAGE_SIZE
 
     return render(
         request, 'omero_iviewer/index.html',
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -207,11 +207,14 @@ def persist_rois(request, conn=None, **kwargs):
     return JsonResponse(ret)
 
 @login_required()
-def rois_by_plane(request, image_id, the_z, the_t, conn=None, **kwargs):
+def rois_by_plane(request, image_id, the_z, the_t, z_end=None, t_end=None,
+                  conn=None, **kwargs):
     """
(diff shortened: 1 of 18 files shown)


pwalczysko:
screen shot 2019-02-07 at 17 11 17

I think the numbers here should be editable. I would like to navigate to the same pages, without making too many mistakes (it takes quite a while before the ROIs on that page load, so it is costly to land on a wrong page)

jburel:
Looking at http://web-dev-merge.openmicroscopy.org/webclient/img_detail/160953/?dataset=27801
* When a shape is selected in the right-hand pane, the shape is selected in the image. We should probably "zoom in"

jburel:
* Select a new roi page
* During the loading phase, the drawing controls are visible and can be selected but the user cannot draw. They should be greyed out

jburel:
Loading of ROIs should be turned off when playing a movie across Z or T, only attempt to load the data when the movie has finished e.g. http://web-dev-merge.openmicroscopy.org/webclient/img_detail/161151/?dataset=27801

plugin/omero_iviewer/views.py
```diff
@@ -207,6 +207,75 @@ def persist_rois(request, conn=None, **kwargs):
...
+    If z_end or t_end are not None, we filter by any shape within the
+    range (inclusive of z/t_end)
+    """
+    query_service = conn.getQueryService()
+
+    params = omero.sys.ParametersI()
+    params.addId(image_id)
+    filter = omero.sys.Filter()
+    filter.offset = rint(request.GET.get("offset", 0))
+    filter.limit = rint(request.GET.get("limit", 1000))
```
jburel: limit should come configuration

plugin/omero_iviewer/views.py
```diff
@@ -207,6 +207,75 @@ def persist_rois(request, conn=None, **kwargs):
...
+        join roi.shapes as shapes
+        where (%s) and (%s)
+        and roi.image.id = :id
+    """ % (where_z, where_t)
+
+    params = omero.sys.ParametersI()
+    params.addId(image_id)
+    filter = omero.sys.Filter()
+    filter.offset = rint(request.GET.get("offset", 0))
+    filter.limit = rint(request.GET.get("limit", 1000))
```
jburel: same

will-moore:
Those commits should address these issues raised above:
 - ROI drawing tools disabled while ROIs are loading
 - Don't load ROIs while movie is playing - load when it stops (via user or end of T)
 - Also should keep ROIs in sync if user hits > button lots (instead of playing movie)
 - ROI_PAGE_SIZE is configurable. ``` bin/omero config set omero.web.iviewer.roi_page_size```
 
I think we made the decision not to zoom in when an ROI is selected since you may already be at the zoom level you want and if you're moving back and forwards between working with ROIs on the image and in the table, zooming each time you select an ROI is going to be annoying. We even decided not to pan the image if the ROI is within the view-port for the same reason.

jburel:
For the zoom level, we might have to revisit that with more feedback. Let's keep it as it is for now

pwalczysko:
Does not work in IE, the ROI table never loads on images with many ROIs.

pwalczysko:
Win 7 on Firefox and Chrome this works fine. Chrome is slightly faster.

will-moore:
@pwalczysko I think the failure to show ROIs in the table was a layout issue in IE.
Hope those commits fix it (I was testing in IE11).

pwalczysko:
It works now on IE11 for me as well. It is quite slow, but it loads the table and I can select the ROIs and change the page.

jburel:
I think iviewer on IE is generally slower than on other browser

pwalczysko:
Loading of ROIs gets sometimes "stuck". this is not browser specific. I still have to find out a foolproof workflow.
Some combination of opening the svs image in multiview and then combinig it with the multi-t image from the same dataset, then clicking on the single windows whilst being on the ROI tab, will get into endless "Loading ROIs" state.

will-moore:
@pwalczysko Just make sure you have the developer tools open so you can see any error there (under Console tab). Also the Network tab will tell you whether the last call failed or returned nothing.

pwalczysko:
@will-moore I got stuck just now in FF Win 7, whilst trying to retrieve 
http://web-dev-merge.openmicroscopy.org/static/omero_iviewer/css/all.min.css?_iviewer-0.6.0


Under console tab, there is a message

``Use of mozImageSmoothingEnabled is deprecated. Please use the unprefixed imageSmoothingEnabled property instead.``

will-moore:
@pwalczysko I think those FF messages are not related to this issue (you can use the XHR filter in the Network tab to exclude css, images etc), but I managed to reproduce what you're seeing in Safari and added a fix.

pwalczysko:
@will-moore Great, thanks for finding out the workflow. Yes, now, in Safari and FF on Windows 7, I have no problems. 

The only issue I see is on IE, where , IE being much slower, when I go through the workflow described by you https://github.com/ome/omero-iviewer/pull/231#issuecomment-464942416 - then there is a long delay in switching between the windows. Up to the point that a message came "openmicroscopy is not responding due to a long-running script". But, in the end, the switch happens, and the ROI table loads.

will-moore:
That last commit fixes a 'silent bug' that causes additional loading of ROIs and might explain some of the long delays seen above. In the case of opening an image when the ROI tab is already open, the previous fix 42f5d68 ensures that we load ROIs immediately (as we did before this PR). However, we don't yet have the image data (including ```roi_count```) in hand at this point so the initial ROI loading doesn't use pagination (just loads first 500 ROIs). Once the image data has loaded, init of Z/T sliders (& setting of Z & T values) causes the ROIs to load again, this time using pagination for the current plane. This doesn't happen until after the first, slow loading of 500 ROIs is done.
Now, with 1e152ff we don't start loading ROIs until the image_data is loaded, so we avoid the wasted (and slow) call to load 500 ROIs.
Unfortunately, in cases where we don't need to paginate by plane, we are now waiting a bit longer before we start loading ROIs, but this can't be helped.

pwalczysko:
I tested on Win 7 IE. The performance is still very slow. It is very hard to say if it is faster than before, the ROIs on svs with multi-view load within 43 secs - but that is such long a time that it is hard to judge it as a speed-up.

I do not see any worsening of the ROI loading speed on FF Mac with ROIs around 200 (non-paginated)

jburel:
Merging this PR for the rc. In the long run, the table is likely to be removed. See discussion on the investigation PR https://github.com/ome/omero-iviewer/pull/242

PR #232, opened 2019-02-06
Fix centreOnGeometry for Point or Ellipse

will-moore:
Probably a bug that was introduced in the big OpenLayers-5 update.
To test:
 - In the ROI table, click on a Point and/or Ellipse that isn't visible in the viewer
 - The viewer should pan to show the shape in the centre


jburel:
This now works as expected

PR #233, opened 2019-02-06
drag and drop to open image

will-moore:
Drag and drop a thumbnail onto centre panel to open image.

To test:
 - Drag a thumbnail into the centre panel of iviewer to perform same behaviour as double-clicking the thumbnail. You will enter multi-viewer mode (if you are not already in it)
 - If you are in multi-viewer mode, you can drag onto an existing viewer to replace the image there OR drag over the outer frame to open a new image viewer. See demo in https://twitter.com/will_j_moore/status/1093349075720626177

NB: Found some potential issues with current double-click behaviour (need to review in another PR):
If you have unsaved ROIs and you double-click or drag to open a different image to the one currently open, you will get a "Save ROIs" dialog etc. But, if you double-click or drag to open the *same* image that is currently open then you don't get a dialog until after the image has opened, so you have 2 viewers of the same image: the original one with unsaved ROIs and the new one with saved ROIs. This seems wrong to me, but the code and use-cases are quite convoluted so I don't want to start tackling that in this PR. Need to define supported workflows.

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -738,7 +738,7 @@ export default class Context {
         let selImgConf = this.getSelectedImageConfig();
         let hasSameImageSelected =
             selImgConf && selImgConf.image_info.image_id === image_id;
-        // show dialogues for modified rois
+        // show dialogs for modified rois
         if (image_config &&
             image_config.regions_info &&
             (image_config.regions_info.hasBeenModified() ||
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -672,6 +672,31 @@ export default class Context {
     }
 
     /**
+     * Get the parent e.g. {type:'dataset', id:123} for example used to
+     * load additional thumbnails.
+     */
+    getParentTypeAndId() {
+        let image_config = this.getSelectedImageConfig();
+        // Dataset ID or Well ID or Image ID
+        let parent_id = null;
+        if (this.initial_type === INITIAL_TYPES.DATASET ||
+            this.initial_type === INITIAL_TYPES.WELL) {
+                parent_id = this.initial_ids[0];
+        } else if (this.initial_type === INITIAL_TYPES.IMAGES &&
+            this.initial_ids.length === 1 &&
+            image_config !== null &&
+            typeof image_config.image_info.parent_id === 'number'){
+                parent_id = image_config.image_info.parent_id;
+        }
+        let parent_type = INITIAL_TYPES.NONE;
+        if (parent_id) {
+            parent_type = this.initial_type === INITIAL_TYPES.IMAGES ?
+                    image_config.image_info.parent_type : this.initial_type;
+        }
+        return {type: parent_type, id: parent_id};
+    }
+
+    /**
      * Click Handler for single/double clicks to converge on:
      * Opens images in single and multi viewer mode
      *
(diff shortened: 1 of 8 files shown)


jburel:
Nice feature. This will improve the discoverability of the multiview option

src/app/context.js
```diff
@@ -684,20 +709,11 @@ export default class Context {
...
-            let parent_type =
-                parent_id === null ? INITIAL_TYPES.NONE :
-                    context.initial_type === INITIAL_TYPES.IMAGES ?
-                        image_config.image_info.parent_type :
-                        context.initial_type;
+            // Dataset ID or Well ID or Image ID
+            let parent = this.getParentTypeAndId();
+            let parent_id = parent.id;
+            let parent_type = parent.type;
+            console.log(parent_id, parent_type);
```
jburel: you may want to remove the console.log

jburel:
* I like the ability to replace the image in an existing viewer
* Dragging an image to a desired location is not completely solved.

pwalczysko:
Really nice on Mac OS X FF.
Will give it a go on Win 7 tomorrow.

snoopycrimecop:
Conflicting PR. Removed from build [iviewer-merge#953](https://ci.openmicroscopy.org/job/iviewer-merge/953/). See the [console output](https://ci.openmicroscopy.org/job/iviewer-merge/953/consoleText) for more details.
Possible conflicts:
  - PR #231 will-moore 'Roi table pagination'
    - src/app/context.js

--conflicts

pwalczysko:
Unfortunately
Win 7, FF:

   - D & D works only when dropping into an existing window. This means, it will never open a new window as a result of a D&D, unlike on Mac.

pwalczysko:
But, I have the same experience now on Mac -> D&D does not work - this was not the case - but something of this PR seems still to be there, as the pagination is still present - what is going on ?

will-moore:
Apologies @pwalczysko I think that this PR might have been excluded because of a merge conflict yesterday (hopefully resolved now).

pwalczysko:
So, after this PR is in the mix. 

Win 7

   - Chrome works as expected (and fast)
   - Firefox works as expected (little bit slower than chrome)
   - IE does not work with D&D at all, no reaction

pwalczysko:
Interestingly, I have noticed on windwos, that in multi-view mode, you can interact (zoom, pan) with a window which is not active. The click into the window does not make the window active (which is highlighted by the color of the top bar of the window). Only the click on the bar itself is changing the color of the bar (darker blue for an active window).

It seems like unexpected behaviour. Do we have a card for this @will-moore ?

will-moore:
@pwalczysko The zoom on a non-active window has always been possible I think, but the click on the image *should* activate that viewer window. It's not Windows-specific (I see it on Mac + Chrome) and it's not from this PR. Probably a bug from the Openlayers update. I'll have a look when I get a chance.

will-moore:
Sorry, I should have mentioned that this is not expected to work in IE: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItem#Browser_compatibility
But since this feature is a short-cut rather than an essential workflow I think this is OK, and IE is only about 5% of browser market now.

jburel:
The fact that it does not work onIE will have to be clearly stated in the announcement and changelog

will-moore:
Added it to the changelog in https://github.com/ome/omero-iviewer/pull/238

PR #234, opened 2019-02-06
Fix thumbnail panel scroll to selected image

will-moore:
To test:

 - Open an image within a Dataset with a bunch of images (so that the iviewer thumbnail panel has a reasonable scroll range)
 - Double-click to open multiple images in centre of iviewer
 - Scroll the thumbnail panel so that opened image(s) are no-longer visible
 - Select the image viewer - when selection changes, the thumbnail panel should scroll to show the selected image thumbnail.


jburel:
* Select the image in the centre panel (while in multi-view)
* Left-hand panel now scroll to the image if it is not visible

PR #242, opened 2019-02-21
Roi tiles

will-moore:
This uses Tile-based ROIs on single-plane images with > 500 ROIs.
On multi-plane images, if there are more than 500 ROIs on any plane, simple pagination is used.

Instead of the previous strategy #227 of replacing the existing Vector Regions layer with Tiled Regions layer, this tries to use both alongside each other. This means that we can still use the existing Regions layer for editing ROIs we have loaded in the ROI table (as in #231) but we can browse Tiled ROIs on the image.
If you click a Tile-based ROI on the image, it loads the ROI into the Table (and the Vector Regions layer) so you can edit them.

If we wanted to pursue this strategy, other TODOs:
 - Remove unused ```all_roi_ids```
 - Consider what to do for multi-plane images with > 500 ROIs per plane.
 - Need proper backend support for querying shapes by region. See https://trello.com/c/Tsv0yVe1/97-query-shapes-by-geometry

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -642,24 +642,16 @@ thumbnail-slider img {
     flex: 0 0 auto;
 }
 
-/* Only contains .regions-list2 */
+/* Contains .pagination-controls, .regions-header and .regions-table */
 .regions-list {
     flex: 1 1 auto;
     float: left;
     overflow: hidden;
     width: 100%;
-    display: flex;
-    flex-direction: column;
-    position: relative;
-}
-
-/* Contains .pagination-controls, .regions-header and .regions-table */
-.regions-list2 {
-    position: relative;
     height: 100%;
-    overflow-y: auto;
     display: flex;
     flex-direction: column;
+    position: relative;
 }
 
 .pagination-controls {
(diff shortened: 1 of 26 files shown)


will-moore:
Showed @jburel the ROI shape counts when zooming out:

screen shot 2019-02-21 at 13 49 54

Discussion:
 - Would be better not to load pages of ROIs into the table since it serves no real purpose (and is slow)
 - Instead, can we edit the shapes that are loaded by tile (or just load those shapes to Regions layer)?
 - The only use of the ROI table currently is:
    - Showing ROIs on other planes (when we are not paginating by Z/T plane) - click to browse to other plane
    - Showing Comments, Lengths, Areas for all Shapes - but not very useful unless we can sort, filter, search etc.
 - A better table for multi-Z/T images would show which planes the ROIs/Shapes are on (in the same way that zooming out 
 - Would be nice to have OMERO.parade - like features (filtering and sorting ROIs by various criteria)

will-moore:
Current status: for single-plane images with > 500 ROIs, we don't load ROIs in ROI table but instead load them via Tiles (TiledRegions layer) on the image. When a shape is clicked on the image, it is transferred to the Vector Regions layer for editing, and also it shows up in the ROI table. You can then edit and Save the shape, and draw new shapes:

screen shot 2019-02-25 at 14 02 24

PR #243, opened 2019-02-28
Fixes

jburel:
Fix version display in About dialog
cc @pwalczysko


sbesson:
@jburel worth releasing as `a2`?

jburel:
I don't think it is necessary

PR #244, opened 2019-02-28
Deploy

jburel:
Run from plugin directory
cc @sbesson


jburel:
We could try a2 with that change and #243

jburel:
Merging now that travis is green
Then I will push a tag

PR #245, opened 2019-03-06
Plane roi grid

will-moore:
This allows you to browse ROIs by plane (find which planes have most ROIs on them).
It implements most of what's on https://github.com/openmicroscopy/design/issues/91

To test:
 - Find a multi-plane image and make sure you have various numbers of Shapes on different planes, including some shapes 'unattached' to Z or T.
 - Under the ROIs tab, click on the "Planes" sub-tab.
 - The grid should show numbers of shapes that appear on each Z/T plane.
 - The 'Key' should show the range of Shape counts per plane
 - Tool-tip on each grid 'cell' should give you the count for that plane.
 - Clicking on the plane should take you to that Z/T position (and show the cell selected)
 - Shift-clicking on a Z-stack image should select a Z-projection range.
 - Single-click (no Shift) should turn off the Z-projection again.

Demo at https://twitter.com/will_j_moore/status/1103206476695236609

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -748,6 +748,7 @@ thumbnail-slider img {
     border: solid white 1px;
     text-align: center;
     overflow: visible;
+    user-select: none;
 }
 .grid_cell.selected {
     border: solid rgb(90, 135, 248) 3px;
--- a/css/app.css
+++ b/css/app.css
@@ -744,6 +744,9 @@ thumbnail-slider img {
     text-align: center;
     overflow: visible;
 }
+.grid_cell.selected {
+    border: solid rgb(90, 135, 248) 3px;
+}
 
 .y_axis, .x_axis {
     font-size: 10px;
--- a/css/app.css
+++ b/css/app.css
@@ -755,6 +755,9 @@ thumbnail-slider img {
 .x_axis {
     color: transparent;
 }
+.y_axis {
+    width: 25px
+}
 
 /* Show the first '1' x_axis label... */
 .col:nth-of-type(2) .x_axis {
--- a/css/app.css
+++ b/css/app.css
@@ -720,6 +720,44 @@ thumbnail-slider img {
     width: 100%;
 }
 
+/* Grid to show Shape counts per plane */
+.grid_table>div {
+    display: inline-block;
+    border-top: solid grey 1px;
+    border-bottom: solid grey 1px;
+    margin-bottom: 8px;
+}
+
+.grid_table>div:first-child {
+    border-left: solid grey 1px;
+}
+
+.grid_table>div:last-child {
+    border-right: solid grey 1px;
+}
+
+.grid_cell {
+    width: 15px;
+    height: 15px;
+    border: solid white 1px;
+    text-align: center;
+    overflow: visible;
+}
+
+.y_axis, .x_axis {
+    font-size: 10px;
+    border-color: transparent;
+}
+.x_axis {
+    color: white;
+}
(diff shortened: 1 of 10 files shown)


manics:
Screen Shot 2019-03-27 at 16 30 21-fullpage

This looks pretty cool! I tested it on an IDR image with 31T 40Z and the grid is wrapped at T=20, it would be better to keep the grid as a single rectangle and scroll horizontally

will-moore:
Fixed tooltip over each grid cell to show ```n Shapes T: t, Z; z```.

@manics The grid wraps as whatever the width of the right panel is. If you resize right panel you should see that. I didn't want users to have to scroll a lot to see all the planes in a grid. E.g. for an image with a lot of time-points and single-Z, this could be really painful.

Screen Shot 2019-04-01 at 14 39 32

As discussed with @jburel, this could probably do with a bit more testing on Images with different Z/T counts and ROI counts.

manics:
The problem in https://github.com/ome/omero-iviewer/pull/245#issuecomment-477241911 is there's a vertical scrollbar instead, and the second group of rows do not have a timne indicator.

If it's broken horizontally I think repeating the `T=N` column will improve the usability.

manics:
Another idea, in the `T=NNN` case could you break on a round number of `T`, e.g multiple of `5` or `10` ?

will-moore:
@manics Since I want the grid to resize dynamically when the panel width changes, this makes it hard to  add ```T: ``` for each row or to break on multiples of 5 or 10.
There may be some more advanced CSS techniques that might help. Will need to check...

manics:
@will-moore If it's not possible in CSS could you instead select the layout based on the size of T? E.g. If `t<=3` use the current layout with wrapping, otherwise don't wrap and instead scroll horizontally?

will-moore:
From @pwalczysko - need more space for plane indexes with large plane counts (> 100): https://trello.com/c/wJjfeoCf/80-bug-regression-css-on-iviewer-many-rois

will-moore:
Discussed layout issues above with @pwalczysko - preference is to try and reduce the need to scroll and make best use of space (don't leave a whole lot of unused space in the panel with a big scroll bar). Decision is to wrap the T-index at e.g. 20 so we can add the Z-index at each row. Also, if we have a large Z-count and only a single T, then wrap the Z instead at e.g. 50 (or based on size when it's loaded?), so that we can see all planes in the same view without having to scroll. cc @manics

will-moore:
Now we show grid like this for Timelapse (single Z) 

Screen Shot 2019-07-17 at 16 26 31

Timelapse with multiple-Z:
Screen Shot 2019-07-17 at 16 25 57

Z-stack (no Time-lapse)
Screen Shot 2019-07-17 at 16 21 59

manics:
Looks pretty good to me. A few UI issues that I mentioned to @pwalczysko but they're not blockers:
- the colour key indicating number of ROIs looks like it's clickable but it's not
- if you create and save a new ROI the number displayed in the top tab doesn't update

will-moore:
I added a 'pointer' cursor to the grid cells, to indicate they are clickable (and the key is not).
Not sure how to make the key look less clickable.
Updating the counts when you Save ROIs would be quite hard. Probably would need to reload them from the server instead but not in this PR.

manics:
Looks good.
The tooltip shows fractional numbers of ROIs:

Screen Shot 2019-08-05 at 11 47 55

But I can't think of a good solution

PR #246, opened 2019-03-06
fix 11 thumbnails in Dataset

will-moore:
To test:
 - Find or create a Dataset with 11 images (or 21, or 31 etc)
 - Open first image in iviewer
 - Scroll thumbnail slider - check that thumbs load correctly and the LAST thumbnail loads.


joshmoore:
Thanks, @will-moore. Do you have any idea how long this has been around? What tests would prevent this type of thing from sneaking through?

Issue #220, opened 2019-01-14
Show ROIs "Pins" in iViewer


c15mnw:
It would be very useful to provide a tool within the viewer to aid the discovery of ROIs.

This could work in 2 ways:

1. Proximity: Moving the mouse across an image should cause ROIs to be "highlighted", when the mouse is near to a ROI - a "pin" would then appear (a la Google Maps), allowing the user to click to it, to show the ROI in the list on the RHS.  This functionality could be switched on and off as a viewer preference.
(This functionality would help as Point ROIs, in particular, are very difficult to locate within the image)

2. Individual ROI: Clicking on a ROI on the RHS, should cause the ROI within the image to be indicated by producing a "Pin" on the image.

Similar functionality is used in the eMouseatlas Histology
( https://www.emouseatlas.org/eAtlasViewer_ema/application/ema/kaufman/plate_38c.php?image=b )
( https://www.emouseatlas.org/emap/about/what_is_eHistology.html )

will-moore:
These features (https://github.com/ome/omero-iviewer/pull/248) were included in the 0.8.0 release of iviewer 
https://www.openmicroscopy.org/2019/08/21/iviewer-0-8-0.html
(0.8.1 is the latest now).

PR #248, opened 2019-03-11
Popup on shapes

will-moore:
This addresses:
 - https://github.com/ome/omero-iviewer/issues/220
 - A request from @jrswedlow I think to allow editing of shape Label on the image itself (instead of the right panel).

This shows a transient 'tooltip-like' popup when you mouse over (or within 5 pixels) of a shape, to show the shape Label (IF it has one). This tracks the mouse.

Screen Shot 2019-03-11 at 09 19 11

When you click on the shape, a more stable popup shows you the shape coordinates and a text-box for editing the shape Label:

Screen Shot 2019-03-11 at 09 00 05

You also see this while dragging, editing and creating a shape, which can help if you want to know the coordinates. Could maybe also add the Length OR Area.

To test.... NB: since this is quite a big change, we probably want to assess this for a while before merging. Need to see whether the popup gets in the way at-all or is annoying to have it all the time. And what info should it show?
 - Draw a new shape - while drawing, the Edit popup should show above the mid-point of the shape (shouldn't get in the way of drawing) and the coordinates and Length/Area should update.
 - When shape has been drawn, you can edit the Label in the popup. Hit Enter to save the label.
 - NB: you can do this while still in Drawing mode and then draw another shape (without having to switch out of draw mode to edit shape label)
 - Mousing over shapes will show the label in a transient smaller "hover" tooltip that tracks the mouse pointer (IF the shape has a Label) and should increase the thickness of the stroke width to indicate which shape is being 'hovered'.
 - Check how the hover tooltip looks with very short and very long labels (e.g. big paragraph).
 - Clicking a shape to select it will hide the transient "hover" tooltip and show the Edit popup.
 - Undo / Redo or editing the shape label in the Right panel "Comment" box will update the label in the Popup, if visible.
 - Change of Z/T index should hide the popup if the current shape isn't visible on the new plane. Same for delete or hiding of shape.
 - For multiple selection, popup will show for most-recently selected shape.
 - If multiple shapes are selected, some can be individually de-selected without hiding the popup on another shape.
 - Do we want to add a 'X' button on the popup to hide it without de-selecting shape?


src/viewers/viewer/source/Regions.js
```diff
@@ -663,6 +679,23 @@ class Regions extends Vector {
         return true;
     }
 
+    /**
+     * Sets the ID of a shape that we are hovering over, to update it's style
```
mtbc: its

pwalczysko:
I think user would ask immediately: How do I switch that off ? It is nice that it is there, but maybe I do not want those popups as you rightly remark ?

pwalczysko:
I do not understand, once I clicked onto a ROI with intention to edit the popup, how do I get rid of it ? It stays there forever, whilst I am driving with my mouse over other ROIs already.
Do not see and Close button in the corner or similar.
Screen Shot 2019-03-18 at 17 18 25

pwalczysko:
Will test the smaller features tomorrow.

will-moore:
You can now close the popup via a X button on it.
Wasn't sure where to add an option to disable the popup. I added it to the context menu (right-click menu) on the image.

will-moore:
Last commit should fix https://trello.com/c/bK7Fm2uC/71-mismatch-between-big-image-overview-and-image-view

pwalczysko:
The close cross in top right of the popup is as expected.
The new menu item in the context menu to disable the popup is nice. 

The problem is that the smaller popups completely vanished, and I cannot see how to switch them on (by "smaller" I mean the ones which were appearing on non-selected ROIs mouseover).

pwalczysko:
I was wondering why is it not possible to copy and paste the information about the ROI such as positions and widths and lenghts from the iviewer and paste it locally ?
I do not mean export now. Users repeatedly expressed such wishes - we have two places from which to try it, one is the rightmost column of the ROI table, the other is now the "bigger" popup. Neither of them can be copied. Can that be done so that the copying would work ?

Screen Shot 2019-03-20 at 11 23 00
Screen Shot 2019-03-20 at 11 23 33

pwalczysko:
I think overall, but it struck me when drawing and immediately editing the labels - 

The silent assumption that the user knows to hit Enter to confirm the edit of the shape is not always good. I caught myself couple of times not hitting Enter and then wondering where is my label when coming back to the shape later.

pwalczysko:
Polyline and freehand polyline:

Those shapes do not update the end coorditnates correctly during drawing. Polyline pretends to be a single line with coordinates of the first segment only. Freehand polyline does not update as you draw,, the end coordinates, just the length.

pwalczysko:
Polygon and freehand polygon

These seem to update the coordinates correctly. The problem is only that the list of coordintates is so long that the last coordinates are not even displayed (and because of the Copy and Paste problem, see my previous comment, these numbers are not reachable for the users, hidden behind three dots in the tooltip)

pwalczysko:
The TEXT shape exposes the "reusing" in the UI - I cannot have another comment on the Text shape than the Text itself - feels strange
Screen Shot 2019-03-20 at 12 07 13

pwalczysko:
Couple of points to discuss and fix.

pwalczysko:
Re https://github.com/ome/omero-iviewer/pull/248#issuecomment-474421432 - why is this in the present PR dealing with ROI tooltips ? This seems to me to be a very different topic ?

pwalczysko:
The https://github.com/ome/omero-iviewer/pull/248#issuecomment-474421432 does not seem to have been fixed at all to me.

will-moore:
Addressed some points above:

 - You don't need to hit Enter to save text now. It's saved when you stop or pause typing (after 0.5 secs). Is this delay about right?
 - When you select a Label, the text is now displayed correctly in the popup, allowing you to edit it.
 - When multiple shapes are selected at once, we don't show the popup. But if multiple shapes are selected and you select 1 more (e.g. with Ctrl-Click on shapes table or Shift-click on the image) the popup will show for that shape.
 - The PolyLine coordinates are now shown as "points:.. " instead of x1, y1, x2, y2. 
 - You can now select and Copy the coordinates (and Area/Length), including the whole Points string. 

Screen Shot 2019-04-03 at 14 31 12

pwalczysko:
the bird's eye view bug https://github.com/ome/omero-iviewer/pull/248#issuecomment-474803320 is fixed now.

The new features listed in https://github.com/ome/omero-iviewer/pull/248#issuecomment-479490270 work as described. Quite happy about them.

The only point is the position of the popup with respect to a ROI in cases when the ROIs are lines or arrows pointing  at 45 degrees angle. In such cases the popup is too far from the ROI. Discussed a solution with @will-moore in the office, where for lines and arrows the position of the popup could be at the topmost end of the ROI, instead of using the centre of the top of bounding box (used for all shapes now)

jburel:
Entering text work and the right-hand panel is updated. Pressing enter after entering text should close the dialog.

jburel:
@pwalczysko do you want to over the last set of changes?

jburel:
Having the dialog popping up every time a shape is selected is a bit annoying.
A general control should be available to turn it on/off

pwalczysko:
@jburel discussed the solution of https://github.com/ome/omero-iviewer/pull/248#issuecomment-511821238 in the office - I guess @will-moore will come up with a solution commit soon.

Otherwise, all is good here, and this is ready to go FMPOV

will-moore:
@jburel - you can turn off the popups with a right-click menu on the image. But as discussed with @pwalczysko this isn't obvious to find it.
We decided that when you click ```X``` to hide a popup, they will say "To turn on the pop-up again, right-click on the image and choose 'show popups" and then we don't show them again till they do that.
I seem to remember the context menu can't get access to the current "show popups" flag, so I'll have to see if there's a way to handle this...

jburel:
I certainly did not think to right click on the image to turn the dialog on/off.

will-moore:
As @pwalczysko noticed on images with many ROIs (where they are paginated). After moving to a new plane, the shapes don't appear selected in the table, clicking on shapes doesn't select on first click, picking line colour doesn't show up on shape.

will-moore:
To test last fixes:
 - When you click ```X``` to hide the popup, you get a message saying how to turn them on again.
 - Fixed the issues with many paginated ROIs. After switching Z/T, shape selection, picking colors etc should continue to work.

Known issue (I think this is not related to this PR): If your line color-picker is showing a color (e.g. red) and you select a shape where the line has no color. Then you click the color-picker and don't choose a different color but just click "choose" to select red, the shape doesn't update because no color change is detected.

pwalczysko:
Screen Shot 2019-07-19 at 14 03 27

Maybe ``...To re-enable popups, right-click on the image and use the context menu.``

pwalczysko:
RFE: If not technically too difficult: Slightly unexpected is that when I re-enable the popups, the ROI which is selected already, does not obtain a popup. Only new selection do. I think better would be the behaviour where the popups would come immediately back after I re-enabled them.

pwalczysko:
Two RFEs otherwise looks good.

pwalczysko:

jburel:
The popup now gives an indication on how to re-activate it, pointing at the right-click menu.

PR #249, opened 2019-03-14
Fix ROI loading on play movie

will-moore:
Added a missing check here to only clear the ROIs if we are loading the ROIs by plane (for images with > 500 ROIs).
I already had this check for re-loading ROIs when the movie stopped, which is why they never re-load.

To test:
 - Find a multi-T image with < 500 ROIs on various planes (and/or 'unattached' to T) e.g. http://web-dev-merge.openmicroscopy.org/webclient/img_detail/9469/ (user-3)
 - Play the movie
 - ROIs should remain loaded and update on each change of T (as before)


jburel:
It now works as expected and it is possible to view the tracks generated by trackmate
cc @manics

PR #250, opened 2019-03-14
Fix get_intensity permissions

will-moore:
See https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8698

To test:

 - Choose another user's image in your non-default group and try to query pixel intensity.


pwalczysko:
Tested on eel (could repeat two situations on outreach, user views data in their non-default group of other users, and admin views data in their non-default group of other users).
All works as expected now

Ready to merge FMPOV

jburel:
@pwalczysko did you check with public user too?

pwalczysko:
@jburel no, but tried this additional test just now: add public user to read-only-1, this is public user's non-default group. Then view image of another user as public user in read-only-1. All works fine.

Issue #252, opened 2019-03-15
smart scale bar


jburel:
From https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8704
Currently, the scale bar is not cool.

The length of the bar stays the same, and instead the number associated with the scale bar changes, often taking a weird number, like, 0.476 mm.

So far the feedback I received about this scale bar is not great.

People prefer to see a scale bar with more sensible numbers, eg. 0.500 mm or 200 µm etc.

Could it be possible to write code to find the right length of the scale bar, eg, 100, 200, 500, 1000, 2000, 5000,...?

This might take a few milliseconds away, but the user experience might be much better.

jburel:
See https://github.com/ome/omero-iviewer/pull/253

PR #253, opened 2019-03-19
Scale bar lengths

will-moore:
See https://github.com/ome/omero-iviewer/issues/252

Now we calculate the length of the scalebar to show a nice rounded number.

To test:
 - find / create images with various pixel sizes (see first 4 images in Dataset http://web-dev-merge.openmicroscopy.org/webclient/?show=dataset-27801 user-3)
 - zoom in & out and check that scalebar is accurate and shows a nice round number.

E.g, zooming in looks something like this:

Screen Shot 2019-03-19 at 04 08 15

Screen Shot 2019-03-19 at 04 08 31

Screen Shot 2019-03-19 at 04 08 50

Screen Shot 2019-03-19 at 04 08 57

Screen Shot 2019-03-19 at 04 09 08

Screen Shot 2019-03-19 at 04 09 15

### DIFF
--- a/src/viewers/viewer/controls/ScaleBar.js
+++ b/src/viewers/viewer/controls/ScaleBar.js
@@ -36,11 +36,11 @@ class ScaleBar extends ScaleLine {
         super(opt_options);
 
         /**
-         * default scale bar width in pixels
+         * minimum scale bar width in pixels
          * @type {number}
          * @private
          */
-        this.bar_width_ = 100;
+        this.bar_min_width_ = 100;
 
         /**
          * the scalebar 'drag' listener
@@ -102,7 +102,9 @@ class ScaleBar extends ScaleLine {
 
         var micronsPerPixel = viewState.projection.getMetersPerUnit();
         var resolution = viewState.resolution;
-        var scaleBarLengthInUnits = micronsPerPixel * this.bar_width_ * resolution;
+
+        // first find the Units and Length for min-length scalebar
+        var scaleBarLengthInUnits = micronsPerPixel * this.bar_min_width_ * resolution;
         var symbol = '\u00B5m';
         for (var u=0;u<UNITS_LENGTH.length;u++) {
             var unit = UNITS_LENGTH[u];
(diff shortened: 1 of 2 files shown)


src/viewers/viewer/controls/ScaleBar.js
```diff
@@ -113,16 +115,21 @@ class ScaleBar extends ScaleLine {
             }
         }
 
-        var html = scaleBarLengthInUnits.toFixed(2) + ' ' + symbol;
-        if (this.renderedHTML_ !== html) {
-            this.innerElement_.innerHTML = html;
-            this.renderedHTML_ = html;
+        // Find a length value BIGGER than our min target length
+        let snapToLengths = [1, 2, 5, 10, 20, 50, 100, 200, 500, 1000, 2000, 5000, 10000];
```
rgozim: Could this be made a "constant" (loose term for javascript)

rgozim:
Added a single comment to the code, but otherwise works fine. Tested on firefox, safari and chrome.

src/viewers/viewer/globals.js
```diff
@@ -222,7 +222,7 @@ export const PROJECTION = {
  */
 export const UNITS_LENGTH = [
     { unit: 'angstrom',
-      threshold: 0.1, multiplier: 10000, symbol: '\u212B'},
+      threshold: 0.01, multiplier: 10000, symbol: '\u212B'},
```
sbesson: What is the impact of modifying this threshold?

will-moore: If, for example, the minimum length of scalebar was ```0.8 microns```, then we look for a UNIT where the threshold is larger than this (starting at the smallest, ```angstrom```). With the old value of ```0.1```, we pick ```angstrom``` as a suitable unit for a scalebar of ```0.8 microns```. When the scalebar is increased from ```800 angstrom``` to a rounded length, we end up with a length of ```1000 angstrom```.
With the change in threshold, we instead end up picking ```nm``` and we get a scalebar of ```100 nm``` which seems more appropriate.

sbesson: Ah I see, so anything below `0.01 microns` will be turned into angstroms rather than  nanometers. Sounds like a fair cutoff and certainly better than `100 nm`. We can review this and further push towards `1nm` if needed.

kouichi-c-nakamura:
Thanks a lot guys for implementing this! People are going to love this (though they may just assume that it has been like this)!

kouichi-c-nakamura:
We have finally updated iviewer for our CHAMBER server and delighted to see the improved scale bar in action. Thanks, guys!

http://chamber.mrc.ox.ac.uk/iviewer/?images=120&dataset=51

PR #255, opened 2019-04-01
Clear cached image settings on save all

will-moore:
See https://trello.com/c/zdbn7Fhh/72-cached-rnd-and-save-to-all

To test:
 - Find a Dataset that has some "compatible" images where Save-to-all works, but also some that are not compatible.
 - Browse through the images in iviewer, so that we cache settings for them.
 - Save to All so that multiple images are updated
 - The cache should be cleared for all "compatible" images where the Save to All was applied. The newly saved rendering settings should be shown for these images
 - For the non-compatible images, the cache should not be cleared. They should still display the last-viewed settings.

cc @pwalczysko

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -945,6 +945,15 @@ export default class Context {
     }
 
     /**
+     * Clears cache of image settings for all images
+     *
+     * @memberof Context
+     */
+    clearCachedImageSettings() {
+        this.cached_image_settings = {};
+    }
+
+    /**
      * Convenience or short hand way of publishing via the internal eventbus.
      * It will just delegate whatever you hand it as arguments
      *
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -945,12 +945,22 @@ export default class Context {
     }
 
     /**
-     * Clears cache of image settings for all images
+     * Clears cache of image settings, specified by ID.
+     * If imageIds is not defined, this clears ALL cached images settings.
      *
+     * @param {Array.<number>} imageIds the IDs of Images to clear from cache
      * @memberof Context
      */
-    clearCachedImageSettings() {
-        this.cached_image_settings = {};
+    clearCachedImageSettings(imageIds) {
+        if (Misc.isArray(imageIds)) {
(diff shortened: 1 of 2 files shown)


pwalczysko:
Save to All is still completely broken on eel. 

Verified that the images in the dataset http://web-dev-merge.openmicroscopy.org/webclient/?show=dataset-18601 of user-3 are compatible by copying and pasting the rnd settings from one image to another. But, Save to All still have no reaction. Even as the server was restarted by @mtbc this morning (thank you).

pwalczysko:
Today, the Save to All on eel simply works. I tried all the workflows from yesterday, but cannot repeat the bug.

pwalczysko:
A bug, but  not a regression due to this PR (behaviour confirmed on nightshade). 

 - 1. Have several images in a dataset
 - 2. Apply on one channel of one of the images a LUT and save the rnd on that image
 - 3. go back to a simple color on that image you just saved the rnd on, but this time, do not save the change
 - 4. click onto another thumb in iviewer
 - 5. click on the thumb of the image you just manipulated -> observe that the LUT is in the main viewer (not expected), in the controls in RHP there is the simple colco (as expected)
 - 6. save the rnd on that image -> this does not fix the problem though. The thumbs suggest that the image is saved with simple color (as intended), but the main viewer still shows LUT

Screenshot: Note that the main viewer is on the LUT, but the controls in RHP show Red in the corresponding channel (situation after step 5 of the scenario above). 
Screen Shot 2019-04-03 at 12 25 13

Screenshot: Note that the main viewer is still on the LUT after step 6 above, although the Save was clicked and the thumbs are Red as expected

Screen Shot 2019-04-03 at 12 34 52

pwalczysko:
No problems with this PR, behaves as expected.

pwalczysko:
For the bug in LUT, https://github.com/ome/omero-iviewer/pull/255#issuecomment-479451497 created https://trello.com/c/tT1jx0Gf/75-bug-lut-and-cached-settings

PR #257, opened 2019-04-01
Export roi coords

will-moore:
See https://trello.com/c/zp3us1k4/64-export-to-excel-additions
This exports shape Text and coordinates in the CSV file.

To test:
 - Create or find a bunch of Shapes, at least one of each kind. Some with label / comment added. Save them.
 - Select various shapes and ROI -> Export 
 - CSV file should contain additional columns for ```Text``` as well as any other columns required for coordinates of selected shapes. E.g. if you only select Rectangles, you will see ```X, Y, Width, Height``` columns.

### DIFF
--- a/src/app/header.js
+++ b/src/app/header.js
@@ -345,7 +345,9 @@ export class Header {
                         }, {});
         coordHeaders = coordHeaders.filter(attr => shapeAttrs[attr]);
 
-        // Add to other column names
+        // Export Comment / Text for all shapes
+        coordHeaders.unshift("Text");
+        // Add coords to other column names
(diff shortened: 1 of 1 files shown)


jburel:
Text and coordinates are added to the CSV file.

mtbc:
:+1: Seems to be working for well me too (Firefox on Linux) once @will-moore helped bypass my tunnel-vision on the right-pane ROI tab looking for the export button and explained I have to select the ROIs first. :smiley: Affine transforms are *not* taken into account but that seems to be a general problem with iviewer beyond this PR.

PR #260, opened 2019-04-03
Use size 96 thumbnail for BirdsEye

will-moore:
See https://trello.com/c/E62b6Qsc/51-bug-birds-eye-view

This uses the default size 96 thumbnail for the Birds Eye view which means we can use the cached thumbnail which is much faster, particularly for big images. This was the behaviour of the old viewer.

To test:
 - Check that the Birds-Eye works for image in the trello card above.

### DIFF
--- a/src/viewers/viewer/controls/BirdsEye.js
+++ b/src/viewers/viewer/controls/BirdsEye.js
@@ -111,7 +111,7 @@ class BirdsEye extends Control {
         * @type {Array.<number>}
         * @private
         */
-        this.thumbnail_size_ = [96, 96];
+        this.thumbnail_size_ = [150, 100];
         var fullSidesRatio = this.full_image_size_[0] / this.full_image_size_[1];
         var tmp = [
             this.thumbnail_size_[0],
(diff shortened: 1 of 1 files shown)


jburel:
Instead of using hard coded value the default size should be retrieved using 
``omero.client.browser.thumb_default_size``

will-moore:
I realised that where I had hard-coded ```96``` in a previous commit wasn't the thumb size but rather the size of the viewport itself. I reverted that (to 150, 100) and the thumbnail now fills the BirdsEye better:

Before:
Screen Shot 2019-04-12 at 15 53 59
After:
Screen Shot 2019-04-12 at 15 54 25

mtbc:
I guess that it is unrelated to this PR that when as `user-2` I go to http://web-dev-merge.openmicroscopy.org/webclient/?show=image-157758 then iviewer shows me a very pink image with a rather gray bird's-eye view: the "classic" viewer does likewise.

will-moore:
@mtbc There's something strange about that image. When I query the pixel intensities on a background region of the image, the green channel is about half the Red and Blue. E.g. 120 vv 255. So it looks pink (and the thumbnail becomes pink when settings are saved). I don't know if this "bug" wasn't there when the thumbnail first got created? I don't see the same issue on other copies of the same image, e.g. http://web-dev-merge.openmicroscopy.org/webclient/?show=image-11
Strange, but not related to this PR.

pwalczysko:
As established today, this would be nice to have - another image was explored and earmarked for a demo which suffers from this problem on demo.op...py.org

PR #263, opened 2019-04-18
ellipse area

will-moore:
This improves the accuracy of the Area reported for an Ellipse, using ```PI * rx * ry``` instead of relying on OpenLayers which treats the Ellipse like a Polygon.
It should fix the discrepancy between the Polygon area seen in iviewer vv exported in Batch_ROI_Export.py script, as reported https://github.com/ome/scripts/pull/147#issuecomment-483335252

### DIFF
--- a/src/viewers/viewer/geom/Ellipse.js
+++ b/src/viewers/viewer/geom/Ellipse.js
@@ -217,6 +217,16 @@ class Ellipse extends Polygon {
     }
 
     /**
+     * Returns the Area as Pi * rx * ry instead of relying on Polygon superclass
+     * method which is less accurate.
+     *
+     * @return {number} the area of the ellipse
+     */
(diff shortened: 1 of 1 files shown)


pwalczysko:
I guess this is on west-ci, not eel ? Help anybody ?

jburel:
It is on eel for web apps

jburel:
on merge-ci, user-1

* Image with ID 1132 has one ellipse
* omero-script:  area 0.0815989624017
* iviewer: area 0.082 when using "export ROI" option or in comment editor dialog

will-moore:
This PR doesn't affect the rounding, just the accuracy (which is fixed in the comment above).
So I'm assuming this is good to merge?

jburel:
Yes. We may want to round in the script.

PR #264, opened 2019-04-24
Rois sorted

will-moore:
See https://trello.com/c/XZRTvaFj/9-sorting-of-rois-in-table

Allows sorting of ROIs in the table. Work in Progress

 - Disable if we have pagination (lots of ROIs)
 - Allow sort on theC, shape type, Length, Area?

When sorting ROIs by an attribute of Shape (e.g. label, theZ, theT etc) it's possible for an ROI to have many shapes. In this case, we pick the first shape!

Screen Shot 2019-04-24 at 11 05 11


jburel:
What is the status of this PR? still WIP?

will-moore:
I think this is ready to test. I'll list it for testing...

jburel:
It is now possible to sort by Z/T and comment
We could look add adding
When adding "area" to the display, part of the bracket is truncated see screenshot
sorting_iviewer

will-moore:
Now we support sorting by Area & Length:
If a shape has no Comment, Area or Length it should be sorted *after* those that do:

Screen Shot 2019-08-06 at 16 07 23

pwalczysko:

src/regions/regions-list.js
```diff
@@ -117,6 +124,17 @@ export default class RegionsList extends EventSubscriber {
         this.waitForRegionsInfoReady();
     }
 
+    sort(value) {
+        console.log('click Sort', value);
```
jburel: Do you want to keep it?

jburel:
Merging travis is green

PR #268, opened 2019-06-04
Show roi.Name in list if set

manics:
`omero.model.RoiI` has a name field. This is currently ignored by iviewer, here it is displayed in the `Comment` column. In this example the ROIs are named `"1"`, `"2"`, `"3"`, ...

Screen Shot 2019-06-04 at 15 35 24-fullpage

### DIFF
--- a/src/model/regions_info.js
+++ b/src/model/regions_info.js
@@ -408,6 +408,7 @@ export default class RegionsInfo  {
             for (let r in data) {
                 let shapes = new Map();
                 let roi = data[r];
+                let name = data[r]['Name'] || '';
                 // add shapes
                 if (Misc.isArray(roi.shapes) && roi.shapes.length > 0) {
                     let roiId = roi['@id'];
@@ -437,6 +438,7 @@ export default class RegionsInfo  {
                         count++;
                     }
                     this.data.set(roiId, {
+                        name: name,
                         shapes: shapes,
                         show: false,
                         deleted: 0
--- a/src/regions/regions-list.html
+++ b/src/regions/regions-list.html
@@ -123,8 +123,8 @@
                         <div class="regions-table-col shape-z"></div>
                         <div class="regions-table-col shape-t"></div>
                         <div class="regions-table-col shape-c"></div>
-                        <div class="regions-table-col shape-comment"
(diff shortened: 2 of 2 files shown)


sbesson:
If I understand correctly the proposed behavior for the `Comment` column is to display either `ROI.name` or `Shape.text` depending on the selected element if the property is not empty?

Overall this feels like a good way to use the existing UI and expose more metadata. Leaving @ome/omero-apps to review this functionally. If that's acceptable, proposing to cut a patch release of 
`omero-iviewer` and bump in the next IDR release?

manics:
AFAIK there's no such thing as a `Comment` on anything- for shapes the `Comment` is actually the `.text` field. ROIs does not have a `.text` field.

manics:
I can't see how to make it editable.

manics:
This is particularly useful when there are too many ROIs and the per-plane UI takes effect, as the ROIs are no longer consecutive.
Screen Shot 2019-06-05 at 16 29 24-fullpage

will-moore:
This looks fine and if it serves your needs then I think it is good to merge.
We should probably make this editable in the future but that might take some re-design since you can't currently select an ROI in the table (only shapes are selectable).
I noticed that you don't see the ROI name if it only has 1 shape, but presumably then you'd see the shape text and that is enough (but again, this would need more thought if we wanted to allow editing of the ROI name).
It looks like there is a ```description``` field for ROIs in the DB. This is probably what @jburel was thinking of for "comment". But I don't think we need to support that at this point.
So, good to merge.

sbesson:
If `description` was a requirement for the first phase, something like `let name = data[r]['Name'] || data[r]['Description'] |  '';` might be an acceptable compromise given that there is only one text column in the current UI. 

Deferring to @jburel on this call. I also vote for releasing this fix as `0.7.1` and start using it in our IDR use cases with ROI-level metadata.

jburel:
The fields have not been really used. We can revisit it later if there is a need for support of description.
Let's merge and release a 0.7.1

sbesson:
@manics now released as https://pypi.org/project/omero-iviewer/0.7.1/

Issue #270, opened 2019-07-30
wrong group context in opening well's images


mtbc:
With our normal test user setup:

1. Identify a plate in `read-annotate-1` owned by `user-5`.
1. Log into OMERO.web as `user-3`.
1. Select the thumbnail of one of the plate's wells (top-center pane).
1. Use the "Open with..." icon to open in OMERO.iviewer.
1. FAIL: See "No Image Data".

mtbc:
Damir reports this bug also occurs with one's own data outside one's default group.

will-moore:
Fixed in https://github.com/ome/omero-iviewer/pull/271

mtbc:
This isn't working for me at all in Firefox or Chromium, I just get a blank tab (with a correct title) and `TypeError` stuff in the JavaScript console.

PR #271, opened 2019-07-31
Use conn.SERVICE_OPTS in well_images()

will-moore:
See https://github.com/ome/omero-iviewer/issues/270

To test:
 - Browse to a Well in your non-default group
 - Open-with iviewer. URL is e.g. ```iviewer/?well=101```
 - Should see images in the well etc instead of "No image data"
 - Also can confirm directly that ```/iviewer/well_images/?id=101``` shows the correct response.

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -482,7 +482,8 @@ def well_images(request, conn=None, **kwargs):
         # get total count first
         count = query_service.projection(
             "select count(distinct ws.id) from WellSample ws " +
-            "where ws.well.id = :well_id", params, )
+            "where ws.well.id = :well_id", params,
+            conn.SERVICE_OPTS)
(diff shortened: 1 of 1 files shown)


mtbc:
This isn't working for me at all in Firefox or Chromium, I just get a blank tab (with a correct title) and `TypeError` stuff in the JavaScript console.

will-moore:
Something else is broken - need to investigate when I'm back.
But I think this PR is working: e.g. user-3 for well https://merge-ci.openmicroscopy.org/web/webclient/?show=well-11014 - this works: https://merge-ci.openmicroscopy.org/web/iviewer/well_images/?id=11014

will-moore:
The error should be fixed with https://github.com/ome/omero-iviewer/pull/264/commits/7e894df503fe8d79c7f850c3ef061f7583462fa1 in #264.

PR #273, opened 2019-08-15
fix draw shape when paginated

will-moore:
After scrolling Z/T with paginated ROIs, drawing a new shape fails since regions.viewer_ is null.
A similar fix was in 5ff0845e53

To test:
 - Open an image in iviewer with many ROIs e.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=image-39657 (idr)
 - Scroll to new Z/T plane
 - Try to draw a shape e.g. Rectangle, Polyline etc.
 - Should work normally


pwalczysko:

PR #276, opened 2019-08-22
Disable autoPan for shape popup

will-moore:
This fixes a minor bug where the image pans when drawing shapes because of the popup appearing for a new shape. The shape popup should never cause the image to Pan.
To test:
 - Select to draw a Shape E.g. Polygon
 - Click on the image near the top of the viewport to start drawing
 - The shape popup will appear and track the Polygon as you draw it, but it should not cause the image to pan. NB: this was noticeably worse when using Shift-draw to draw freehand.
 - Also, clicking to select existing shapes when they are close to the edge of the viewport should also not cause the image to pan.
  - In multi-viewer mode, popups and tooltip should work properly on both viewers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8359,6 +8359,12 @@
         }
       }
     },
+    "reflect-metadata": {
+      "version": "0.1.13",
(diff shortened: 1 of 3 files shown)


pwalczysko:

PR #277, opened 2019-08-27
roi outside image csv fix

will-moore:
This fixes a bug in CSV export of ROIs when an ROI is outside the limits of the image.

To test:
 - create a bunch of ROIs, 1 or more should be outside the limits of the image
 - Export to CSV.
 - Check that the ROI that comes after the outside ROI is on a new line and the CSV is all formatted correctly.


jburel:
The key point to reproduce the bug is not to have the shape outside the image been the last selected one
Tested against latest and merge.
latest:
latest_iviewer
merge:
merge_iviewer

Bug is fixed

PR #278, opened 2019-09-08
show hide shapes in ROI

will-moore:
See https://trello.com/c/DtbhtoAA/81-show-hide-all-shapes-in-an-roi
from https://forum.image.sc/t/large-mask-roi-on-large-image/29200/7

To test:
 - Find an image with multiple Shapes per ROI (or create in Insight)
 - Check ROI checkbox to show/hide all shapes in the ROI
 - Toggling individual Shape visibility should 'check' the ROI checkbox if All shapes in the ROI are checked (same behaviour as the "Show" checkbox at the top of the table).
 - The "Show" checkbox at the top of the table should also continue to work as expected.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -966,6 +966,10 @@ thumbnail-slider img {
     -webkit-user-select: none;
 }
 
+.regions-table-row input[type='checkbox'] {
+    margin: 0;
+}
+
 .regions-table-col {
     float: left;
     min-width: 10px;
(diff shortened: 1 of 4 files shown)


mtbc:
Seems odd to me that when I first go to the ROI tab for multi-shape ROIs the "show"s for the shapes are all checked but the "show" for their ROI isn't. (If I uncheck a shape then check it again then the ROI is then checked too.)

mtbc:
Otherwise seems to be working well for me!

will-moore:
https://github.com/ome/omero-iviewer/pull/278#issuecomment-529899413 should be fixed (ROI checkbox is initially checked)

PR #279, opened 2019-09-09
Fix multi-viewer save rois

will-moore:
To test:

 - Open the same image in 2 viewers at the same time
 - Draw some ROIs in one, don't save.
 - Select the other viewer.
 - Dialog says "Save?"  -> Yes
 - Other viewer should load the newly saved ROIs
 - Also, in multi-viewer mode, editing and saving ROIs should work normally (ROIs updated from unsaved to saved state in the table)

### DIFF
--- a/src/viewers/viewer/Viewer.js
+++ b/src/viewers/viewer/Viewer.js
@@ -2302,7 +2302,7 @@ class Viewer extends OlObject {
             }
         };
 
-        ret = [];
+        var ret = [];
         for (var r in rois)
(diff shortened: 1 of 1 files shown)


pwalczysko:
merge-ci server is not usable now because of https://openmicroscopy.slack.com/archives/C0K5EED3R/p1568108686329900 - will wait with testing here until this is resolved.

pwalczysko:
Actually, the iviewer works fine on merge-ci. The functionality here is as expected.

Ready to merge FMPOV

PR #281, opened 2019-09-20
Projected image

jburel:
unify description of projected image
See https://github.com/ome/omero-insight/pull/78
cc @will-moore


will-moore:
Looks good, 👍

PR #302, opened 2020-01-29
Have made cross origin login fix

### LINKED ISSUE #283: haveMadeCrossOriginLogin_ in URL
See https://forum.image.sc/t/omero-iviewer-a-strage-url-change-when-clicked-on-a-thumbnail/30260

haveMadeCrossOriginLogin_ is a flag used when running the development server and shouldn't be seen on released iviewer.

will-moore:
Fixes #283 

To test:
 - Use Open-With to open iviewer - this gives URL like https://merge-ci.openmicroscopy.org/web/iviewer/?images=18189
 - Click on another thumbnail to show another Image. The URL updates
 - The URL should not contain the ```haveMadeCrossOriginLogin_``` flag

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -38,21 +38,7 @@ require('../css/app.css');
 Bluebird.config({ warnings: { wForgottenReturn: false } });
 window['encoding-indexes'] = {"windows-1252": WINDOWS_1252};
 
-/* IMPORTANT:
- * we have to set the public path here to include any potential prefix
- * has to happen before the bootstrap!
- */
 let req = window.INITIAL_REQUEST_PARAMS || {};
-let is_dev_server = false;
-// #if process.env.NODE_ENV === 'dev-server'
-is_dev_server = true;
-// #endif
-if (!is_dev_server) {
-    let prefix =
-        typeof req[URI_PREFIX] === 'string' ?
-            Misc.prepareURI(req[URI_PREFIX]) : "";
-    __webpack_public_path__ = prefix + '/static/' + PLUGIN_NAME + '/';
-}
 
 /**
  * IViewer bootstrap function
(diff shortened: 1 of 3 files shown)


pwalczysko:
Looks good to me. 
Ready to merge FMPOV

src/main.js
```diff
@@ -38,21 +38,7 @@ require('../css/app.css');
...
 let req = window.INITIAL_REQUEST_PARAMS || {};
-let is_dev_server = false;
-// #if process.env.NODE_ENV === 'dev-server'
-is_dev_server = true;
-// #endif
-if (!is_dev_server) {
-    let prefix =
-        typeof req[URI_PREFIX] === 'string' ?
-            Misc.prepareURI(req[URI_PREFIX]) : "";
-    __webpack_public_path__ = prefix + '/static/' + PLUGIN_NAME + '/';
```
joshmoore: In general, it seems like you've stripped things out regarding for when `dev_server` is true which seems safe enough, but isn't this clause for the _non_-`dev_server` case?

will-moore: I don't fully understand all of the use cases here, but as far as I can see, this code was hard-coded NOT to run (since ```is_dev_server = true;``` came right before that if).
So I don't know when that code has ever run.
It's only job is to set ```__webpack_public_path__``` .Webpack dev server isn't used in production. Webpack is used to package the JS bundle, but we've been doing that a lot without this code, so I can only assume this isn't needed in our current workflows.

joshmoore: _Urgh_. Thanks for the cleanup :+1:

joshmoore: see https://github.com/ome/omero-iviewer/pull/142 for the introduction of this.

joshmoore:
Merging and leaving someone else to suggest release.

PR #305, opened 2020-02-17
confine projection

will-moore:
See https://github.com/ome/omero-web/pull/115

This ports the same logic to the iviewer to restrict the size of projections.

To test:
 - import an image where ```Z * X * Y * C * bits-per-pixel``` is > 1024 * 1024 * 256.
 - Z projection should be disabled.


will-moore:
That second commit handles the scenario at https://github.com/ome/omero-web/pull/115#issuecomment-586977502 where we enable/disable projection when user turns channels on/off.
The decision for old viewer was to simply use the total number of channels when deciding whether to disable projection (see discussion at that link). The above commit adds quite a lot of code to maintain and test, to handle the case of "image with a large number of channels is above the projection limit but we want to allow projection when a small number of channels are enabled". So we should probably keep things simple and revert that to same behaviour as old viewer)?

pwalczysko:
> The above commit adds quite a lot of code to maintain and test, to handle the case of "image with a large number of channels is above the projection limit but we want to allow projection when a small number of channels are enabled". So we should probably keep things simple and revert that to same behaviour as old viewer)?

Yes, I think so. Keep simple

pwalczysko:
Interesting side-question: Multi-t and multi-z images. Atm, I am gettting a multi-t projections when I have multi-t images with (orginally) multi-z. See for example user-3 https://merge-ci.openmicroscopy.org/web/webclient/?show=image-100718

pwalczysko:
Yes, the behaviour here is as expected. See the under-size-limit images from previous comment and a over-limit-image-when-all-channels-on in https://merge-ci.openmicroscopy.org/web/webclient/?show=image-100727

But I guess it would be better to not to introduce complex logic with "when channels on..." etc.

jburel:
I think it is better to stick to sizeC instead of active channels. 
This could potentially annoy user but it is easier to understand when the functionality is on/off

pwalczysko:
All looks good. Checked iviewer, and also "old" viewer with images from https://merge-ci.openmicroscopy.org/web/webclient/?show=project-15107

Ready to merge fmpov

src/controls/dimension-slider.js
```diff
@@ -517,7 +517,13 @@ export default class DimensionSlider {
      */
     getZProjectionDisabled(handle, forwards) {
         let dims = this.image_config.image_info.dimensions;
-        return (handle !== null && forwards || (dims.max_x * dims.max_y) > UNTILED_RETRIEVAL_LIMIT);
+        let tiled = (dims.max_x * dims.max_y) > UNTILED_RETRIEVAL_LIMIT;
+        let bytes_per_pixel = Math.ceil(Math.log2(this.image_config.image_info.range[1]) / 8.0);
```
jburel: you may want to add a ``TODO`` and create an issue so we can use the ``omero-model`` property when available https://github.com/ome/omero-model/pull/47

will-moore:
Maybe we should make this configurable within iviewer itself while we wait for configuration to be available on the server. I'd hate for someone to upgrade and then be blocked from viewing images that they have previously viewed OK.

jburel:
That will be certainly a good option

src/controls/dimension-slider.js
joshmoore: Ah interesting, so: see if someone set a value explicitly for iviewer, then check the server default, then use a client fallback.

will-moore: Currently we don't have the config available from the server.
When that is available, we could teach iviewer to use that as a default value, but if it is set in iviewer itself then use that. That could allow iviewer to be more restrictive than the server (but not *less* restrictive since the server would enforce it's own limit).

plugin/omero_iviewer/iviewer_settings.py
```diff
@@ -25,6 +25,13 @@
 # load settings
 IVIEWER_SETTINGS_MAPPING = {
 
+    "omero.web.iviewer.max_projection_bytes":
+        ["MAX_PROJECTION_BYTES",
+         1024 * 1024 * 256,
+         int,
+         ("Maximum bytes of raw pixel data allowed for Z-projection. "
+          "Above this threshold, Z-projection is disabled")],
```
jburel: Maybe indicate that this value will be ignored if the limit is greater that the one set server-side.
Server side is work-in-progress but when it is in-place that will be the rule.

will-moore: Seems wrong for the documentation to describe *future* functionality. 
Since I'll need to open a PR to update the functionality, I can change the description at the time to match (depending on what we end up with).

jburel: we can review the wording later on when we implement the size limit from server
My point is that if people start setting a high limit, this will be ignored later on and will default to the server one. This might come at a surprise for that.

PR #311, opened 2020-03-24
Roi query support

### LINKED ISSUE #308: Direct links to ROIs
* From the viewer, users should be able to copy a ROI url link.
 * A user going to that link should preferably see the same view as the user sending the link, i.e. viewer

will-moore:
Support ```/iviewer/?rois=1,2``` to open ROIs in iviewer.
Shows a thumbnail view of multiple ROIs.

To test:
 - Open iviewer and select some saved Shapes
 - Check the Link icon in top-right of the ROIs tab - (Button should only be enabled if 1 shape is selected)
 - Click the button. Link to `?shape=ID` should be shown and selected (if unsaved Shapes are selected, should see a suitable message).
 - Copy the link and paste into browser...
 - Page should open iviewer to show the shape
    - Test with shape on non-default Z/T and not on first 'page' (>500 ROIs) e.g. https://merge-ci.openmicroscopy.org/web/iviewer?shape=223406
 - Thumbnails on left should load as expected for single Image:
    - If shape is on Orphaned Image, or multiple parents - No thumbnails
    - If shape is on Dataset Image, show Dataset thumbnails
    - If shape is on SPW image, show Well thumbnails
 - Browser link/history should behave as for single Image:
    - Clicking the thumnail will update the URL to point to that image
 - Also test linking by ROI ID: `/iviewer/?roi=1`. The *first* shape will be selected.
 - Probably a good idea to test other URLs still work as expected (thumbnails & URL updates):
    - Test `/iviewer/?dataset=1`
    - Test `/iviewer/?images=1` and `?images=1,2`
    - Test default viewer URL `/webclient/img_detail/2/?dataset=2`

Fixes #308 

This requires a new release of `omero-web` with https://github.com/ome/omero-web/pull/157 and https://github.com/ome/omero-web/pull/159 and https://github.com/ome/omero-web/pull/178
 y

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -16,7 +16,7 @@
 #
 
 from django.shortcuts import render
-from django.http import JsonResponse
+from django.http import JsonResponse, Http404
 from django.conf import settings
 from django.core.urlresolvers import reverse
 
@@ -347,16 +347,36 @@ def plane_shape_counts(request, image_id, conn=None, **kwargs):
     return JsonResponse({'data': counts})
 
 
+def get_image_id_for_shape(conn, shape_id):
+    """Returns (roi_id, image_id) tuple for shape"""
+    params = omero.sys.ParametersI()
+    params.addId(shape_id)
+    query = """select roi from Roi roi
+        left outer join roi.shapes as shape
+        where shape.id=:id"""
+    query_service = conn.getQueryService()
+    result = query_service.findByQuery(query, params, conn.SERVICE_OPTS)
+    if result:
+        return (result.id.val, result.image.id.val)
+    return (None, None)
+
+
 @login_required()
-def roi_page_data(request, roi_id, conn=None, **kwargs):
+def roi_page_data(request, obj_type, obj_id, conn=None, **kwargs):
     """
     Get info for loading the correct 'page' of ROIs
 
     Returns {image: {'id': 1}, roi_index: 123, roi_count: 3456}
     """
-    roi = conn.getQueryService().get('Roi', int(roi_id))
-    image_id = roi.image.id.val
-
+    image_id = None
+    if obj_type == 'roi':
+        roi_id = int(obj_id)
+        roi = conn.getQueryService().get('Roi', roi_id)
+        image_id = roi.image.id.val
+    elif obj_type == 'shape':
+        roi_id, image_id = get_image_id_for_shape(conn, obj_id)
+    if image_id is None:
+        raise Http404(f'Could not find {obj_type}: {obj_id}')
     qs = conn.getQueryService()
     params = omero.sys.ParametersI()
 
@@ -364,9 +384,10 @@ def roi_page_data(request, roi_id, conn=None, **kwargs):
     query = "select roi.id from Roi roi where roi.image.id = :id"
     ids = [i[0].val for i in qs.projection(query, params, conn.SERVICE_OPTS)]
     ids.sort()
-    index = ids.index(int(roi_id))
+    index = ids.index(roi_id)
     rsp = {
         'image': {'id': image_id},
+        'roi': {'id': roi_id},
         'roi_index': index,
         'roi_count': len(ids)
     }
@@ -374,10 +395,17 @@ def roi_page_data(request, roi_id, conn=None, **kwargs):
 
 
 @login_required()
-def roi_image_data(request, roi_id, conn=None, **kwargs):
+def roi_image_data(request, obj_type, obj_id, conn=None, **kwargs):
     """ Get image_data for image linked to ROI """
-    roi = conn.getQueryService().get('Roi', int(roi_id))
-    image_id = roi.image.id.val
+    image_id = None
+    if obj_type == 'roi':
+        roi = conn.getQueryService().get('Roi', int(obj_id))
+        if roi:
+            image_id = roi.image.id.val
+    elif obj_type == 'shape':
+        roi_id, image_id = get_image_id_for_shape(conn, obj_id)
+    if image_id is None:
+        raise Http404(f'Could not find {obj_type}: {obj_id}')
     return image_data(request, image_id, conn=None, **kwargs)
 
 
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -346,6 +346,7 @@ def plane_shape_counts(request, image_id, conn=None, **kwargs):
 
     return JsonResponse({'data': counts})
 
+
 @login_required()
 def roi_page_data(request, roi_id, conn=None, **kwargs):
     """
@@ -378,6 +379,7 @@ def roi_image_data(request, roi_id, conn=None, **kwargs):
     roi = conn.getQueryService().get('Roi', int(roi_id))
     image_id = roi.image.id.val
     return image_data(request, image_id, conn=None, **kwargs)
+
 
 @login_required()
 def image_data(request, image_id, conn=None, **kwargs):
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -346,6 +346,12 @@ def plane_shape_counts(request, image_id, conn=None, **kwargs):
 
     return JsonResponse({'data': counts})
 
+@login_required()
+def roi_image_data(request, roi_id, conn=None, **kwargs):
+    """ Get image_data for image linked to ROI """
+    roi = conn.getQueryService().get('Roi', int(roi_id))
+    image_id = roi.image.id.val
+    return image_data(request, image_id, conn=None, **kwargs)
 
 @login_required()
 def image_data(request, image_id, conn=None, **kwargs):
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -36,6 +36,7 @@ import omero
 from omero.rtypes import rint, rlong, unwrap
 from omero_sys_ParametersI import ParametersI
 import omero.util.pixelstypetopython as pixelstypetopython
+from omeroweb.webclient.show import get_image_roi_id_for_shape
 
 from .version import __version__
 from omero_version import omero_version
@@ -347,20 +348,6 @@ def plane_shape_counts(request, image_id, conn=None, **kwargs):
     return JsonResponse({'data': counts})
 
 
-def get_image_id_for_shape(conn, shape_id):
-    """Returns (roi_id, image_id) tuple for shape"""
-    params = omero.sys.ParametersI()
-    params.addId(shape_id)
-    query = """select roi from Roi roi
-        left outer join roi.shapes as shape
-        where shape.id=:id"""
-    query_service = conn.getQueryService()
-    result = query_service.findByQuery(query, params, conn.SERVICE_OPTS)
-    if result:
-        return (result.id.val, result.image.id.val)
-    return (None, None)
-
-
 @login_required()
 def roi_page_data(request, obj_type, obj_id, conn=None, **kwargs):
     """
@@ -374,7 +361,7 @@ def roi_page_data(request, obj_type, obj_id, conn=None, **kwargs):
         roi = conn.getQueryService().get('Roi', roi_id)
         image_id = roi.image.id.val
     elif obj_type == 'shape':
-        roi_id, image_id = get_image_id_for_shape(conn, obj_id)
+        image_id, roi_id = get_image_roi_id_for_shape(conn, obj_id)
     if image_id is None:
         raise Http404(f'Could not find {obj_type}: {obj_id}')
     qs = conn.getQueryService()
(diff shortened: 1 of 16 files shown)


jburel:
From the screenshot, it seems that I should see the roi on the thumbnails. Is that the case?

jburel:
I will expect to have the shapes selected via the url to be selected in the tree too
Screenshot 2020-06-08 at 14 06 00

will-moore:
> From the screenshot, it seems that I should see the roi on the thumbnails. Is that the case?
Yes, I think they will be white since that's what we get from webgateway/render_roi_thumbnails

> old viewer and ended up unable to open the image if a large number of ROIs was selected. So we should avoid making the same mistake
Sorry - not sure what you mean?

> shapes selected via the url to be selected in the tree too
By "tree" do you mean the thumbnails on the left?
This is kinda tricky - E.g. would you expect the left panel highlighting to change when you are selecting different ROIs within an image? 

E.g. You click on the first thumbnail (ROI:1) -> this loads an Image and selects ROI:1.
Then you click on another ROI in the image (ROI:2). If there is no ROI:2 in the thumbnail list then you see no thumbnails selected. If there is an ROI:2 there then it becomes selected?

I'll see if that's possible...

jburel:
By ROI on Thumbnail, generating a thumbnail with ROI could be costly if we have many rois and many thumbnails to generate

will-moore:
From your screenshot above with "I will expect to have the shapes selected via the url to be selected in the tree too" I can see a shape selected in the table on the right, and this is working for me. So I'm not sure what's not working?

> we have been that route with the old viewer and ended up unable to open the image if a large number of ROIs was selected. So we should avoid making the same mistake

In that case, it was possible for an Image to have hundreds of ROIs and we didn't want to load the thumbnails for them all by default. I don't think we were unable to open the image.
This is a different workflow. You'd have to create a URL with hundreds of ROIs, in order to request that number of thumbnails.
I think we can choose a threshold maximum number of ROI thumbnails that we support and ignore any over that limit. Any ideas what that threshold should be? E.g. 200?

jburel:
I think the UI is misleading 
left-hand panel will so sometimes the selected images or the one in a dataset, and another time X copies of the same image with a shape on it.
If multiple rois are selected, only select them in the centre and right hand pane and do not try to generate a thumbnail with that shape.
Also here we are looking at roi with one shape, what about the case of roi with a large number of shapes.
The thumbnail strategy will not really work for very big images also here we are looking at 512x512 plane. Now it is very common to have 2kx2k plane.

jburel:
this might be related to this PR. Opening a dataset no longer works

sbesson:
Confirmed https://github.com/ome/omero-iviewer/pull/311#issuecomment-641094903 independently while testing merge-ci in a different context. The list of HTTP requests with their status code reported by the Network tools of Firefox is the following

```
200 GET /web/webgateway/luts/?_=1591688803994
200 GET /web/api/v0/m/datasets/19769/images/?offset=0&limit=10&_=1591688708980
500 GET /web/iviewer/roi/122574/image_data/?_=1591688803996
```

The ROI endpoint request certainly feels out of place.

will-moore:
Maybe @joshmoore can comment since it was he who requested this feature at #308

jburel:
There is no mention of generating a thumbnail in the original issue. 
The selection of the shape in the centre and the table works for one shape.

joshmoore:
@will-moore : you're asking my opinion on "The ROI endpoint request certainly feels out of place in the initial calls when a dataset/well is selected" ?

Assuming yes, @sbesson, you're saying not that it's odd to ask for ROIs when visiting a URL, but that the URL for the loading seems unexpected?

Assuming yes, I'd agree that this doesn't feel particularly iviewer-specific, though trying to do anything else would lead to coupled PRs, right, @will-moore ?

sbesson:
Sorry, trying to rephase my comment for clarity:

- ,y initial troubleshooting of the broken `Open Dataset/Well with iviewer fails`  functionality seems to be indicate it is related to this PR as a 500 HTTP error code is returned while making a `GET /web/iviewer/roi/....` call.
- maybe more precise than `out of place`, I was surprised by the presence of a GET request to the iviewer ROI API immediately when opening a dataset. I was expecting `/iviewer/image_data/...` but

will-moore:
I think maybe we need a call to discuss this PR.
We all seem to have very different ideas about how the feature should work.

will-moore:
@seb When you load the iviewer with just an ROI ID, we need to load the image_data based on the ROI ID. We could use a URL like `/iviewer/image_data/?roi=1` but it is a bit nicer not to use query strings for essential IDs. So a URL like `iviewer/roi/33/image_data/` is better. E.g. in the JSON api, to load images in a Dataset we do `/api/v0/m/datasets/52/images/` to filter images by Dataset ID. So `iviewer/roi/33/image_data/` is filtering image data by ROI ID.

will-moore:
@joshmoore No, sorry - I was asking your opinion about JM's "I think the UI is misleading".
I've implemented what I thought was the best solution for #308 but that RFE didn't specify what the UI should look like.

will-moore:
@jburel If the ROIs in the URL all come from different images, or are on different 'pages' of the same image or are even on different Z/T indexes within the same image, you won't be able to see them all selected at once on the image. And as soon as you click on the image, the selection will change, so you won't be able to use the current selection of ROIs as a way to browse all the ROIs you specified in the URL.

Using the thumbnails on the left, showing an ROI for each thumbnail allows you to browse all the ROIs you specified in the URL.

jburel:
Not all the elements have been taken into account in my view:
* Rois with many shapes
* Large images e.g.. Do you zoom in. Cost of generating a huge quantity of thumbnails for those images.
* The left-hand pane will give 2 different view depending on the query made e.g. rois vs image selection. This needs be reviewed to avoid major confusion from a user perspective. You are in a way introducing a roi browser within the same image or across images. People have asked for a "plane" browser in the past (able to see all the planes in of the same image), here it is "selected" roi browser. Let's not "mix" everything in the left-hand panel without careful consideration.
Let's discuss and design before diving into coding.

will-moore:
@jburel The number of ROI thumbnails that we show does not depend on whether the image is big or how many ROIs it has on it. It depends on the number of ROIs in the URL. There is no "Cost of generating a huge quantity of thumbnails for those images".

> Let's discuss and design before diving into coding.

Too late! But yes - let's discuss before I do more.

joshmoore:
> I was asking your opinion about JM's "I think the UI is misleading".

Oops, sorry. I don't know if this suggests two different mechanisms or not, but I think from #308 I'd be more concerned with being able to share a single ROI that has the optimal view (in terms of usability, zoom, etc.). Happy to discuss.

plugin/omero_iviewer/views.py
```diff
@@ -347,6 +347,68 @@ def plane_shape_counts(request, image_id, conn=None, **kwargs):
...
+        where shape.id=:id"""
+    query_service = conn.getQueryService()
+    result = query_service.findByQuery(query, params, conn.SERVICE_OPTS)
+    if result:
+        return (result.id.val, result.image.id.val)
+    return (None, None)
+
+
+@login_required()
+def roi_page_data(request, obj_type, obj_id, conn=None, **kwargs):
```
jburel, line 365: this could be a useful method to have out of iviewer.

jburel:
```
Page should open iviewer to show the shape
Test with shape on non-default Z/T and not on first 'page' (>500 ROIs) e.g. https://merge-ci.openmicroscopy.org/web/iviewer?shape=223406
```
Impossible to view the shape on that big image without zooming in

jburel:
``
Check the Link icon in top-right of the ROIs tab - (Button should only be enabled if 1 shape is selected)
``
This works as expected

jburel:
```
Thumbnails on left should load as expected for single Image:
```
- [x] If shape is on Orphaned Image, or multiple parents - No thumbnails
- [x] If shape is on Dataset Image, show Dataset thumbnails
- [x] If shape is on SPW image, show Well thumbnails

jburel:
* Tested the various links: dataset, images: it works as expected
* https://merge-ci.openmicroscopy.org/web/iviewer/?roi=291069 works as indicated
* Missing: ability to copy the roi link from the UI using the "link" icon

plugin/omero_iviewer/views.py
```diff
@@ -347,6 +347,68 @@ def plane_shape_counts(request, image_id, conn=None, **kwargs):
     return JsonResponse({'data': counts})
 
 
+def get_image_id_for_shape(conn, shape_id):
```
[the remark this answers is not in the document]
will-moore: I'm not sure that the amount of code duplication here is enough to justify switching to use omero-web method.

will-moore:
I don't think we want to offer the ability to copy a link to an ROI, since you can't select an ROI (only a Shape). I think the decision when discussed recently was just to support link to shape from the UI (to allow to return to the shape you've currently selected), but also allow link to ROI if you know the ROI ID.

I will look at supporting zoom...

plugin/omero_iviewer/views.py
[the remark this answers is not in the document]
jburel: I disagree.

will-moore:
With those commits, linking to a Shape or ROI should zoom in so that the selected Shape is nicely visible and pan so the shape is in the centre of view.
Test with drawing shapes of very different sizes and types (only tested with Polygons so far - not Points etc), particularly on a Big image.
Also check that existing selection behaviour hasn't changed when selecting ROIs in the Table (don't change zoom on selection and don't pan if the selected shape is already visible in the viewport.

jburel:
using https://merge-ci.openmicroscopy.org/web/iviewer/?shape=223406 (user-3) for review
* Image loads in "default" view when loading the rois
* Loading the rois was very slow
* When the rois are loaded, the shape is selected and image zoomed 
Screenshot 2020-07-02 at 10 47 17

jburel:
Using this time not a big image https://merge-ci.openmicroscopy.org/web/webclient/img_detail/83221/?dataset=5220 (user-3)
* when the shape is selected, the image is "shifted" up https://merge-ci.openmicroscopy.org/web/iviewer/?shape=279241

screenshot when the image is opened from the web
Screenshot 2020-07-02 at 10 51 45

screenshot when the image is open using the url with the shape
Screenshot 2020-07-02 at 10 51 56

jburel:
I use this time the "roi" filter on a big image
https://merge-ci.openmicroscopy.org/web/iviewer/?roi=217842
the shape is correctly selected and the image zoomed in

src/regions/regions.js
```diff
@@ -153,4 +155,35 @@ export default class Regions {
...
+     */
+    getRoisLink(shapes) {
+        if (!shapes || shapes.length === 0) {
+            return "No shapes selected";
+        }
+        // Only take first ROI:shape
+        let shapeId = shapes[0].split(':')[1];
+        // If unsaved, will have negative IDs. Return message instead
+        if (shapeId < 0) {
+            return "Can't link to unsaved ROIs";
```
jburel, line 185: Shouldn't be unsaved Shapes since the link only allows to select shapes

src/viewers/viewer/Viewer.js
```diff
@@ -829,19 +844,16 @@ class Viewer extends OlObject {
 
     /**
      * Centers view on the middle of a geometry
+     * unless geometry is already in viewport,
```
jburel, line 847: this is not what is currently happening cf. testing report

will-moore: That is describing the default behaviour of the method.
You can use the `forceCentre` argument to ensure that you get the shape centred.

will-moore:
This is getting complex.
We currently have 2 zoom/pan behaviours for when we select a shape:
 - When you select a shape in the table: don't zoom and only pan if the shape doesn't intersect with the viewport
 - When you link to an roi/shape: zoom if you need to (shape is small) and always pan the shape to centre

So what should the new link-to-shape behaviour be?
 - Zoom if you need to. If you didn't need to zoom then only pan if the shape doesn't intersect with the viewport?

src/regions/regions.js
will-moore: I guess it could be either. Since we're in the "ROIs" tab and when you "Save" you're saving ROIs. We don't really use "Shapes" terminology anywhere else in the UI I think.

jburel:
The "shift" problem described in https://github.com/ome/omero-iviewer/pull/311#issuecomment-652909790 is still there

will-moore:
Sorry - I *think* I just got the zoom/resolution the wrong way around. Hopefully fixed now.

jburel:
The shift problem is now fixed. Thanks
From a functional testing, it is now working.
Still the issue of code duplication i.e. ``get_image_id_for_shape``

will-moore:
Done: 7ea2569
Also bumped omero-web requirement (even though it's not released yet). Might fail build/deploy?

will-moore:
As expected, travis failed with "Could not find a version that satisfies the requirement omero-web>=5.7.0"

jburel:
Everything looks good now. 
We should now prep a release. @will-moore could you prep a release?

PR #320, opened 2020-05-01
Ignore rendering settings

will-moore:
This is a potential emergency patch to disable rendering settings.
We don't disable the controls, but the settings are ignored (not added to the tile URL).
cc @joshmoore


joshmoore:
Thanks, @will-moore . I'll keep this in my back pocket ;)

PR #325, opened 2020-06-19
Allow higher max zoom for Big images

will-moore:
See https://forum.image.sc/t/omero-iviewer-zoom-limits/39220

This allows higher zoom levels for big images.
To test, try zooming in on a big image. Should now be able to go to 644% (instead of 161%).

Also check that max zoom of non-Big images is unchanged at 4000% (Not sure if we should reduce this?).

### DIFF
--- a/src/viewers/viewer/Viewer.js
+++ b/src/viewers/viewer/Viewer.js
@@ -534,6 +534,11 @@ class Viewer extends OlObject {
             }
         }
 
+        let highestRes = possibleResolutions[possibleResolutions.length-1];
+        // For Big images, allow zooming in further (from 160% up to 322%)
+        if (highestRes > 0.5) {
+            possibleResolutions.push(highestRes/2);
+        }
         // we need a View object for the map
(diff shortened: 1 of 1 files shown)


AarghIMApiratez:
Hi Will,

Thanks for the quick response and starting a pull request.

This is purely from a user perspective. But I think it makes sense to allow more zoom for a larger images That is, regardless of an image's size you end up with pixels that are of the same size on your screen.

See my two images attached (Viewed without interpolate).
The two images are imaged at the same resolution / settings etc.

Small image A.  allows 4000% zoom. Where pixels are huge. (~2 times my cursor height?)
Larger image B. allows 161% zoom (now changed by you to 322%). Here I can barely make them out.

In my view a good value would be 500% (image C). 322 might do the trick, but should be the same regardless of size.


Image A
image

Image B
image

Image C
image



Mind that I  havent thought about if/how resolutions on different screens might affect what is reasonable.


Friendly regards, Staffan

PR #327, opened 2020-06-23
Link to usage guide

manics:
I couldn't remember how to use the advanced features in iviewer. This makes it easier to find the relevant documentation.


jburel:
Thanks. We could probably do the same for the other apps

PR #328, opened 2020-06-23
Link image ID back to webclient

manics:
Previously there was no way to go back to the image in the webclient, you could only go to the dataset. This makes the Image ID a link.

Testing: Click on the Image ID, you should be taken to the image in the webclient


jburel:
it works nicely. Thanks

PR #330, opened 2020-06-30
fix link following re-organisation

jburel:
Now that we are promoting the guide more and more we will need to be careful not to break link
cc @pwalczysko

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ Usage
 =====
 
 A guide to using OMERO.iviewer can be found on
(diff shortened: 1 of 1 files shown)


pwalczysko:
> we will need to be careful not to break link

How to know that a link is broken ? Even more importantly, how to build all the guides so that one can see all the cross-linking and can check it at least manually ?

jburel:
Changing the name of the rst file will automatically break a link. @sbesson has kindly started a CI job (WIP)
getting there slowly.

pwalczysko:
Works fine, the link is pointing to the right place.

Thank you

Ready to merge fmpov

PR #334, opened 2020-08-03
test: print to console

jburel:
print error output to the console when running the tests.
This was useful when investigating https://github.com/ome/omero-plugins/pull/42

cc @sbesson


will-moore:
Not sure how to test this but it looks reasonable.
Ran `$ ant unit-tests` on this branch but don't expect to see any `infrastructure_error` output.

Issue #335, opened 2020-08-07
large number of ROIs: tool not responsive


jburel:
The tool is totally unusable when an image has a large number of ROIs with several shapes.
* Loading the data is very slow (nightshade)
* interaction with roi/shape on left-hand panel implies that a wrong click and the user will have to wait
* if the user opts to show one roi only and click on a shape belonging to that rois, all the other rois are again loaded and display even if "show" is not selected.
* No option to delete all the rois.

Basically there is no way to follow one roi over time when playing the movie
cc @pwalczysko @will-moore

will-moore:
- Slow loading is certainly a problem, but this is due to the JSON api. With `omero-marshal` we load a lot of extra JSON object.details that we don't need.
 - We never really tested pagination of ROIs (when you have more than 500 on a multi-plane image) AND multiple shapes per ROI. I also found this very painful with an image in idr0085.
 - We only allow you to delete Shapes (can't even select ROIs)

![Screenshot 2020-08-07 at 15 55 44](https://user-images.githubusercontent.com/900055/89658793-e2e4ef80-d8c6-11ea-9736-14ae5616f39f.png)

jburel:
<img width="657" alt="Screenshot 2020-08-07 at 16 12 40" src="https://user-images.githubusercontent.com/1022396/89660290-0dd04300-d8c9-11ea-9b1a-db572aadd24f.png">

only one shape (top left-corner) belongs to the selected rois but all the shapes under the "unselected" rois are selected when switching to another plane i.e. out of synch

jburel:
* Fix selection bug when tracking an object i.e. select a shape in the table (right) when we move to the new plane. The selected shape is no longer selected
* Review loading of data, this might imply not using json api during the investigation's phase and look into porting it to the API
* Look into per ROI pagination instead of Plane pagination

will-moore:
I started looking at optimised querying of OMERO via projection to improve performance, only loading bare minimum values.
For example, loading 1104 ROIs (11169 Points) from a local copy of this Image:
http://idr.openmicroscopy.org/webclient/?show=image-9846159
using the `webgateway/get_rois_json/` in the old viewer took about 4 seconds for the query using the
ROI Service.
Using the query service to load minimal coordinates via a projection reduced this to about 1 second:

```
    query = """
        select shape.id,
               shape.roi.id,
               shape.x,
               shape.y,
               shape.theZ,
               shape.theT
        from Shape shape
        join shape.roi roi
[...]

PR #342, opened 2020-08-24
Unset initial_shape_id once selected

will-moore:
Fixes one of the issues reported at #335 and a related bug:

To test:
 - Find a multi-plane image that has > 500 ROIs (where ROIs are paginated by Z/T plane
 - Browse to a Shape and copy link to it. (workflow from https://github.com/ome/omero-iviewer/pull/311)
 - Paste link into browser - iviewer should browse to that Z/T plane and select the Shape
 - Try browsing to a different Z/T plane, either with Z/T sliders or via clicking on another Shape
 - Viewer should show the newly-selected Z/T plane and NOT be redirected back to the original Z/T with Shape selected.
 - If browsing via clicking on a Shape, the newly-clicked shape should be selected when the ROIs load for new plane.
 - Also, test the above workflow on an image where ROIs aren't paginated

cc @jburel

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -425,7 +425,6 @@ def roi_page_data(request, obj_type, obj_id, conn=None, **kwargs):
             params.addId(image_id)
             result = qs.projection(query, params, conn.SERVICE_OPTS)
             for roi in result:
-                print(unwrap(roi[0]))
                 ids.append(unwrap(roi[0]))
     if image_id is None:
         raise Http404(f'Could not find {obj_type}: {obj_id}')
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -219,6 +219,42 @@ def persist_rois(request, conn=None, **kwargs):
     return JsonResponse(ret)
 
 
+def get_query_for_rois_by_plane(the_z=None, the_t=None, z_end=None,
+        t_end=None, load_shapes=False):
+
+    clauses = ['roi.image.id = :id']
+    if the_z is not None:
+        where_z = "(shapes.theZ = %s or shapes.theZ is null)" % the_z
+        if z_end is not None:
+            where_z = """((shapes.theZ >= %s and shapes.theZ <= %s)
+                or shapes.theZ is null)""" % (the_z, z_end)
+        clauses.append(where_z)
+    if the_t is not None:
+        where_t = "(shapes.theT = %s or shapes.theT is null)" % the_t
+        if t_end is not None:
+            where_t = """((shapes.theT >= %s and shapes.theT <= %s)
+                or shapes.theT is null)""" % (the_t, t_end)
+        clauses.append(where_t)
+
+    query = """
+        select distinct(roi.id) from Roi roi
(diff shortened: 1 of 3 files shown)


jburel:
Not related to this PR but noticed while testing https://github.com/ome/omero-iviewer/issues/346

jburel:
* Select a shape in the right-hand table. The plate is loaded and shape selected
Screenshot 2020-08-26 at 06 51 50
* Copy the shape link and paste in another tab. Shape is not selected
Screenshot 2020-08-26 at 06 52 07

will-moore:
OK, I found 2 bugs for loading `?shape=328263` above.
When we want to load ROIs (using shape ID) from an image where ROIs are paginated (>500 ROIs on the image), we use `iviewer/shape/3328263/page_data/` to look-up which page the ROI is on. In this case we get:
```
{
"image": {
"id": 139315
},
"roi": {
"id": 319825
},
"roi_index": 263,
"roi_count": 1104
}
```
Then we load the appropriate page of ROIs to show on the Image and can highlight the appropriate shape.
This works when we have a single-plane image with paginated ROIs (> 500) and also works on a multi-plane image
where we don't have paginated ROIs. Here, although the first plane that loads is not the correct one for the shape,
but triggering a 'shape select' event, we re-load the correct plane (as if you've clicked on the shape in the table).

However, when we have a multi-plane image with paginated ROIs we have a couple of problems:
Loading the ROIs that are on the initial 'default' plane won't necessarily include the target Shape, so when we trigger 'select shape', that shape isn't found (not been loaded) so we don't update to the correct Z/T plane.

Also, if the `page_data` info above indicated that the ROI would be on page 2 then we use page 2 `?offset=500&limit=500` when loading the ROIs for the default plane - even though there are only a small number of ROIs on that plane!

So, the `plane_data` needs to include the Z and T indices for the Shape and the `roi_count` and `roi_index` needs to refer to the ROIs on *that* plane! Then iviewer needs to load ROIs for the correct Plane directly.

will-moore:
Phew - that was a lot of work to hopefully fix that bug.
Lots has changed

jburel:
bug described in https://github.com/ome/omero-iviewer/pull/342#issuecomment-680672486 is now fixed

jburel:
Looks at image with limited number of ROIs, behaviour is as expected

jburel:
Outside the scope of this PR, looking at https://merge-ci.openmicroscopy.org/web/webclient/img_detail/105665 (user-3) > 150000 ROIs, there is no way to identify the page to select when we are interested in one area. Ideally user should be able to select a region and load the rois for that region, not the opposite

will-moore:
I looked into loading ROIs by region in https://github.com/ome/omero-iviewer/pull/242. But there are a number of limitations - lot least is the lack of any way to query Shapes by coordinates.

Another approach to browsing ROIs by x/y coordinates is in my parade prototype (see https://youtu.be/FyjGhZxx6es?t=422) but this is more about plotting analysis results and is very much a read-only use-case.

jburel:
The problem that this PR tries to fix is now solved.
Merging. More work to be done in follow-up PRs

PR #343, opened 2020-08-24
delta-T units fix

### LINKED ISSUE #340: OMERO.iviewer ignores plane DeltaT units in timestamp display
Reproduced with the current development version of iviewer as well as with the current released version.

To test, the two following sample OME-TIFF filesets can be used:

[timelapse-hours.zip](https://github.com/ome/omero-iviewer/files/5090948/timelapse-hours.zip)
[timelapse-seconds.zip](https://github.com/ome/omero-iviewer/files/5090949/timelapse-seconds.zip)

These are strictly identical timelapse images with the only difference being the unit of the DeltaT in the plane metadata:

```diff
sbesson@ls30630:tmp $ diff test.companion.ome test2.companion.ome 
23,27c23,27
<           <Plane DeltaT="0" DeltaTUnit="h" TheC="0" TheT="0" TheZ="0"/>
<           <Plane DeltaT="1" DeltaTUnit="h" TheC="0" TheT="1" TheZ="0"/>
<           <Plane DeltaT="2" DeltaTUnit="h" TheC="0" TheT="2" TheZ="0"/>
<           <Plane DeltaT="3" DeltaTUnit="h" TheC="0" TheT="3" TheZ="0"/>
<           <Plane DeltaT="4" DeltaTUnit="h" TheC="0" TheT="4" TheZ="0"/>
---
>           <Plane DeltaT="0" DeltaTUnit="s" TheC="0" TheT="0" TheZ="0"/>
>           <Plane DeltaT="1" DeltaTUnit="s" TheC="0" TheT="1" TheZ="0"/>
>           <Plane DeltaT="2" DeltaTUnit="s" TheC="0" TheT="2" TheZ="0"/>
>           <Plane DeltaT="3" DeltaTUnit="s" TheC="0" TheT="3" TheZ="0"/>
>           <Plane DeltaT="4" DeltaTUnit="s" TheC="0" TheT="4" TheZ="0"/>
```

The image with plane timestamps in seconds displays as expected both in the Acquisition pane as well as in the timestamp in the central pane of iviewer

Screenshot 2020-08-18 at 15 23 33
Screenshot 2020-08-18 at 15 23 56

The image with plane timestamps in hours have the correct unit in the Acquisition pane but the timestamp displayed in the central pane of iviewer is identical to the above e.g. it displayes as `00:00:02:000` while I would have expected `02:00:00:000`

Screenshot 2020-08-18 at 15 23 38
Screenshot 2020-08-18 at 15 24 17

will-moore:
Fixes #340 

To test:
 - Import the sample files from issue above
 - Check timestamps in iviewer for delta-T in Hours vv Secs.

cc @sbesson

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -483,7 +483,7 @@ def format_value_with_units(value):
     if value is None:
         return (None, "")
     length = value.getValue()
-    unit = value.getUnit()
+    unit = str(value.getUnit())
     if unit == "MICROMETER":
         unit = lengthunit(length)
         length = lengthformat(length)


sbesson:
Retested using the uploaded sample files

Screenshot 2020-08-25 at 11 06 55
Screenshot 2020-08-25 at 11 06 57

The change fixes the issue immediately. Proposing to get this as as an immediate patch release of OMERO.iviewer for deployment in the upcoming release of IDR if no objection?

As a side note, I was looking at the code which is handling the conversion. I was confused by the handling of time vs length. One section of the code seems to be converting to the most appropriate unit but the other section unconditionally converts to seconds. Should these be separate functions? Additionally, the list of source units and the conversion factors are completely hardcoded. I would assume some refactoring should happen to consume the logic from the upstream `omero_model_{TimeI,LengthI...}` classes unless there is a specific reason not to do so. I can create an appropriate RFE issue if that's the best place.

will-moore:
Yes, sounds good.

PR #336, opened 2020-08-11
Help: provide help

manics:
The help menu doesn't currently provide any help.
image

This PR adds a new link to the guide.

PR #344, opened 2020-08-25
Update link to how-to guide

sbesson:
Introduced in #336, this unifies the target link with what is in the README and points at the guide published as part of the omero-guides project.

### DIFF
--- a/src/app/header.html
+++ b/src/app/header.html
@@ -108,7 +108,7 @@
                 <ul class="dropdown-menu">
                     <li>
                         <a target="_blank"
-                            href="https://omero-guides.readthedocs.io/projects/iviewer/">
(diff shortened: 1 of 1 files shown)


manics (approved):

src/app/header.html
```diff
@@ -108,7 +108,7 @@
                 <ul class="dropdown-menu">
                     <li>
                         <a target="_blank"
-                            href="https://omero-guides.readthedocs.io/projects/iviewer/">
+                            href="https://omero-guides.readthedocs.io/en/latest/iviewer/docs/index.html">
```
manics, line 111: Could optionally remove the `index.html`, I think it makes the URL look a bit nicer but not essential

sbesson: :+1:

manics (approved):

PR #349, opened 2020-08-28
Tiling threshold max projection bytes

### LINKED ISSUE #341: Projections and tiled images threshold
I'm only now seeing that projections are not available for some images because of this PR: https://github.com/ome/omero-iviewer/pull/202
Is there a specific reason to set the threshold for tiled images at [4000000](https://github.com/ome/omero-iviewer/blob/master/src/viewers/viewer/globals.js#L111) (multiple of 1000) instead of 4194304 (multiple of 1024)? 

2048x2048 images are excluded although they're quite common (I think)...

I guess people can always ask for a slightly higher threshold, but I'm under the impression that images often have resolutions based on powers of 2, and the closest threshold for 2000 would be 2048, which is close, where as bigger images with a "standard" resolution would surely be at least 2560 (if not 3072 or 4096), which is considerably larger.

will-moore:
Fixes #341. 

This boosts the "Big Image" (use tiles) threshold in iviewer from 2k * 2k to `2048 * 2048`.
Also, it makes iviewer respect the value of the server config: `omero config set omero.pixeldata.max_projection_bytes`.
So, if you boost this value on the server, iviewer will use that value, instead of needing to also update the corresponding config of `omero.web.iviewer.max_projection_bytes` in iviewer.

To test, an image should be enabled for projection in iviewer IF...

With default settings:
- The image must be below `1024 * 1024 * 256` bytes (e.g. 1 channel 8-bit image of X, Y, Z: 2048 * 2048 * 64) AND...
- Image `X * Y` must be below `2048 * 2048`.

 If `omero.pixeldata.max_projection_bytes` setting is doubled to `1073741824` (`1024 * 1024 * 512 * 2`):

- The image bytes must be below this limit (e.g. 2 channel 8-bit image of X, Y, Z: 2048 * 2048 * 64) AND...
- Image `X * Y` must be below `2048 * 2048`.
  

If `omero.web.iviewer.max_projection_bytes` setting is set to LESS than `omero.pixeldata.max_projection_bytes` then this lower threshold will be used:
- The image bytes must be below this limit (e.g. 2 channel 8-bit image of X, Y, Z: 2048 * 2048 * 64) AND...
- Image `X * Y` must be below `2048 * 2048`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,13 +50,19 @@ https://omero-guides.readthedocs.io/en/latest/iviewer/docs/index.html
 Settings
 ========
 
-OMERO.iviewer limits the size of Z-projections to reduce load on the server.
+OMERO limits the size of Z-projections to reduce load on the server.
 The limit is defined as the number of bytes of raw pixel data in a Z-stack and
-is equivalent to 1024 * 1024 * 256 bytes.
-For example, an 8-bit image (1 byte per pixel) of size 1024 * 1024 * 256 is
-equal to the default threshold. To double the limit, use::
+the OMERO.server default is equivalent to 1024 * 1024 * 256 bytes.
+For example, a single-channel 8-bit image (1 byte per pixel) of XYZ size
+1024 * 1024 * 256 is equal to the default threshold.
 
-    $ omero config set omero.web.iviewer.max_projection_bytes 536870912
+To double the limit, use::
+
+    $ omero config set omero.pixeldata.max_projection_bytes 536870912
+
+If you wish to set a threshold for iviewer that is *lower* than for the server:
+
+    $ omero config set omero.web.iviewer.max_projection_bytes 268435456
 
 NB: Z-projection is not supported for tiled images in OMERO
 (Images larger than 2048 * 2048 pixels per plane are tiled in iviewer).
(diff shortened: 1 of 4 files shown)


jburel:
https://merge-ci.openmicroscopy.org/web/webclient/img_detail/138051/?dataset=24051 (user3) has a dimension of ``2048 x 2044``
and 31 z-sections. The projection is not enabled

will-moore:
The default `max_projection_bytes` on the server and iviewer is `1024 * 1024 * 256`.
So if you have a `2048 x 2048 x 31` image then you are only allowed 2 bytes per pixel.
e.g. a 2-channel 8-bit image. 
That image above has 4 channels of uint16 so it is quite a bit over that threshold.

We need to test setting of a higher values for `omero.pixeldata.max_projection_bytes` on the merge-ci server.

jburel:
That does not match ``an image of 2048 x 2048 should be enabled for Max-intensity projection``

will-moore:
OK, sorry - I forgot to say that it still has to be below the max_projection_bytes threshold, which is quite a low cut-off for images that are 2048 x 2048, as discussed at https://github.com/ome/omero-iviewer/issues/341#issuecomment-677465843

So, if we leave the `omero.pixeldata.max_projection_bytes` at the default setting then we need an image of just the right size to test that smaller `2048 x 2048` images can be projected.
But we also need to test that different settings are respected by iviewer (and the server) to allow bigger images to be projected.

jburel:
retested.
Thanks

sbesson:
See https://github.com/IDR/deployment/pull/322#issuecomment-808046899 for a compatibility issue with OMERO.server 5.6.0

PR #351, opened 2020-09-07
Lazy load shapes

will-moore:
This addresses just the first 2 bullet-points from https://github.com/ome/omero-iviewer/issues/335#issuecomment-686383167

In this PR, when ROIs are paginated (over 500 ROIs), we load just those shapes that are on the current plane.
The ROI JSON includes `shape_count` so we can continue to display that.
When an ROI is expanded (either by clicking on the ROI, or when a shape within the ROI is selected), then the Shapes are loaded.

This PR needs https://github.com/ome/omero-marshal/pull/69

To test:
 - Need to test all ROI/shape workflows: load, edit, delete, create, save etc.
 - Test on images paginated ROIs (over 500) and non-paginated.


plugin/omero_iviewer/views.py
```diff
@@ -219,6 +219,51 @@ def persist_rois(request, conn=None, **kwargs):
...
 
 
+def get_query_for_rois_by_plane(the_z=None, the_t=None,
+                                z_end=None, t_end=None):
+
+    clauses = ['roi.image.id = :id']
+    if the_z is not None:
+        where_z = "(shapes.theZ = %s or shapes.theZ is null)" % the_z
+        if z_end is not None:
+            where_z = """((shapes.theZ >= %s and shapes.theZ <= %s)
```
mtbc, line 229: Interesting you can't use equals here, curious about that!

mtbc: Oh, sorry, misread it, I understand now, ignore me!

jburel:
From last week discussion, I understood that if a shape in a roi is selected, then a movie either across z or t  is played the shape(s) in that roi will be displayed if there is any on the plane. This is not the case.

will-moore:
See description "This addresses just the first 2 bullet-points from #335 (comment)"
They are:

 - Load only the shapes on the plane that we can look at. The ROI count is loaded
 - When the user select a shape, the other shapes will be loaded

This doesn't include the next 2 points:

 - If user plays a movie, only the selected "loaded" shapes will be displayed
 - Keep the shapes not linked to a given plane in the display

Actually, it also does support the last point (no change needed)
 - Go back to the first step when the user stops playing the movie

jburel:
@will-moore apologies. I got ahead of the game.

will-moore:
That commit should cover the next bullet points (but not tested yet with "shapes not linked to a given plane":

 - If user plays a movie, only the selected "loaded" shapes will be displayed
 - Keep the shapes not linked to a given plane in the display

The user experience is definitely better than it was, but there remains an issue with the frame rate of image pixel data loading is different from the rate that ROIs are updating.
This is not related to this PR: it's an issue on existing iviewer (when ROIs are not paginated/lazy-loaded).

jburel:
Previously selecting a ROI, will select all the shapes in that ROI. This means that when I start playing a movie. I have the "selected" bold border around the shape. Now when I select a ROI, the last shape only is selected, so when I play a movie the selection is gone

jburel:
I can imagine a user being annoyed by the chosen approach so we might have to rethink this use cases.
User has several tracks, he/she selects only one ROI (spent time unchecking the boxes) start playing the movie, the shapes for the select rois are displayed. He/she stops the movie to view a specific plane for example half-way through the tracks. At that stage, we are back to the first step, the user will have to unselect again the boxes before continuing playing the movie.

will-moore:
OK, there's a couple of issues here:
 - Selection of all shapes in the ROI:
    - The behaviour is always to try and select all Shapes when we click on the ROI in the table (not on the expand arrow).
    - This is done sequentially, setting selected = true on each Shape in turn.
    - Since the last shape is the last one to be selected, we end up browsing to that Z/T plane (This is probably NOT what we want?!)
    - If we have all the Shapes already loaded (no pagination) then they all stay selected. BUT, if we're reloading Shapes for the new (last) Plane then only the last Shape is selected. Only a single shape_id is used to remember the previous selection (see previous PR):
https://github.com/ome/omero-iviewer/pull/342/files#diff-b445f40c6859f02edacb925f6ab66d49R876 
and it is only the selection of this shape (that is loaded because it's on the new Z/T plane) that causes the ROI to expand and the other Shapes in that ROI to be loaded. So, we'd need to remember the multi-shape selection up until that point and select those shapes when they are re-loaded.

 - Hiding of Shapes
    - The hide/show state of previously-loaded ROIs (and shapes) is lost when we load ROIs for a new plane.
    - However, that is only part of the problem. You will also be seeing ROIs on the new plane that weren't even loaded when you started the movie (so you couldn't have hidden them - even if the hiding state was preserved). So, we're always going to risk seeing additional shapes appear on a new plane, even when we can preserve the hide/show state.

It seems with both these issues, we are in trouble because we wipe out existing ROI layer for the new plane, and re-create from scratch, instead of updating the current one. So maybe I'll try keeping the ROI layer and all ROIs that were previously loaded and simply add the new ones in when they're loaded.

Also I'll try to fix the ROI selection to NOT navigate to the plane of the last Shape. If the current plane has one of the Selected shapes on, don't change Z/T - otherwise navigate to the plane of the FIRST shape?

pwalczysko:
Noted that the "Show" checkbox is not present on merge-ci atm.

Screen Shot 2020-09-15 at 10 49 39

will-moore:
That last commit should fix the 'Show' checkbox.
I think this is ready for another round of testing.

Now, this should work for image with paginated ROIs (> 500)
 - Open ROIs tab - ROIs with shapes on the current plane are loaded - other shapes in those ROIs are not loaded
 - Click on a shape on the viewer to select it (expands the ROI in the table) - or manually expand the ROI in the table
 - This should load the other Shapes for that ROI
 - Can click on the ROI in the table to select all the Shapes
 - Can then play the movie. The selected shapes will be shown on each plane in turn (don't need to load new shapes while movie is playing)
 - When the movie stops, we load all the Shapes for that plane
 - Should be able to select another shape and repeat...

I also added a few notes to the docs/README and a diagram to try and retain some of the knowledge about how this all works, since it's not possible to keep it all in my head and takes a while to work out from the code.

jburel:
I used https://merge-ci.openmicroscopy.org/web/webclient/img_detail/105665/
``Show`` out of synch:
 * Go to the ROI tab, the regions are loaded
 * Untick the ``Show`` checkbox, select a couple of shape
 * Move to another page, all the shapes are selected but in that case, the ``Show`` is not selected

Screenshot 2020-09-17 at 14 25 30
Screenshot 2020-09-17 at 14 26 53

jburel:
I used https://merge-ci.openmicroscopy.org/web/webclient/img_detail/139315/
ROI selection out of synch
* untick ``Show``
* Click on the arrow to expand the rois, all the shapes are selected but not the rois

Screenshot 2020-09-17 at 14 35 14

docs/README.md
```diff
@@ -159,14 +173,73 @@ This adds a Vector layer to the OpenLayers Map:
...
+
+Pagination of ROIs & Shapes
+---------------------------
+
+If the number of ROIs on the image is greater than ``ROI_PAGE_SIZE`` (default setting is 500)
+then we only load ROIs that have a Shape on the current Z/T plane.
+ROIs will also be paginated within each plane if over 500 ROIs per plane.
+ROIs will only have loaded the Shapes on the current plane. Other Shapes in the ROI won't
+be loaded, but the shape_count is still displayed on the `regions-list` table.
+When a user expends the ROI in the table (or a Shape is selected by clicking on the image or
```
pwalczysko, line 202:
```suggestion
When a user expands the ROI in the table (or a Shape is selected by clicking on the image or
```

jburel:
I used  https://merge-ci.openmicroscopy.org/web/webclient/img_detail/139315/
* Select one roi, and play movie across z. This is now possible
* Stop the movie at 291. Regions are loaded. But some regions with no shapes on that plane are also loaded. The roi is not selected
Screenshot 2020-09-17 at 14 39 16

docs/README.md
```diff
@@ -159,14 +173,73 @@ This adds a Vector layer to the OpenLayers Map:
...
+via URL ?shape=1) then `roi.expanded` is set to `true`. This is `@observed` by the
```
pwalczysko, line 203:
```suggestion
via URL ?shape=1) then `roi.expanded` is set to `true`. This is observed by the
```

docs/README.md
pwalczysko: Maybe the observed syntax is intentional, but it does not match the other "observed" occurences in this doc.

docs/README.md
pwalczysko: Maybe the "@observed" syntax is intentional, but it does not match the other "observed" occurences in this doc.

pwalczysko:
> Regions are loaded. But some regions with no shapes on that plane are also loaded. The roi is not selected

Can confirm @jburel 's observation. If you do not select any ROI and then play the movie, then only the ROIs which are on a given frame will be loaded once you stop the movie. 
But, if you select a ROI first, then play a movie, then many ROIs load when you stop the movie at any given frame which have no shapes on that frame. If you expand such ROIs (which are loaded in the table, but not checked as "show", their shapes, as indicated, are not belonging to that plane, but some of the shapes are checked as "show" in the table and some are not, without any apparent logic.

Screen Shot 2020-09-17 at 17 41 34

will-moore:
There's a couple of issues here.
I'm currently not discarding previously-load ROIs and shapes when you navigate to a new Plane.
So, the first plane you load, *only* ROIs with shapes on that plane will be loaded. But when you go to another plane, now you will have shapes from both planes. I could revert that. I didn't think that it would be confusing since I didn't expect the user to care. But I guess I'm also not getting any benefit from keeping them in hand.

There isn't yet a place in the JS model that stores the 'checked' state of an ROI checkbox. So when loading shapes for the ROI, there's no way to know whether they should be shown. So I'll have to store the ROI checkbox state.

will-moore:
Having spent a fair bit more time on this, it's proving very hard to manage the selection state of Shapes while changing Z/T and loading new ROIs/Shapes.
The previous behaviour was to clear all shapes from the `regions_info` model class and the OpenLayers Regions layer when loading new ROIs for new Z/T.
But this loses all the selection and visibility info.
The alternative above was to NOT clear existing Shapes when new ones are loaded. But then we have the issue of having some ROIs/Shapes loaded that are not on the current Z/T plane.
ef3f2f2 is an attempt to use the `regions_info.selected_shapes` list to set the selected state of Shapes when they are re-loaded. However, this doesn't do the same for the OpenLayers Regions layer, so we get out of sync.
Also, when loading the ROIs/Shapes for a newly chosen Z/T plane, there is no trigger to expand the ROIs that have selected Shapes, so the shapes aren't loaded.

Probably there are fewer issues to deal with when we *don't* clear all the existing Shapes.
How much of a problem is it to have ROIs/Shapes loaded that aren't on the current Z/T plane?

pwalczysko:
> How much of a problem is it to have ROIs/Shapes loaded that aren't on the current Z/T plane?

It is not great in the sense that it is an ``inexplicable behaviour`` from users standpoint. For the smoothness of the workflow, I think it only becomes a problem when the video playing starts to be slow because of that.
I think we will profit from some longer videos which we have in https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-25919 where @jburel is planning to add the ROIs (tracks and points or elipses) to have a good judgement.

will-moore:
The video doesn't start to slow because of keeping Shapes in hand.
Just trying again locally using commit 7096856: keeping shapes in hand has some other benefits...
 - If you move back and forth between 2 or more planes, you don't have a delay to load shapes each time and you see them instantly on the viewer (actually, I still do the request - could probably avoid that).
 - Also, if I get rid of shapes that are not on current plane then we lose selection state: E.g select a shape, move to next Z/T then back again. The same shape should still be selected, but we lose the selection if we 'only' have shapes for the current plane loaded each time.
 - The main reason to load shapes per-plane is to make the loading faster, not because we only want to have the shapes for one plane loaded at at time.

So, if I can fix the other issue reported above, hopefully this will improve the experience. I think actually there's only one:
 - When you uncheck 'Show' on an ROI, then expand the ROI, all the newly loaded shapes should not be visible.

will-moore:
OK, so that commit ee00f4e should fix "When you uncheck 'Show' on an ROI, then expand the ROI, all the newly loaded shapes should not be visible."
I also avoid re-loading ROIs for a plane if we've already loaded them. This means we don't have "Loading Regions..." if we have already loaded them.

pwalczysko:
@will-moore studying https://merge-ci.openmicroscopy.org/web/webclient/img_detail/144215/?dataset=25919 (thanks @jburel for the ROIs), I can see

1. massive improvement, now I can watch single cells and judge whether or not that particular track is valid
2. the biggest drawback (comparably minor to Ad. 1):  As the shapes are ordered inside the ROIs by time, the low-times shapes are at the bottom, the high-time shapes are at the top. This seems unnatural to me (as I have to always start at the top, because have to expand the ROI to show the shape). Thus, I have to scroll towareds the bottom of the table to find the first shape and click on it (to have a starting point of my video play, where the cell is first detached) and then play the video. The re-ordering by clicking onto the "T" in the header of the column seems to get ignoerd, instead, it will close the expanded ROI on me if I click the ``T``

But as indicated, already now it is a massive improvement imho

pwalczysko:
> OK, so that commit [ee00f4e](https://github.com/ome/omero-iviewer/commit/ee00f4ea527cbfe0455103f28b34b3f9cc11d171) should fix "When you uncheck 'Show' on an ROI, then expand the ROI, all the newly loaded shapes should not be visible."
> I also avoid re-loading ROIs for a plane if we've already loaded them. This means we don't have "Loading Regions..." if we have already loaded them.

Both points above work as described. The unchecking of the ROI indeed removes the Shapes from display even as you are reloading the ROIs and shapes anew after a new plane was navigated. The planes which are re-visited do not get the "loading ROIs" message and wait.

plugin/omero_iviewer/views.py
```diff
@@ -276,18 +287,25 @@ def rois_by_plane(request, image_id, the_z, the_t, z_end=None, t_end=None,
...
+    roi_counts = get_shape_counts(conn, roi_ids)
+
+    for roi in marshalled:
+        if roi['@id'] in roi_counts:
+            roi['shape_count'] = roi_counts[roi['@id']]
+
     # Modify query to only select count() and NOT paginate
     query = get_query_for_rois_by_plane(the_z, the_t, z_end, t_end)
-    query = query.replace("distinct(roi.id)", "count(distinct roi.id)")
+    query = query.replace("select roi", "select count(distinct roi.id)")
```
joshmoore, line 307: This could be an argument to `get_query_...()`.

plugin/omero_iviewer/views.py
```diff
@@ -219,8 +219,8 @@ def persist_rois(request, conn=None, **kwargs):
     return JsonResponse(ret)
 
 
-def get_query_for_rois_by_plane(the_z=None, the_t=None, z_end=None,
-                                t_end=None, load_shapes=False):
+def get_query_for_rois_by_plane(the_z=None, the_t=None,
+                                z_end=None, t_end=None):
```
joshmoore, line 223: Does this count as a breaking change?

joshmoore (commented):
Thoughts from a first go through, also:
 - Looking at calls to `get_rois_by_plane` which take ~4s, am I right that this could best be next optimized by having a `get_rois_by_tile`?
 - After running a movie (I think with no ROI selected) under inspection in Chrome, then turning off the movie, then opening/closing/selecting ROIs and shapes, I had some odd behavior, almost like callbacks triggering after the fact.

plugin/omero_iviewer/views.py
will-moore: This method hasn't been released yet. It was introduced in https://github.com/ome/omero-iviewer/pull/342 which was merged after the last (0.10.1) release.

will-moore:
Having a `get_rois_by_tile()` would be a game-changer, particularly for Big images.
The lack of such a query was the reason that https://github.com/ome/omero-iviewer/pull/242 was blocked.
That PR also had other usability issues that need discussing - so we may want to take a different path in future, but certainly need the querying first.

plugin/omero_iviewer/views.py
joshmoore: :+1:

plugin/omero_iviewer/views.py
will-moore: Done in 8fa122c

will-moore:
@pwalczysko  You say "As the shapes are ordered inside the ROIs by time, the low-times shapes are at the bottom, the high-time shapes are at the top." but on https://merge-ci.openmicroscopy.org/web/webclient/img_detail/144215/?dataset=25919 I see to opposite:

Screenshot 2020-09-28 at 12 55 11

Sorting by T column has no effect since this sorts ROIs not shapes within an ROI. I guess it should do both, but the default Shapes order is at least in order of ascending T as I see it?

pwalczysko:
> @pwalczysko You say "As the shapes are ordered inside the ROIs by time, the low-times shapes are at the bottom, the high-time shapes are at the top." but on https://merge-ci.openmicroscopy.org/web/webclient/img_detail/144215/?dataset=25919 I see to opposite:
> 

> 
> Sorting by T column has no effect since this sorts ROIs not shapes within an ROI. I guess it should do both, but the default Shapes order is at least in order of ascending T as I see it?

No, not in my hands. See screenshot below from https://merge-ci.openmicroscopy.org/web/webclient/img_detail/144215/?dataset=25919 
Screen Shot 2020-09-28 at 15 20 07

But this is the same image you are pointing me to. Not sure why do we see different behaviour ?

will-moore:
For some reason, Firefox showed it in reverse. Not sure exactly why, but I think I fixed it.

pwalczysko:
https://github.com/ome/omero-iviewer/pull/351/commits/1f07bf716a82ac30d256c8de919acde09ce3106f works as expected now in Firefox, it indeed reversed the order of shapes

will-moore:
@jburel I think Petr is happy with this now, but you've not had a look in a while. Could you give it a go and see what you think needs doing? Thanks.

PR #359, opened 2020-11-25
validate zoomLevelScaling

### LINKED ISSUE #358: Resolutions must be in descending order
Some images, e.g see svs images (QA 29719/test.svs) from
https://forum.image.sc/t/omero-iviewer-shows-blank-for-some-images/44631/6
don't have each zoom level smaller than the last.

For that image the `zoomLevelScaling` is:
```
"zoomLevelScaling": {
    "0": 1,
    "1": 0.25,
    "2": 0.0625,
    "3": 0.0625
}
```

This causes an error in iviewer: https://openlayers.org/en/v5.3.0/doc/errors/#17 "Resolutions must be in descending order".  (old PanoJS viewer is OK).

In this case we can simply ignore any levels that are not smaller than the last.

will-moore:
Fixes #358. 

To test:
 - Import QA 29719/test.svs and open in iviewer.

### DIFF
--- a/src/viewers/viewer/Viewer.js
+++ b/src/viewers/viewer/Viewer.js
@@ -380,8 +380,15 @@ class Viewer extends OlObject {
         var dims = this.image_info_['size'];
         if (this.image_info_['zoomLevelScaling']) {
             var tmp = [];
-            for (var r in this.image_info_['zoomLevelScaling'])
-                tmp.push(1 / this.image_info_['zoomLevelScaling'][r]);
+            for (var r in this.image_info_['zoomLevelScaling']) {
+                var scale = 1 / this.image_info_['zoomLevelScaling'][r];
+                if (scale <= tmp[tmp.length - 1]) {
+                    // "Resolutions must be in descending order"
+                    // https://github.com/ome/omero-iviewer/issues/358
+                    break;
+                }
+                tmp.push(scale);
+            }
             zoomLevelScaling = tmp.reverse();
         }
         var zoom = zoomLevelScaling ? zoomLevelScaling.length : -1;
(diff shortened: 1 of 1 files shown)


manics (commented):
Imported the test image, and it's working today on merge-ci. I'll leave the code review to someone else.

src/viewers/viewer/Viewer.js
```diff
@@ -380,8 +380,15 @@ class Viewer extends OlObject {
         var dims = this.image_info_['size'];
         if (this.image_info_['zoomLevelScaling']) {
             var tmp = [];
-            for (var r in this.image_info_['zoomLevelScaling'])
-                tmp.push(1 / this.image_info_['zoomLevelScaling'][r]);
+            for (var r in this.image_info_['zoomLevelScaling']) {
+                var scale = 1 / this.image_info_['zoomLevelScaling'][r];
```
jburel, line 384: to avoid surprises, I think we need to add a check that ``this.image_info_['zoomLevelScaling'][r]`` is not ``0``

will-moore: These values come from the server via https://github.com/ome/omero-py/blob/99ff5fd87b327a774dc55613b5d639fce93f4eaf/src/omero/gateway/__init__.py#L8714.
So it is very unlikely we will get a `0` here, and if so then there is something very wrong with the image so we won't be able to provide an alternative value here that will allow the image to be viewed.
And I think it's a bit of an edge case to be implementing an error dialog etc.

jburel: could you add just a note ``unlikely to get 0`` coming from ...
so we know that is expected

jburel (approved):

Issue #362, opened 2020-12-16
Lack of license file


andor-pierdelacabeza:
Other ome repositories have a LICENSE file specifying the project license.
I can't see it here so I'm not sure if it's GPLv2 as some of the other repos.
Could it be included?

Thanks!

jburel:
There is a license file. 
Due to the layout of this repositories it is probably not where you were expecting it
https://github.com/ome/omero-iviewer/blob/master/plugin/LICENSE

PR #363, opened 2020-12-16
fix license file

jburel:
Fix header in the file
Nothing to test

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,1 @@
+plugin/LICENSE
--- a/plugin/LICENSE
+++ b/plugin/LICENSE
@@ -1,4 +1,4 @@
-NU AFFERO GENERAL PUBLIC LICENSE
+                      GNU AFFERO GENERAL PUBLIC LICENSE
                        Version 3, 19 November 2007
 
  Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>


sbesson:
As per #362 should the license be moved to the root of this repository? Files in other sub-directories than `plugin` refer to it and that would fix the auto-detection by GitHub.

jburel:
could do, the build will have to be reviewed since it is included in the manifest

sbesson:
Understood, As suggested by @manics, proposing to move the `LICENSE` to the top-level as this is the standard layout and add a symlink pointing to `../LICENSE` to prevent changes to the build.

joshmoore:
Or symlinking down?

will-moore:
When browsing https://github.com/jburel/omero-iviewer/tree/license the LICENSE symlink doesn't allow you to view the LICENSE, you just see https://github.com/jburel/omero-iviewer/blob/license/LICENSE. 
So I guess it depends how "user friendly" we need this to be.
Would it be better to move the LICENSE to the top level and sym-link to it there? Apologies if this is too nit-picky?

jburel:
It is due to the way symlinks are handled by git. 
```
Symbolic links are just another type of object in Git. Git saves the symlink (name and what it points to) but not the actual contents of the destination file/directory. When you checkout the commit, Git restores the symlink in your working directory.
```
I am reluctant to move the LICENSE up a folder and create a symlink since it will require a review of the build and we will have the same display on GitHub

sbesson:
Merging and we will see how GitHub detects this one

PR #369, opened 2021-02-18
Roi tests

will-moore:
Adding python integration tests for saving ROIs.

Looking to test that saving changes to Shapes doesn't unlink Annotations.
Initial tests seem to pass OK (annotations not lost).

Maybe need to reproduce this in the UI viewer first, before trying to replicate in tests?

cc @knabar @emilroz @jburel

### DIFF
--- a/test/python/integration/pytest.ini
+++ b/test/python/integration/pytest.ini
@@ -1,0 +1,6 @@
+[pytest]
+# pip install pytest-django
+DJANGO_SETTINGS_MODULE = omeroweb.settings
+markers =
+    broken: mark the test as broken, i.e. it may be intermittent or failing without a fully understood cause
+    fs_suite: group together FS tests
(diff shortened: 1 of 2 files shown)


will-moore:
@knabar @emilroz Can you give me steps on how to reproduce the ROI or Shape annotations being lost when saving in iviewer? Ideally I was trying to reproduce this in the tests (this PR) - I haven't tried in the viewer itself yet...

knabar:
@will-moore We have not been able to reproduce the issue, so there is a possibility it only happened in our dev environment with a certain combination of versions of the viewers etc.  We'll of course keep an eye on it, but right now everything looks fine.

For reference, we are currently using the following annotations:

* On the image:
  * XmlAnnotation
  * BooleanAnnotation
* On the ROI:
  * MapAnnotation
  * XmlAnnotation
* On the shape:
  * XmlAnnotation

PR #373, opened 2021-03-15
Load delta_t after image_data

### LINKED ISSUE #371: 512 timepoints takes long time to load settings
As reported by @sbesson on IDR, sample image with 512 timepoints takes 14 seconds to load the `Settings` right panel in iviewer.

will-moore:
Investigating solutions for slow loading of delta_t info.
Fixes #371.

This allows for loading of `delta_t` info to happen *after* the initial load of `image_data` and the update of the UI to show the image.

NB: `time.sleep(10)` used to simulate slow loading of delta_t, (reported as over 10 secs in issue above).
This slow loading should not cause any issues with the loading of other image data, or update of the UI.

To test (with fake delay temporarily added as above):

 - Open Time-lapse images in iviewer 
 - Should see that timestamp info isn't initially loaded (but this doesn't otherwise delay image opening, and being able to interact with the image)
 - Timestamp info should eventually be loaded and displayed after about 10 seconds
 - Switching between images faster than this shouldn't cause any issues (e.g. late loading of `delta_t` info doesn't get displayed on wrong image.)

If this all works fine, then I'll remove the `time.sleep(10)` and we can try testing on a real image which has slow loading of timestamp info.

cc @sbesson

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -280,30 +280,31 @@ a.ui-button:active,
 }
 
 .time-slider {
-    left: 90px;
-    right: 55px;
+    left: 0;
     z-index: 3;
+    display: flex;
+    right: 0;
 }
 .show-zSlider .time-slider {
-    left: 125px;
+    left: 35px;
 }
 
 .plane-slider span, .time-slider span {
-    position: absolute;
     font-size: 18px;
 }
 
-.plane-slider.glyphicon, .time-slider .glyphicon {
+.plane-slider.glyphicon {
     position: absolute;
 }
 
 .plane-slider span {
+    position: absolute;
     left: 0;
     width: 35px;
     text-align: center;
 }
 .time-slider span {
-    top: 6px;
+    margin: 6px;
 }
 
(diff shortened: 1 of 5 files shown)


pwalczysko:
in my hands, this works fine, as described. Possibly we can step forward with the removal of the ``time_sleep``. If @sbesson could point us to the image, I can import it and test on that.

pwalczysko:
Display problem of the timestamp on images with many timepoints. FF, Mac, fairly large screen, tried to expand the central pane in iviewer too, did not help. The  timepoints number is printed over by the timestamp, see screenshot.

Screenshot 2021-03-24 at 12 08 06

pwalczysko:
- the timestamp display is fixed
- the behaviour on the long timelapse is as expected

Ready to merge fmpov

sbesson (approved):
Same comment as https://github.com/ome/omero-iviewer/pull/377#issuecomment-851384803

PR #377, opened 2021-04-06
ignore missing projection-bytes config

### LINKED ISSUE #375: Compatible with OMERO.server 5.6.0
See https://github.com/IDR/deployment/pull/322#issuecomment-808046899

Need to check server version before trying to get server-side config.

cc @sbesson

will-moore:
Fixes #375.

Handle missing server config max_projection_bytes
Not found on OMERO.server before 5.6.1

To test:
 - deploy with OMERO.server 5.6.0

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -88,14 +88,21 @@ def index(request, iid=None, conn=None, **kwargs):
     params['ROI_PAGE_SIZE'] = ROI_PAGE_SIZE
 
     c = conn.getConfigService()
-    max_bytes = c.getConfigValue('omero.pixeldata.max_projection_bytes')
-    # check if MAX_PROJECTION_BYTES should override server setting
-    if max_bytes is None or len(max_bytes) == 0 or (
-            MAX_PROJECTION_BYTES > 0 and
-            MAX_PROJECTION_BYTES < int(max_bytes)):
-        max_bytes = MAX_PROJECTION_BYTES
-    else:
-        max_bytes = int(max_bytes)
+    max_bytes = None
+    try:
+        max_bytes = c.getConfigValue('omero.pixeldata.max_projection_bytes')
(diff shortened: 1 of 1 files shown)


sbesson (commented):
Tested on a pilot IDR server (with a 5.6.0 server). While OMERO.iviewer was previously failing with the last release, it now opens as expected.

Also tested on a couple of datasets with various Z sizes that the projection are enabled/disabled with the same criteria as current production IDR.

Using the developer tools, I can confirm that `MAX_PROJECTION_BYTES` is set to `"None"` which I assume is the expected behavior
```
            window.INITIAL_REQUEST_PARAMS['WEBGATEWAY'] = "/webgateway/";
            window.INITIAL_REQUEST_PARAMS['WEBCLIENT'] = "/webclient/";
            window.INITIAL_REQUEST_PARAMS['WEB_API_BASE'] = "/api/v0/";
            window.INITIAL_REQUEST_PARAMS['ROI_PAGE_SIZE'] = "500";
            window.INITIAL_REQUEST_PARAMS['MAX_PROJECTION_BYTES'] = "None";
```

The above made me wonder: where is the threshold logic now happening since `MAX_PROJECTION_BYTES` is set to `-1` and there is no server configuration specifying the maximum pixel size for projections?

will-moore:
I think that if there is no `max_projection_bytes` set anywhere, then iviewer will enable projection for non-tiled images below a plane size of 2048 x 2048 (see https://github.com/ome/omero-iviewer/pull/349).

sbesson (approved):
Thanks. At least from the IDR use case standpoint, this is good to merge and I could conceive an immediate iviewer release that we could consume in `prod96` (I can handle the release if everyone is happy with the suggestion) /cc @jburel @joshmoore

sbesson:
Discussed today. Merging to move towards a 0.11.0 release of OMERO.iviewer that we can deploy on a pilot IDR for testing.

PR #384, opened 2021-05-31
Changelog 0.11.0

sbesson:
Captures all the PRs currently merged and proposed for inclusion in the next minor release.
Suggestions on better wording welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+0.11.0 (June 2021)
+------------------
+
+- Delay the loading of the timestamp metadata [https://github.com/ome/omero-iviewer/pull/373](https://github.com/ome/omero-iviewer/pull/373)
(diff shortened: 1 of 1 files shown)


will-moore:
Looks good 👍

will-moore:
Just found https://github.com/ome/omero-iviewer/pull/372 which has been tested and fixes a user-reported bug.
Would be good to consider that for release. cc @jburel

sbesson:
While constructing this changelog, I looked at https://github.com/ome/omero-iviewer/pull/372 and was wondering whether it would be a candidate for this release. But from the last comment it looks like it opened a wider discussion which needs to be resolved about the indexing.

jburel:
I will not include it until we sort out the old viewer url.
Having 2 styles within the same framework will be extremely confusing.

sbesson:
Understood, I'll go ahead for `0.11.0` as it should unlock some upcoming IDR submission and leave the discussion to continue on  #372

PR #386, opened 2021-07-06
Time unit fixes

### LINKED ISSUE #385: Units in planeInfo
image from idr0113 do not have time information
The time is given in days so I used ``time.setUnit(omero.model.enums.UnitsTime.DAY)``
Unfortunately the units is not taken into account by the viewer and the time is indicated in seconds

In the screenshot below it should be 8 days not 8 seconds
Screenshot 2021-07-01 at 17 04 22

will-moore:
Fixes #385

This PR handles time values up to Days (previously, anything over 99 hours was not displayed correctly).
It also handles any time units.

Tested using script at https://github.com/ome/omero-iviewer/issues/385#issuecomment-874824752 to set units for existing image.

When doing this for a sample (FRAP) image, timestamps that are less than a day, no 'days' are shown:

Screenshot 2021-07-06 at 16 02 33

But when the time is longer than a day it is show like this (NB: open to suggestions for best way to format days)?

Screenshot 2021-07-06 at 16 01 58

If the time is an exact number of days, e.g. `3 days`, we still show `hh:mm:ss.sss`. Is it worth eliminating that if all those values are zero?

cc @jburel

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -494,19 +494,19 @@ def image_data(request, image_id, conn=None, **kwargs):
         px = image.getPrimaryPixels().getPhysicalSizeX()
         if (px is not None):
             size = image.getPixelSizeX(True)
-            value = format_value_with_units(size)
+            value = format_pixel_size_with_units(size)
             rv['pixel_size']['unit_x'] = value[0]
             rv['pixel_size']['symbol_x'] = value[1]
         py = image.getPrimaryPixels().getPhysicalSizeY()
         if (py is not None):
             size = image.getPixelSizeY(True)
-            value = format_value_with_units(size)
+            value = format_pixel_size_with_units(size)
             rv['pixel_size']['unit_y'] = value[0]
             rv['pixel_size']['symbol_y'] = value[1]
         pz = image.getPrimaryPixels().getPhysicalSizeZ()
         if (pz is not None):
             size = image.getPixelSizeZ(True)
-            value = format_value_with_units(size)
+            value = format_pixel_size_with_units(size)
             rv['pixel_size']['unit_z'] = value[0]
             rv['pixel_size']['symbol_z'] = value[1]
 
@@ -551,10 +551,11 @@ def delta_t_data(request, image_id, conn=None, **kwargs):
         for info in info_list:
             t_index = info.theT.getValue()
             if info.deltaT is not None:
-                value = format_value_with_units(info.deltaT)
-                timemap[t_index] = value[0]
+                secs = get_converted_value(info.deltaT, "SECOND")
+                timemap[t_index] = secs
                 if delta_t_unit_symbol is None:
-                    delta_t_unit_symbol = value[1]
+                    # Get unit symbol for first info only
+                    delta_t_unit_symbol = info.deltaT.getSymbol()
         for t in range(image.getSizeT()):
             if t in timemap:
                 time_list.append(timemap[t])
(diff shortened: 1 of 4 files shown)


jburel:
> If the time is an exact number of days, e.g. 3 days, we still show hh:mm:ss.sss. Is it worth eliminating that if all those values are zero?
My vote is yes
Also the precision ``ss.sss`` when hours or even minutes are displayed seems too much
From your screenshot it looks like the time is a negative value when no days are displayed

will-moore:
Hmmm - I'm not sure I agree. When my digital clock says 11 o'clock, it's `11:00` not `11`. Same for a stop-watch. The display doesn't change just because you hit a round number.

If move the T-slider from `1 day` to `1 day 00:00:00.001` then there's going to be a big shift in the length of the displayed time string, causing the slider to contract to provide more space.

If a user sees time displayed as `1 day` they won't know whether that's really "1 day 2 hours" or "1 day 1 second" rounded to `1 day`, or if it's precisely 1 day. Showing the full accuracy and being consistent across all time-points is much clearer.

If we take a more advanced strategy later, and we display times differently depending on the units on the server, then we could consider some other behaviours (as well as showing the server-side units), but that is a whole different change.

The negative values in that screenshot are correct. It's a FRAP movie where frames before the bleach event have negative timestamps.

I ran the script https://github.com/ome/omero-iviewer/issues/385#issuecomment-874824752 on this FRAP Image (user-3):
https://merge-ci.openmicroscopy.org/web/webclient/img_detail/83221/ and this DV: https://merge-ci.openmicroscopy.org/web/webclient/?show=image-122915

will-moore:
An example from IDR where this will affect the formatting: https://idr.openmicroscopy.org/webclient/img_detail/10340802/

`96:00:00.000` will be `4 days 00:00:00.000`.

sbesson:
Following up on the comment from https://github.com/ome/omero-iviewer/pull/386#issuecomment-878124142 and this morning's discussion, my biggest concern is about the custom nature of the proposed format. For usability, I agree it makes complete sense to convert the information into the most relevant unit i.e. 4 days is clearer and more explicit than 96h. It also has the benefit of keeping the time suffix compatible with ISO 8601 where the hour is bounded between 00 and 23.

My main issue is that the proposed representation mixes unit written as words and unit separators which I found confusing. I would have almost expect something `4 days 4 hours 2 minutes 10 seconds`. Otherwise, would a truncated version of the ISO standard i.e. `dd hh:mm:ss[.mmm]` be an option?

jburel:
``4 days 4 hours 2 minutes 10 seconds`` will be too long to fit in the available space.
So I think the suggested truncated version of the ISO standard is probably the most appropriate one

will-moore:
I've fixed the formatting, adding a tooltip to improve clarity:

Screenshot 2021-07-13 at 16 15 31

Needed to change a bit more code, so please test values shown while sliding, beside the slider AND tooltip.

jburel:
Looking at https://merge-ci.openmicroscopy.org/web/webclient/img_detail/83221/
If possible we could be a be smarter in the tooltip ``1 days`` always makes me jump

will-moore:
@jburel Done

jburel:
I have installed this PR in pilot-0113
The time is now displayed in days instead of seconds.
Screenshot 2021-07-21 at 15 24 07

jburel:
The tooltip still displays 0hours but this can be fixed in a follow-up PR

jburel:
> https://idr.openmicroscopy.org/webclient/img_detail/10340802/

idr 
Screenshot 2021-07-21 at 15 34 20

pilot with this PR
Screenshot 2021-07-21 at 15 34 29

jburel:
suggest to merge it and release as 0.11.1

jburel:
Merging and tagging

PR #399, opened 2022-02-16
Update to Django 3.2

will-moore:
To correspond to https://github.com/ome/omero-web/pull/356


chris-allan:
This is similar to what we've had to do with PathViewer and some of our other plugins. It has the added advantage of also being compatible with Django 1.11.x so can proactively go into a release allowing the support of both 1.11.x and 3.2.x assuming no other changes are required.

PR #403, opened 2022-02-28
pin omero-web

jburel:
Pin omero-web
Proposing to tag as 0.11.2
cc @will-moore @sbesson

### DIFF
--- a/plugin/setup.py
+++ b/plugin/setup.py
@@ -118,7 +118,7 @@ setup(name="omero-iviewer",
       url="https://github.com/ome/omero-iviewer/",
       download_url='https://github.com/ome/omero-iviewer/archive/v%s.tar.gz' % version,  # NOQA
       keywords=['OMERO.web', 'plugin'],
-      install_requires=['omero-web>=5.7.0'],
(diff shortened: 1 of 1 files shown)


will-moore:
I assume we only need this if we're going to release `omero-web` with Django 3.2 *before* we release compatible iviewer https://github.com/ome/omero-iviewer/pull/399. But I don't see any reason to do that. The new `omero-web` won't be usable by anyone who wants to use iviewer, and the compatible iviewer should be ready by the time `omero-web` is released.

Same for all the other apps I think?

jburel:
Possible scenario:
* an app that has not been updated to support django 3.2 is installed
* user wants iviewer and omero-web with django 3.2 support has been released
* everything breaks if we do not cap

will-moore:
My main concern is the trouble we would cause by releasing `omero-web 6.0.0` without compatible iviewer, figure etc, because that would also break everything.
Are we committed to avoiding that scenario?

will-moore:
As discussed this morning, steps will be https://github.com/orgs/ome/projects/21#card-78410459
and I believe we agreed on `omero-web 6.0.0` for the Django 3.2 upgrade.

PR #405, opened 2022-03-25
Fix X close button for ROI popup.

will-moore:
The unicode sometimes renders incorrectly, possible when font loading is slow??

Fix with tip from https://twitter.com/wesbos/status/499245255703949312

Bug: *sometimes* the X looks like this. Doesn't happen very often
Screenshot 2022-03-25 at 11 40 03
Mostly looks like this:
Screenshot 2022-03-25 at 11 41 23

### DIFF
--- a/css/ol3-viewer.css
+++ b/css/ol3-viewer.css
@@ -490,9 +490,6 @@
 .shape-edit-popup-closer {
     text-decoration: none;
     position: absolute;
-    top: 2px;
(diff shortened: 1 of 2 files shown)


jburel:
It is not affecting the functionality. 
Looking at it on Chrome, it is thinner that the version before but it is fine.
I could not reproduce the original issue but this change is not a regression :-)

Merging

PR #408, opened 2022-05-19
Add note on calling pip

### LINKED ISSUE #407: Installation documentation
Hej,

This is just a small recommendation for the [documentation of the installation procedure](https://github.com/ome/omero-iviewer/blob/master/plugin/omero_iviewer/README.rst). It might be very obvious after you are working with OMERO for a while but not so much at the beginning. I would add that the installation of the omero-iviewer should be done from root using the pip of the OMERO.web venv. Something along the lines:

Run as root:
$VENV_WEB/bin/pip install -U omero-iviewer  
where in my case the VENV_WEB=/opt/omero/web/venv3

Then just add that all other steps are done from the omero-web user.

Best regards,
Rafa

will-moore:
Fixes #407 

cc @CamachoDejay @jburel

### DIFF
--- a/plugin/omero_iviewer/README.rst
+++ b/plugin/omero_iviewer/README.rst
@@ -21,7 +21,11 @@ Installing from PyPI
 
 This section assumes that an OMERO.web is already installed.
 
-Install the app using `pip <https://pip.pypa.io/en/stable/>`_:
+Install the app using `pip <https://pip.pypa.io/en/stable/>`_.
+
+NB: You need to ensure that you are running `pip` from the python environment
+where `omero-web` is installed. Depending on your install, you may need to
+call `pip` with `/path/to/venv3/bin/pip install ...`:
 
 ::
 
--- a/plugin/omero_iviewer/README.rst
+++ b/plugin/omero_iviewer/README.rst
@@ -21,11 +21,11 @@ Installing from PyPI
 
 This section assumes that an OMERO.web is already installed.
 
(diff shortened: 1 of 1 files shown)


jburel:
I think we should add information to the OMERO.web installation pages too
since the same will apply to all the apps

will-moore:
@jburel If you're happy with the change here, I'll apply the same to `omero-figure`?

jburel:
the formatting is not good i.e. it is not displayed as a code snippet.
Also ``venv3`` is an example.

will-moore:
@jburel How's that?

jburel:
We have a general section about web app 
e.g. https://docs.openmicroscopy.org/omero/5.6.4/sysadmins/unix/install-web/walkthrough/omeroweb-install-ubuntu1804-ice3.6.html
Maybe we should show in the doc in the installation of the "popular" ones in the walkthrough

will-moore:
@jburel Yes, it would be nice to have, but there's a trade-off with maintaining install info for figure/iviewer in more than 1 place.
Maybe could add the core install commands in the docs:
```
$ pip install -U omero-iviewer
$ omero config append omero.web.apps '"omero_iviewer"'
```
but still link to the apps themselves for full instructions? Although that might also be prone to users missing the full instructions? 🤷

joshmoore:
As a side note, in some installation paths, pip doesn't get (or at least, previously wasn't getting) added as a binary. In those cases, `python -m pip` tends to work.

PR #420, opened 2022-08-16
Add release process to README

will-moore:
Will be helpful since we don't use bumpversion here.

cc @jburel


jburel:
We should unify with other repositories and use bump2version for version bump.
Pushing the tag will trigger the GitHub action workflow pushing to pypi

PR #443, opened 2023-01-13
Don't scale labels by default

### LINKED ISSUE #442: Tooltip with shape comment displayed on mouse hover even if "Show comments" is not checked
I don't know what is the expected behavior for the "Show comments" checkbox on the "ROIs" tab and I couldn't find something relevant in the code:

Here https://github.com/ome/omero-iviewer/blob/master/src/viewers/viewer/source/Regions.js#L114 it says, that this flag should determine if texts for shapes other than labels should be displayed. Is "text" == "comment" in this context?

The only place where this flag is used is here https://github.com/ome/omero-iviewer/blob/master/src/viewers/viewer/utils/Style.js#L258, but I don't see how this effects the visibility for comments of shapes in the viewer.

I noticed, that comments for shapes (other than labels) are displayed in a small tooltip on mouse hover if the shape is not selected.

Now I would expect that those mouse hover tooltips are only displayed on mouse over if "Show comments" is checked. That is not the case at the moment. The tooltips are displayed always.

I added a check for `this.regions_.show_comments_` here https://github.com/ome/omero-iviewer/blob/master/src/viewers/viewer/interaction/Hover.js#L88 and now it works as expected. Should I create a pull request for this?

will-moore:
Fixes #442.

This adds a config option to disable the scaling of labels and shape-text when you zoom:

```
omero config set omero.web.iviewer.scale_text false
```

Often the behaviour of NOT scaling labels as you zoom (typical behaviour on maps etc) is preferable.

But that would be a breaking change - E.g. in this case, all the labels are 200px high, and if shown that large at lower zoom levels then they would obscure the image (and each other): https://chamber.mrc.ox.ac.uk/iviewer/?images=124

For testing, the above config has been set-up on merge-ci:
 - Add text to shapes and a Text label
 - Zooming in and out shouldn't change the size of the label - always shown on screen at 100%

### DIFF
--- a/plugin/omero_iviewer/iviewer_settings.py
+++ b/plugin/omero_iviewer/iviewer_settings.py
@@ -25,6 +25,14 @@ from omeroweb.settings import process_custom_settings, report_settings
 # load settings
 IVIEWER_SETTINGS_MAPPING = {
 
+    "omero.web.iviewer.scale_text":
+        ["SCALE_TEXT",
+         True,
+         str,
+         # NB: use str() here because bool("False") is True!
+         # whereas str() allows setting this to "False"
+         ("Scale ROI text labels when zooming")],
+
     "omero.web.iviewer.max_projection_bytes":
         ["MAX_PROJECTION_BYTES",
          -1,
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -46,6 +46,7 @@ from . import iviewer_settings
 WEB_API_VERSION = 0
 MAX_LIMIT = max(1, API_MAX_LIMIT)
 
+SCALE_TEXT = getattr(iviewer_settings, 'SCALE_TEXT')
 ROI_PAGE_SIZE = getattr(iviewer_settings, 'ROI_PAGE_SIZE')
 ROI_PAGE_SIZE = min(MAX_LIMIT, ROI_PAGE_SIZE)
 MAX_PROJECTION_BYTES = getattr(iviewer_settings, 'MAX_PROJECTION_BYTES')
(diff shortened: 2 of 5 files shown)


pwalczysko:
Trying to compare 
https://merge-ci.openmicroscopy.org/web/webclient/img_detail/211615/?dataset=50619
and
https://latest-ci.openmicroscopy.org/web/webclient/img_detail/85715/?dataset=43469

Not sure, maybe need an detailed explanation what is the difference in the two scalebar behaviours when zooming those images ?

will-moore:
Sorry, that last commit won't be deployed on merge-ci yet.
And this PR has nothing to do with scalebar.
It's about how text labels scale when you zoom an image.
Default behaviour is to zoom the text when you zoom. This PR and merge-ci is configured to disable that.

pwalczysko:
> Sorry, that last commit won't be deployed on merge-ci yet. And this PR has nothing to do with scalebar. It's about how text labels scale when you zoom an image. Default behaviour is to zoom the text when you zoom. This PR and merge-ci is configured to disable that.

Ah, sorrry, I can see it now. Yes, the text does not zoom on an image in merge-ci.

pwalczysko:
Problem:

1. Zoom out on an svs image to 30%. 
2. Draw a Text label which approx. as big as the size of the white area in the top-left corner of the screenshot below
3. Observe that once you have finished drawing, the label jumps in size to be more than twice as big as you intended

Screenshot 2023-02-06 at 14 50 25

pwalczysko (changes requested):
Please address the drawing of new labels problem.

will-moore:
Couldn't find a fix for that problem despite a few different attempts.

PR #460, opened 2023-11-09
Channel min spinner fix

will-moore:
Fixes issues for 'start' channel slider input as reported at https://github.com/ome/ZarrReader/issues/67#issuecomment-1794622742.

 - With a very large negative number, the formatting of the number as required for floats was resulting in Nan.
 - Also, the spinner behaviour (jQuery plugin) doesn't work with very large negative numbers. Reducing the minimum number fixes this issue.

### DIFF
--- a/src/model/image_info.js
+++ b/src/model/image_info.js
@@ -896,10 +896,11 @@ export default class ImageInfo {
                     precision <= 0) precision = 3;
         } else precision = 0;
 
+        if (precision != 0 && this.image_pixels_type === 'float') {
+            start_min = Misc.roundAtDecimal(start_min, precision)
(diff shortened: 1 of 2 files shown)


jburel:
is the change deployed on idr-testing?

jburel:
Does this PR need to be considered for the upcoming release?

will-moore:
I tried to find images where this bug was still showing, so as to demonstrate the fix, but e.g. with the image reported originally I don't see this issue any more: https://idr-testing.openmicroscopy.org/webclient/img_detail/10648053/

It's possible that some other dependency update addressed the issue? 🤷

PR #463, opened 2024-02-14
Fix build

jburel:
This PR
* updates action version to avoid warning
* fixes flake8 issue
* fixes build failure due to an older version of nodejs being used. Nodejs is now at version 20
* uses ``dnf`` instead of ``yum``


will-moore (approved):
Build is green, thx 👍

PR #465, opened 2024-03-14
Save as Figure

will-moore:
To test:

- Open 1 or more images, zoom, pan, change rendering settings and add ROIs.
- File > Save Viewport(s) as Figure
- Enter a name for figure
- Click on link in the success dialog -> Open figure in new tab
- Only the ROIs (Shapes) visible at the time of export (and intersect with the viewport) will be included in the Figure.
- Pixel sizes, timestamps, Dataset names etc should be included in the figure

Screenshot 2024-03-19 at 12 34 15

Screenshot 2024-03-19 at 12 35 08

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -18,7 +18,7 @@
 from django.shortcuts import render
 from django.http import JsonResponse, Http404
 from django.conf import settings
-from django.urls import reverse
+from django.urls import reverse, NoReverseMatch
 
 from os.path import splitext
 from collections import defaultdict
@@ -84,6 +84,11 @@ def index(request, iid=None, conn=None, **kwargs):
     # we add the (possibly prefixed) uris
     params['WEBGATEWAY'] = reverse('webgateway')
     params['WEBCLIENT'] = reverse('webindex')
+    try:
+        params['OMERO_FIGURE'] = reverse('figure_index')
+    except NoReverseMatch:
+        # omero-figure not installed
+        pass
     params['WEB_API_BASE'] = reverse(
         'api_base', kwargs={'api_version': WEB_API_VERSION})
     if settings.FORCE_SCRIPT_NAME is not None:
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -508,18 +508,22 @@ def image_data(request, image_id, conn=None, **kwargs):
             value = format_pixel_size_with_units(size)
             rv['pixel_size']['unit_x'] = value[0]
             rv['pixel_size']['symbol_x'] = value[1]
+            # id e.g. 'MICROMETER' is used for export to OMERO.figure
+            rv['pixel_size']['unit_id_x'] = value[2]
         py = image.getPrimaryPixels().getPhysicalSizeY()
         if (py is not None):
             size = image.getPixelSizeY(True)
             value = format_pixel_size_with_units(size)
             rv['pixel_size']['unit_y'] = value[0]
             rv['pixel_size']['symbol_y'] = value[1]
+            rv['pixel_size']['unit_id_y'] = value[2]
         pz = image.getPrimaryPixels().getPhysicalSizeZ()
         if (pz is not None):
             size = image.getPixelSizeZ(True)
             value = format_pixel_size_with_units(size)
             rv['pixel_size']['unit_z'] = value[0]
             rv['pixel_size']['symbol_z'] = value[1]
+            rv['pixel_size']['unit_id_z'] = value[2]
 
         delta_t_unit_symbol = None
         rv['delta_t_unit_symbol'] = delta_t_unit_symbol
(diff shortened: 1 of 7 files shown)


pwalczysko:
Bug: The panel in the newly created figure is shifted with respect to the ROIs.

1. Create a "figure from viewport" with couple of precisely drawn freehand polygon ROIs (e.g. "segmentation" of nuclei in a large image) - you can either draw these ROIs manually at the time of opening the image in iviewer, or have them already drawn and saved, or, draw them and save these ROIs
2. In the newly created "figure from viewport" - when inspected in OMERO.figure, observe that the ROIs which were in the viewport when the "figure from viewport" was created are positioned as expected.
3. Still in OMERO.figure, select the panel, go to Labels tab, click Edit in the ROI section and add the ROIs which are saved from step 1. above in OMERO (these are, or rather should be, the same ROIs which you just  transported onto your figure using the new "figure from viewport" feature in iviewer
4. Observe that the ROIs which you are newly adding now are slightly shifted, always towards bottom right - see screenshot.
5. Save your figure, close it. Reopen it.
6. Repeat the addtion of the ROIs step in OMERO.figure as per Ad 3. above - observe again that the newly added ROIs are shifted one more time towards bottom right.
Screenshot 2024-04-04 at 12 30 03

pwalczysko:
RFE:

Some misformatting of the dialog - note the inverted commas are only on one side of the ID (there probably should not be any inverted commas)

Screenshot 2024-04-04 at 11 40 19

pwalczysko (changes requested):
Two problems noted, the shift of the pane with respect to ROI position probably needs a full rethink.

pwalczysko:
To comment further on the ROI-shifting problem https://github.com/ome/omero-iviewer/pull/465#issuecomment-2036930397 - this behaviour is present only when working with figures created by the feature implemented in the present PR https://github.com/ome/omero-iviewer/pull/465 - the figures which are created within OMERO.figure itself do not have any problems with shifted ROIs, i.e. the ongoing work on https://github.com/ome/omero-figure/pull/477 does not seem to be cause of this, cc @jburel

will-moore:
@pwalczysko thanks for the testing. Re: https://github.com/ome/omero-iviewer/pull/465#issuecomment-2036930397 - I think that what is going on here is that OMERO.figure is trying not to add duplicate ROIs on top of one-another. If a Shape already exists when you try to add one, the Shape is automatically offset to the lower-right. In the same way that when you copy and paste a Shape it is offset to the lower-right. If you first delete the Shapes in OMERO.figure before loading and adding them, you should see that they are added in the right place.

pwalczysko:
I will wait for tomorrow's build for retest of the double-quotes https://github.com/ome/omero-iviewer/pull/465/commits/efda99710532f9743896717e88f0e68ca23ec3b2

pwalczysko:
Double-quotes are fixed.

The only remaining issue is that points are ignored.
As this is done without warning, I wonder if we should add one ?
If points are not supported in Figure, then this is fine... but still they are supported in iviewer - if someone gets into great lenght and creates a lot of points, then they might be unpleasantly surprised not to find them in figure later...

pwalczysko (changes requested):
https://github.com/ome/omero-iviewer/pull/465#issuecomment-2051806129

will-moore:
@pwalczysko Points are now converted to Ellipses in figure...

Screenshot 2024-04-15 at 17 59 47

pwalczysko:
> @pwalczysko Points are now converted to Ellipses in figure...

I am really not sure about this. If somebody has already some Ellipses (which were meant to be ellipses) on an image, and adds points, then, in Figure, they get a surprise.

Maybe we can discuss in a meeting tomorrow - my preference would be to drop the points if there are any and warn the user. Offering some distortion of the ROIs by default is not the best practice imho.

will-moore:
@jburel Just to clarify, do we need to have Points support in OMERO.figure before this PR is approved?

@pwalczysko suggested at https://github.com/ome/omero-iviewer/pull/465#issuecomment-2051806129 that a warning might be enough, although I think he's changing his mind?!

If we need to wait for Points in OMERO.figure, that needs the Vite PR to be merged, then a follow-up PR. Then iviewer will depend on Points support in OMERO.figure, so we'll need to only enable "Save as Figure" if we have the most recent version of Figure. This will have to be detected in some manual way (we don't want to add omero-figure as a python dependency of iviewer).
Currently we detect if Figure is installed and don't show the "Save As Figure" option if it's not found, but detecting the version will be a bit harder.

jburel:
While looking at it, I don't think it will solve we are currently facing with big images
* The viewport is then a "crop" in figure, preventing the user to pan to a new location if a small mistake was made
* The "context" of that viewport is also not present. So the user does not know the part of the image the crop is related to.

I feel that going back and forth between the apps to manipulate the images is not going to work for big images unfortunately

will-moore:
@jburel There is no cropping here. All panels in Figure can be zoomed and panned over the whole image.
You may get the impression of a crop when you pan the viewport, but that is just because the rendered region is restricted around the viewport and doesn't cover the whole image. When you pan and drop, it re-renders around the new viewport region.

It's possible to create an Inset panel in the normal way (maybe even with a shortcut in the new workflow) to show where the viewport is, but that is outside the scope of this PR.

will-moore:
Tested again with OMERO.figure now supporting Points:

Screenshot 2025-03-11 at 10 05 14

pwalczysko:
lgtm, including the points, thank you

pwalczysko (approved):

PR #464, opened 2024-02-22
Tile load error warning

will-moore:
As discussed at IDR meeting, instead of failing silently and showing white background when tiles fail to load, we should warn the user and encourage them to refresh the viewer since tiles may load on 2nd attempt.

Screenshot 2024-02-22 at 15 00 22

EDIT: also added a spinner when the map loads, based on example at https://openlayers.org/en/latest/examples/load-events.html

To test,
 - Open an Image in ivewer - should see a spinner on the map until the last tile has loaded.
 - Go back to webclient and logout
 - Zoom / pan or adjust Z/T or rendering settings to try and load more tiles - should see error above.

PR #456, opened 2023-06-27
Don't use tiles for non-big movies

will-moore:
Fixes #453.

If the image is not a Big (tiled) image, then don't use tile-based loading if the image is a Movie.

PR #470, opened 2024-04-17
Movie playing waits for planes

will-moore:
This replaces (builds on) #464 and #456 and also supports movie playing waiting on image loading.

These PRs are combined to avoid conflicts, since the functionality is all related. Also makes it easier to deploy.

 - Tile loading errors are reported with message (as already tested in #464)
 - A spinner is shown while tiles/planes are loaded (as already tested in #464)
 - For "large" movies (between 2k x 2k and 3k x 3k) load whole planes instead of tiles (not tested yet, see #456)
 - When playing a movie, we wait at least 250 millisecs (default delay between frames) but we wait longer if we are still waiting for the image to render from a previous update.

Questions:
 - When we are playing a movie, do we want to show a spinner? If the rate is limited by rendering speed, then the spinner will **always** be shown. Even if rendering is quite fast, we will still see the spinner a lot of the time.
 - Do we want to change the default delay between frames? 250 millisecs is quite short.
 - What criteria do we use for deciding to load tiles? (see discussion in comment below)

To test:
 - Movies should be rendered as a single plane, unless they are really BIG (over 3k * 3k)
 - Play movie, stop/start lots. Check that we don't get in a bad state. 
 - Movie should stop playing when it gets to the end. Clicking start again should start it from the beginning.

### DIFF
--- a/css/app.css
+++ b/css/app.css
@@ -290,7 +290,7 @@ span.sp-thumb-el {
     position: absolute;
 }
 
-.cursor_spinner {
+.cursor_spinner * {
     cursor: wait;
 }
 
--- a/css/app.css
+++ b/css/app.css
@@ -551,6 +551,28 @@ span.sp-thumb-el {
     height: 100%;
 }
 
+@keyframes spinner {
+    to {
+        transform: rotate(360deg);
+    }
+}
+
+.map_spinner:after {
+    content: "";
+    box-sizing: border-box;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    width: 40px;
+    height: 40px;
+    margin-top: -20px;
+    margin-left: -20px;
+    border-radius: 50%;
+    border: 5px solid rgba(180, 180, 180, 0.6);
+    border-top-color: rgba(0, 0, 0, 0.6);
+    animation: spinner 0.6s linear infinite;
+}
+
 .thumbnail-panel {
     padding-top: 30px;
     width: 100px;
--- a/css/app.css
+++ b/css/app.css
@@ -290,6 +290,14 @@ span.sp-thumb-el {
     position: absolute;
 }
 
+.cursor_spinner {
+    cursor: wait;
+}
+
+.cursor_spinner .dim-arrows {
+    cursor: wait;
+}
+
 .plane-slider {
     bottom: 115px;
     top: 45px;
--- a/src/controls/dimension-slider.js
+++ b/src/controls/dimension-slider.js
@@ -120,14 +120,16 @@ export default class DimensionSlider {
             this.player_info.forwards === forwards;
 
         // make history entry for stop
-        if (stop)
+        if (stop) {
             conf.addHistory({
                 prop: ['image_info', 'dimensions', this.dim],
                 old_val : this.last_player_start,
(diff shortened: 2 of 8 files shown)


joshmoore:
A few thoughts/questions looking at IDR's image 4007801:
 - can a lower resolution be used if loading is taking too long?
 - rather than blocking the view, could a subtle loading bar (e.g., under the image) be used?
 - is any pre-loading happening? it seems like a *little* would still be useful.
 - does "movie" also include progressing through z stacks, e.g.?

will-moore:
Looking for options for loading lower resolution layers in OpenLayers...
https://stackoverflow.com/questions/43538345/how-to-force-load-tiles-for-lower-resolution talks only about adding lower resolution layers to the pyramid. It mentions "preload" option:
https://openlayers.org/en/latest/examples/preload.html

I tested this, but when playing a movie (in this PR as it stands, we simply try to load the lower resolution(s) AND the correct resolution at the same time for each timepoint (as expected, it doesn't pre-load other timepoints and doesn't avoid loading full current resolution):
```
diff --git a/src/viewers/viewer/Viewer.js b/src/viewers/viewer/Viewer.js
index 3a1843a..21bb4ff 100644
--- a/src/viewers/viewer/Viewer.js
+++ b/src/viewers/viewer/Viewer.js
@@ -590,7 +590,7 @@ class Viewer extends OlObject {
             interactions: interactions,
-            layers: [new Tile({source: source})],
+            layers: [new Tile({source: source, preload: 1})],
             target: this.container_,
```

will-moore:
Also looked at a potential 'hack' of loading tiles from the lower resolution by updating the tile url function, but the lower resolution tiles are not rescaled to the appropriate size by OpenLayers, so the don't fill the space:

```
+++ b/src/viewers/viewer/source/Image.js
@@ -265,7 +265,7 @@ const OmeroImage = function(options) {
                     this.tileGrid.resolutions_.length - tileCoord[0] - 1 : 0;
                 if (this.tiled_) {
-                    url += 'tile=' + zoom  + ',' +
+                    url += 'tile=' + (zoom + 1)  + ',' +
                         tileCoord[1] + ',' + (-tileCoord[2]-1);
```

will-moore:
deployed those changes onto idr-testing

jburel:
I don't see any change when playing movie across Z

jburel:
Attempting to play https://idr-testing.openmicroscopy.org/webclient/img_detail/10647409/?dataset=11901
It took a while for the second frame to come so I stopped the movie but the spinner is still displayed in that case. I was expecting the spinner to be removed and the loading being cancelled

will-moore:
@jburel I think you're not seeing my changes? It took a few minutes for my deployment script on idr-testing to complete, and so I probably commented a bit prematurely above, or possibly something got cached.
It's working for me but I'll also try cache busting...

Could you give it another go?

Handling the Stop button when the plane is still loading is a bit tricky. We don't really have a way to cancel the image plane request except to undo the last Z/T increment (IF the plane hasn't loaded yet).

jburel:
Much nicer thank you
have the last 2 commits been deployed too? they were pushed after this morning discussion

pwalczysko (approved):

PR #471, opened 2024-04-18
Z projection disabled tooltip

will-moore:
See https://forum.image.sc/t/z-projections-disabled/95042/7

Discussion above shows that we don't give any feedback about why projection is disabled. Users can't tell if the max projection bytes config is being taken into account or if projection is disabled for some other reason.

This adds a useful tooltip to provide info on why projection is disabled (movie is playing or image is tiled or image bytes greater than projection:

Screenshot 2024-04-18 at 12 22 11

### DIFF
--- a/src/controls/dimension-slider.html
+++ b/src/controls/dimension-slider.html
@@ -37,7 +37,7 @@
         </span>
         <span class="dim-projection
                      ${getZProjectionDisabled(player_info.handle, player_info.forwards) ? 'disabled-color' : ''}"
-              title="${getZProjectionDisabled(player_info.handle, player_info.forwards) ? 'Projection disabled' :
+              title="${getZProjectionDisabled(player_info.handle, player_info.forwards) ? getZProjectionTooltip() :
                             image_config.image_info.projection === INTMAX ?
                                 'Turn off projection' : 'Turn on projection'}"
               css="${image_config.image_info.projection === INTMAX ?
--- a/src/controls/dimension-slider.js
+++ b/src/controls/dimension-slider.js
@@ -527,6 +527,26 @@ export default class DimensionSlider {
         return (handle !== null && forwards || tiled || stack_size > proj_limit);
     }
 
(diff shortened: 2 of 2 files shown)


pwalczysko:
The tooltip works on merge-ci and imho is useful.

Nevertheless, could you please explain what does the message in the tooltip attempt to say ?
I genuinely understand only that "some" parameter is bigger than the cutoff for the projection showing.
RFE for the tooltip:

1. Use much simpler wording - what are bytes ? Maybe get rid of this ?
2. Do not use brackets
3. Do not use ``x size X x sizeY...`` - very hard to read and understand on a small font on a tooltip - the ``x`` in first case means "times", in second case it is the sizeX - very unfriendly, hard to read and understand
4. Can a small paragraph be made in the tooltip so that the equation is not over two lines ? (but also see ad 1. above, is the complex equation necessary ?)

Thank you

pwalczysko (changes requested):
Please simplify&modify the content of the tooltip as per https://github.com/ome/omero-iviewer/pull/471#issuecomment-2093250668

will-moore:
This PR aims to address issues on the forum discussion (see description link above). It's mostly meant for a sysadmin to setup their system but it would also be useful for a user who wants to ask their sysadmin to update their config to enable projection for a particular image:
 
 - Admin doesn't know if their `omero.pixeldata.max_projection_bytes` config has been taken into account when calculating the threshold or if a lower threshold might be in place from iviewer config, so we want to include that number in the tooltip.
 - User forgot that `sizeC` is taken into account in this calculation (NB: sizeT makes no difference)
 - Also, users may not know how many bytes per pixel for the current pixelType although I guess we can omit this.

It's tricky to explain to regular users what `bytes` are or to remove them from the tooltip completely since that's what the message is all about.

How about:

```
Projection Disabled: Image size XYZ and Channels exceeds the max_projection_bytes limit of 268,435,456.
```

will-moore:
@pwalczysko - how about the suggested change in my previous comment?

PR #474, opened 2024-04-24
Img data error handling

### LINKED ISSUE #473: Handle errors loading image metadata imgData JSON
Various errors in Rendering Engine, or corrupt images etc can cause the imgData JSON loading to fail completely or to return incomplete JSON.
In most cases, errors are not shown to the user and so they are left waiting forever.

Need appropriate error messages to handle:
 - E.g. 500 server error
 - Incomplete metadata
 - ConcurrencyException / Missing Pyramids etc.

will-moore:
Fixes #473 

This improves the error handling for loading image metadata.
Previously we only handled `404 Image not found`, but if any JSON data was returned, even with exceptions, then these were ignored and the viewer would silently fail.

I have handled the possible Errors etc that are reported by the python error handling at various stages, such as ConcurrencyException and other catch-all exceptions:

Screenshot 2024-04-24 at 16 51 31

Screenshot 2024-04-24 at 16 52 33

To test, you need to generate or find broken images (or missing pyramids etc).

Most of the testing I've done has been by artificially adding errors at particular points in the python code.

### DIFF
--- a/src/model/image_info.js
+++ b/src/model/image_info.js
@@ -19,6 +19,7 @@
 import {noView} from 'aurelia-framework';
 import Misc from '../utils/misc';
 import Ui from '../utils/ui';
+var escapeHtml = require('escape-html')
 import {
     APP_TITLE, CHANNEL_SETTINGS_MODE, INITIAL_TYPES, IVIEWER,
     PROJECTION, REQUEST_PARAMS, WEBCLIENT, WEBGATEWAY
@@ -318,6 +319,13 @@ export default class ImageInfo {
                     this.image_id = response.id;
                 }
 
+                // validate response
+                // check for Exceptions and show error dialog.
+                const valid = this.validateImageInfo(response);
+                if (!valid) {
+                    return;
+                }
+
                 // read initial request params
                 this.initializeImageInfo(response, refresh);
                 // check for a parent id (if not well)
@@ -358,6 +366,39 @@ export default class ImageInfo {
                 Ui.showModalMessage(errMsg, 'OK');
             }
         });
+    }
+
+    /**
+     * Checks that the imgData JSON response is valid, no exceptions etc.
+     *
+     * @private
+     * @param {Object} response the response object
+     * @memberof ImageInfo
+     * @return {Boolean} True if data is valid. Also shows dialogs with errors
+     */
+    validateImageInfo(response) {
+
+        if (response.ConcurrencyException) {
+            Ui.showModalMessage(`<p>Image is not currently viewable</p>
+                <pre>ConcurrencyException</pre>
+                <small>A pyramid of zoom levels is not available. Generation may be in progress if this is enabled on your server.</small>`, "OK");
+            return false;
+        }
+
+        if (response.error || response.Exception) {
+            const msg = response.error || response.Exception;
+            Ui.showModalMessage(`<p>Error loading Image data</p>
+                <pre>${escapeHtml(msg)}</pre>`, "OK");
+            return false;
+        }
+
+        if (response.channels == undefined || response.channels.length === 0) {
+            Ui.showModalMessage(`<p>No channel data loaded</p>
+                <pre>${escapeHtml(JSON.stringify(response, null, 4))}</pre>`, "OK");
+            return false;
+        }
+
+        return true;
     }
 
     /**
--- a/css/app.css
+++ b/css/app.css
@@ -1365,6 +1365,10 @@ thumbnail-slider img {
     overflow-y: hidden;
 }
 
+.modal-dialog pre {
+    max-height: 200px;
+}
+
 .text-weight-bold {
    font-weight: 700;
  }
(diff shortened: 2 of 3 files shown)


pwalczysko:
I have removed renamed a file in ManagedRepo and then opened the image in iviewer.

I get

Screenshot 2024-04-26 at 15 31 02

src/model/image_info.js
```diff
@@ -360,6 +368,39 @@ export default class ImageInfo {
...
+     * @param {Object} response the response object
+     * @memberof ImageInfo
+     * @return {Boolean} True if data is valid. Also shows dialogs with errors
+     */
+    validateImageInfo(response) {
+
+        if (response.ConcurrencyException) {
+            Ui.showModalMessage(`<p>Image is not currently viewable</p>
+                <pre>ConcurrencyException</pre>
+                <small>A pyramid of zoom levels is not available. Generation may be in progress if this is enabled on your server.</small>`, "OK");
```
pwalczysko, line 384:
```suggestion
                <small>A pyramid of zoom levels is not available. Please contact your administrator, see https://omero.readthedocs.io/en/stable/sysadmins/limitations.html#large-images.</small>`, "OK");
```

I realize this ^^^ is not ideal suggestion, but I am against making of false hopes - as we do not recommend pyramid generation to be enabled, we should not assume (as the present comment does) that "waiting it out" is a good/default option.

pwalczysko:
How to elicit ``no channel data loaded`` error ?

pwalczysko (changes requested):
one reformulation request and a question about testing setup, thank you

src/model/image_info.js
will-moore: Bit of a side track... but those docs don't describe how to enable/disable pyramid generation. Do we have install docs that recommend and describe how to disable pyramid generation? This is likely to have an effect on how likely it is for a particular server to have pyramids disabled, since the default is still to be enabled, right?
Also those docs should mention https://www.glencoesoftware.com/products/ngff-converter/ .

Unfortunately it's not possible for a client to know if pyramids are disabled, since this gives a SecurityViolation:
```
conn.getConfigService().getConfigValue("omero.server.nodedescriptor")
```

How about.
"A pyramid of zoom levels is not available. Please contact your administrator if this does not resolve itself. See <a href='https://omero.readthedocs.io/en/stable/sysadmins/limitations.html#large-images'>Limitations: Large Images</a>"

sbesson: The property is actually `"omero.server.nodedescriptors"` and has been made user-visible since OMERO 5.6.6. Older versions of the server will definitely throw a SecurityViolation.

Note this is not necessarily the only way to disable the `PixelData` service e.g. you might have modified your templates manually but I think this is by far the option that is the most usable from an API perspective and the documentation should mostly likely document this workflow if any.

will-moore:
The `no channel data` is designed to handle the response returned from https://github.com/ome/omero-web/blob/e1b1355ebd4f9de8b855dd2c55f59d86c061a81b/omeroweb/webgateway/marshal.py#L257 

The code comment says `Will happen if an image has bad or missing pixel data` but I don't know if that's still true or how to trigger it`. 
I can create that exception by adding `test = "fail" + 1` into the try/except block above:

Screenshot 2024-04-26 at 17 06 27

src/model/image_info.js
will-moore: @sbesson Thanks for catching my typo!
When connected to nightshade I get:
```
>>> conn.getConfigService().getConfigValue("omero.server.nodedescriptors")
''
```
Which would indicate that pyramid generation is enabled there?
For `outreach` server I get
```
'master:Blitz-0,Indexer-0,Processor-0,Storm,Tables-0'
```

sbesson: An empty value (default) means all services are enabled via configuration - see https://omero.readthedocs.io/en/stable/sysadmins/config.html#omero-server-nodedescriptors

will-moore: Great. Since clients can access this, I'll load it from iviewer so we can show a more useful error message for ConcurrencyException. It probably makes sense that we do something similar in webclient too, but most useful would be the Importer, since you need to know at import time if you're importing a Big Image without pyramids.

will-moore:
With that last commit, we try to load the `omero.server.nodedescriptors` config from the server.
If we can load it and it is set and it doesn't contain "PixelData", then we know that pyramid creation is disabled and the error with `ConcurrencyException` should look like:

Screenshot 2024-04-29 at 16 12 11

Otherwise it's probable that we can generate pyramids (unless this is disabled on an older server that doesn't allow us to query it), so we show this:

Screenshot 2024-04-29 at 16 12 40

pwalczysko:
Screenshot 2024-05-02 at 13 22 41
Screenshot 2024-05-02 at 13 02 10


With a "large" image without pyramids, I am having the two messages if pyramid gen is a) disabled (first screenshot) or b) enabled (second screenshot). Lgtm

pwalczysko:
I understand that https://github.com/ome/omero-iviewer/pull/474#issuecomment-2079684660 is very hard to elicit and was never encountered, actually the error is being caught for completeness sake.

Lgtm in total.

PR #485, opened 2024-10-10
Dynamic LUT for iviewer

Tom-TBT:
Here is the implementation of the dynamic LUT handling for the iviewer.
dynamic LUT in omero-web: https://github.com/ome/omero-web/pull/581

Same as for https://github.com/ome/omero-figure/pull/596, if omero-web can provide the LUT png (generated dynamically from the LUTs present on the server), then it will use that png to show all LUT previews.

If not, it falls back to the old behavior (not showing previews of the extra LUTs added on the server).

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -302,11 +302,11 @@ export default class Context {
      */
     setUpLuts() {
         $.ajax(
-            {url : this.server + this.getPrefixedURI(IVIEWER) + "/is_dynamic_lut/",
+            {url : this.server + this.getPrefixedURI(WEBGATEWAY) + "/luts/",
             success : (response) => {
                 // Check first whether omero-web can provides LUT dynamically
                 // and set URL accordingly
-                let is_dynamic_lut = response.is_dynamic_lut;
+                let is_dynamic_lut = Boolean(response.png_luts_new);
                 if (is_dynamic_lut) {
                     this.luts_png.url =
                         this.server + this.getPrefixedURI(WEBGATEWAY, false) + "/luts_png/";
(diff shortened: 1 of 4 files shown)


jburel:
@will-moore could you please validate the following:
* omero-web PR included https://github.com/ome/omero-web/pull/581 and this PR: 8 new LUTs in the menu
* omero-web without https://github.com/ome/omero-web/pull/581 and this PR: menu without the 8 new ones (static image)

src/app/context.js
```diff
@@ -301,41 +301,56 @@ export default class Context {
...
-            for (let [id, conf] of this.image_configs) conf.changed();
-        }
-        lutsPng.src = this.luts_png.url;
-
-        // now query the luts list
-        let server = this.server;
-        let uri_prefix =  this.getPrefixedURI(WEBGATEWAY);
         $.ajax(
-            {url : server + uri_prefix + "/luts/",
+            {url : this.server + this.getPrefixedURI(IVIEWER) + "/is_dynamic_lut/",
```
will-moore, line 305: I'm wondering if you could save yourself a call (and not need the new `is_dynamic_lut/` endpoint) if you simply load the `/luts/` json as before, then check for `response.png_luts_new` and if this exists then `is_dynamic_lut` is True?

I just tried this and my diff looks like this (not updated the indents yet to minimise the diff):

```
+++ b/src/app/context.js
@@ -302,11 +302,12 @@ export default class Context {
      */
     setUpLuts() {
         $.ajax(
-            {url : this.server + this.getPrefixedURI(IVIEWER) + "/is_dynamic_lut/",
+            {url : this.server + this.getPrefixedURI(WEBGATEWAY) + "/luts/",
             success : (response) => {
                 // Check first whether omero-web can provides LUT dynamically
                 // and set URL accordingly
-                let is_dynamic_lut = response.is_dynamic_lut;
+                let is_dynamic_lut = Boolean(response.png_luts_new);
                 if (is_dynamic_lut) {
                     this.luts_png.url =
                         this.server + this.getPrefixedURI(WEBGATEWAY, false) + "/luts_png/";
@@ -324,15 +325,6 @@ export default class Context {
                 }
                 lutsPng.src = this.luts_png.url;
 
-                // now query the luts list
-                let server = this.server;
-                let uri_prefix =  this.getPrefixedURI(WEBGATEWAY);
-                $.ajax(
-                    {url : server + uri_prefix + "/luts/",
-                    success : (response) => {
-                        if (typeof response !== 'object' || response === null ||
-                            !Misc.isArray(response.luts)) return;
-
                         if (is_dynamic_lut) {
                             // If it's dynamic, uses the new list instead
                             response.png_luts = response.png_luts_new
@@ -349,8 +341,6 @@ export default class Context {
                                 this.luts.set(mapValue.name, mapValue);
                             });
                         for (let [id, conf] of this.image_configs) conf.changed();
-                    }
-                });
             }
         });
     }
```

It seems to work for me. Does that make sense?

Tom-TBT: It makes sense yes. I think the reason I did that is because I originally returned a single LUT list with the updated `this.getPrefixedURI(WEBGATEWAY) + "/luts/"`

But since it now returns `png_luts_new` in the response, there's a way to know if it's the old or new response.

Can you please commit these changes?

There should be a similar solution for omero-figure then. I'll look into it.

will-moore (approved):
Re-deployed those changes on merge-ci and looking good as before.

jburel:
@will-moore Nobody had a chance to review your commits. Please do not merge without review
This PR was not fixing a build problem as far as I am aware

jburel:
This was unclear in the sequence. Thanks for the clarification

PR #500, opened 2025-01-09
Support max_active_channels

### LINKED ISSUE #472: Limit number of active Channels
When using the `omero-ms-image-region` microservice, the number of active channels is limited by default to 10 (configurable at https://github.com/glencoesoftware/omero-ms-image-region/blob/a3c6e61ecb02c608f4c1f8c296c259d7de9f5cc3/src/dist/conf/config.yaml#L12)

The limit can be queried at 
https://github.com/glencoesoftware/omero-ms-image-region/blob/master/src/main/java/com/glencoesoftware/omero/ms/image/region/ImageRegionMicroserviceVerticle.java#L486

Although I don't see how that method maps to a URL that can be accessed by iviewer?

Even when microservices are not deployed, users may wish to configure a limit on active channels. This could (should) be a new server-side configuration (allow it to be accessed via the rendering engine itself as well as other clients.

will-moore:
Fixes #472 

This turns off all channels over the `max_active_channels` limit when the image loads and also prevents the user from turning on more channels than this limit. Use will see a message E.g. "The maximum number of active channels is 10."

When microservices are not available, we use a default value for max_active_channels.
This can be configured with:

```
omero config set omero.web.iviewer.max_active_channels 10
```
But what should the default value be if without the user setting anything? 
Currently we don't limit the active channels, so to maintain that behaviour, we could set the value quite high, e.g. 50.
I seem to remember that saving rendering settings fails for around 60+ channels.

Or we go for 10 (same as microservices) or a bit higher, e.g. 20?


francesw:
Tested a few images from idr0076 and idr0158 on idr-testing.  Works well, max 10 channels message appears when trying to click on channel 11.  Perhaps include in the message that you have to deselect a channel before you can reselect a channel.

pwalczysko:
Tested on idr-testing with idr0076 and idr0158. Wfm and I am happy with the level of messaging.

jburel:
Other settings are described in the README
Since the list is increasing, I will suggest to start a separate "configuration" page with a link to it in the README

will-moore:
For comparison, Pathviewer message is "Cannot activate channel, maximum number of channels already active".

Screenshot 2025-01-20 at 11 37 07

francesw:
@will-moore On Pathviewer, they also have a count of the active channels with the maximum number stated, which can be useful and helps to make max channel number clear.  Not sure if this is something we would also like to have.  
Screenshot 2025-01-20 at 13 36 44

will-moore:
@francesw That's kinda nice, but I would prefer not to add it as it's a fair bit more work, unless it's really required?

jburel:
Discussed with @will-moore. My comment about the description of the configuration will be fixed in a follow up PR.

PR #503, opened 2025-02-26
Basic working tile-coordinates grid shown over the image

will-moore:
Investigating the display of tile coordinates as a possible tool for users to keep track of which parts of a big Image they are viewing (or have viewed before).

Functionality is similar to https://openlayers.org/en/latest/examples/canvas-tiles.html but I couldn't get that code to work since it complained about 'no projection' when I added a `Debug` layer in iviewer.
So instead, similar effect was produced with a custom `tileLoadFunction` in a subclass of the OmeroImage layer (should use same tile coordinates etc).

I was also trying to keep track of the tiles where `tileLoadFunction()` has been called before, so that when we render a tile, we can show if it's been loaded before `count: 1` etc in the screenshot. However, it seems to not get called more than once for each tile.

Screenshot 2025-02-26 at 14 15 01

### DIFF
--- a/src/viewers/viewer/Viewer.js
+++ b/src/viewers/viewer/Viewer.js
@@ -30,6 +30,8 @@ import Tile from 'ol/layer/Tile';
 import Vector from 'ol/layer/Vector';
 import View from 'ol/View';
 import OlMap from 'ol/Map';
+import TileState from 'ol/TileState';
+import EventType from 'ol/events/EventType';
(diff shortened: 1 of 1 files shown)


dinoflagelado:
a good start, that is how I imagine it would look like.  The user has to have the capability to design the tile structure (how many rows and columns per tile), as the contents of slides can be very different.

PR #511, opened 2025-12-04
Support img_detail/9837058/?labelimage=9837059 as extra Tile layer

will-moore:
Experimental only...

Specify an Image ID of a segmentation label image to load as a transparent png layer on top of the image.

We manually render the image to png using `image.getPrimaryPixels().getPlane(z, c, t)` and numpy so we don't get jpeg artifacts. Not implemented for tiles yet (only for whole planes).

E.g. using idr0079;

Screenshot 2025-12-04 at 09 43 03


will-moore:
Closing in favour of #532

PR #517, opened 2026-03-20
Fix roi link permissions

will-moore:
Fixes issue reported at https://forum.image.sc/t/rois-link-in-omero-table-not-working/119676/4 where we failed to use a "cross-group" query, so a link to ROI in iviewer didn't work if the ROI wasn't in your default group.

To test:
 - As user-3 on merge-ci, login and then enter URL `/web/iviewer/?roi=3819` (that ROI is owned by user-3 but not in their default group)
 - Currently this fails with `Error` dialog in iviewer
 - This should find the correct image and open it in iviewer

### DIFF
--- a/plugin/omero_iviewer/views.py
+++ b/plugin/omero_iviewer/views.py
@@ -481,7 +481,8 @@ def roi_image_data(request, obj_type, obj_id, conn=None, **kwargs):
     """ Get image_data for image linked to ROI """
     image_id = None
     if obj_type == 'roi':
-        roi = conn.getQueryService().get('Roi', int(obj_id))
+        roi = conn.getQueryService().get('Roi', int(obj_id),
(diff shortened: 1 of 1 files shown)


pwalczysko:
I was able to repeat the bug on outreach server:

In default group, the link to a ROI works fine.
In non-default group, the link to a ROI fails with error dialog in iviewer

With this PR:

Works as expected. I am noting that the ROI is in a non-default group of user-3

Maybe for another PR:
I am seeing an uinexpected behaviour with another link connected to iviewer. On the very ROI https://merge-ci.openmicroscopy.org/web/iviewer/?roi=3819, select the Info tab in RHP, then click on Image ID (4003). Observe that the image does not get selected in webclient (the Dataset just beneath it does if clicked)

pwalczysko (approved):

PR #526, opened 2026-05-01
Add keyboard handling for Left, Right, Up, Down Arrows to pan

will-moore:
As discussed with Carlos, cc @pwalczysko 

To test:
 - Use Left, Right, Up, Down keys to pan viewport in iviewer
 - Check that we DON'T pan when using the Arrow Keys for other purposes, e.g:
   - Focus a Z/T slider and use Arrows to increment
   - When editing Comment text boxes etc
   - Any other usages you can think of??


pwalczysko:
Tested on Safari, Chrome, Firefox. When the z-slider in focus, does not pan as expected.
Hitting key when editing text does not pan as expected.

lgtm

pwalczysko (approved):

PR #529, opened 2026-05-18
Image id and roi id initial params

will-moore:
When linking from e.g. OMERO.tables or anywhere that we have ONLY a Roi/Shape ID, we use e.g. `/iviewer/?roi=789` or `/iviewer/?shape=456`.

However, this doesn't use `webclient/img_detail/123/` URL so it ignores any setting we have for `omero.web.viewer.view` to use a different viewer (such as idr-gallery) for handling images.

In this PR, when iviewer loads we check if the `image ID` has been passed in the URL (which tells us whether we're at e.g. `webclient/img_detail/IMAGE_ID/`) and if we're NOT at that URL, then we lookup the image ID (via ROI or SHAPE) and then redirect to `/img_detail/123/?shape=456`.

Also, we now handle Image ID AND Roi/Shape IDs, so that `webclient/img_detail/123/?shape=456` will load the image *and then* select the Shape. Previously we *only* handled Image ID *or* Shape ID but not both.

The only time when you might not want the redirect behaviour is if you have some other viewer set at `omero.web.viewer.view` (e.g. Pathviewer) but you still want iviewer to be available at `/iviewer/?images=123`. So, we only redirect if 

```
$ omero config set omero.web.iviewer.redirect_iviewer True
```

This is added to the README.

To test:

 - The config above has already been set on merge-ci OMERO.web job
 - Find or create & Save a Shape in iviewer
 - Click on the link icon (top right of ROI panel in iviewer) and copy the link. E.g. /iviewer/?shape=456
 - Paste it in the URL and you should get redirected to /webclient/img_detail/IMAGEID/?shape=456
 - Also, selecting multiple images in the webclient and Open With > iviewer which gives /iviewer/?images=7,8,9 should also redirect to webclient/img_details/7/?images=7,8,9 and open the multiple images
 - Select a Well in webclient and in the right panel Open with > iviewer. This will redirect to show all the images in the well as in previous test above

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -67,7 +67,7 @@ This may first remove other ``image_configs`` before creating a new one and load
 
     // src/app/context.js
 
-    let image_config = new ImageConfig(this, image_id, parent_id, parent_type);
+    let image_config = new ImageConfig(this, image_id, roi_id, shape_id, parent_id, parent_type);
     // store the image config in the map and make it the selected one
     this.image_configs.set(image_config.id, image_config);
     this.selectConfig(image_config.id);
--- a/plugin/omero_iviewer/README.rst
+++ b/plugin/omero_iviewer/README.rst
@@ -52,7 +52,7 @@ To enable the "open with" feature:
        "script_url": "omero_iviewer/openwith.js", "label": "OMERO.iviewer"}]'
 
 If you want to redirect ``/iviewer/`` URLs to instead use ``/webclient/img_detail/ID`` so
-that the viewer configured at ``omero.web.viewer.view`` is used then use the 
+that a different viewer configured at ``omero.web.viewer.view`` is used then use the 
 config below.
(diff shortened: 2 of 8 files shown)


dominikl:
Thanks Will. Looks good, all working, except the "Well" workflow: Went to a well, and "open with iviewer" (https://merge-ci.openmicroscopy.org/web/iviewer/?well=654 ), got error:
```
...
File "/home/omero/workspace/OMERO-web/.venv3/lib64/python3.11/site-packages/omero_iviewer/views.py", line 96, in index
    if image_id is not None:
       ^^^^^^^^
UnboundLocalError: cannot access local variable 'image_id' where it is not associated with a value
<WSGIRequest: GET '/web/iviewer/?well=654'>
...
```

dominikl:
Oh sorry, just saw you already fixed that, but it's not on merge-ci yet!

dominikl (approved):
👍 All good. Everything works. And all routes are going through `webclient/` as far as I can see.

PR #534, opened 2026-05-26
Fix settings.md link to master branch. Move iviewer redirect settings

will-moore:
I realised the link under https://github.com/ome/omero-iviewer#settings is not working.

Also added the same link on the `plugin/omero_iviewer/README.rst` page after install instructions and moved the `iviewer redirect` info from here into the settings page (since I'd forgotten about the settings page when I added that).


dominikl (approved):
👍 Looks good to me!

PR #562, opened 2026-08-13
Keyboard select roi

### LINKED ISSUE #561: Keyboard shortcuts for ROI selection
"It would be very useful to navigate the ROIs list using the arrows in the keyboard. We have images with thousands of Rois and moving from one roi to the next is a process we do very often.
Right now, it can only be done with the mouse.  It would be very helpful if we could move from one roi to the next using the keyboard."

Since we already use Arrows for panning images, we need to check whether we can also use Arrows for ROI selection (if the ROI panel has focus) or if it might be clearer to use other Keys for ROI selection?

will-moore:
Fixes #561 

To test:

 - Select an ROI and use Up/Down Arrow keys to select prev/next ROI
 - If the ROIs tab is not selected, or NO ROIs are selected, the Up/Down arrows will instead cause the Image to Pan as usual
 - Left/Right keys still do only image panning

Video below shows:
 - Use of Arrow keys for panning
 - Drawing ROIs: since NO ROIs are selected, Arrow keys continue to work for panning while drawing ROIs
 - ROIs saved and ROI selected: Now, Up/Down Arrows change ROI selection (and image pans to selected ROIs)

https://github.com/user-attachments/assets/bf73b15e-9ae8-4479-8c9c-d9110414ecbc

### DIFF
--- a/src/app/context.js
+++ b/src/app/context.js
@@ -593,22 +593,22 @@ export default class Context {
                 if (typeof keyHandlers === 'undefined' ||
                     event.target.nodeName.toUpperCase() === 'INPUT') return;
 
-                // we allow the browser's default action and event
-                // bubbling unless one handler returns false
-                let allowDefaultAndPropagation = true;
+                // sort to put "global" last - give other groups (tabs) a chance to handle first
+                let keys = Object.keys(keyHandlers).sort((x, y) => x === 'global' ? 1 : -1);
                 try {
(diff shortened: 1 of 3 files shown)


pwalczysko (approved):
tested on merge-ci. Works as expected. Thanks

dinoflagelado:
that is how I imagined the feature, thanks

